PR #16, opened 2016-05-11
Cling CodeCompleteConsumer

CristinaCristescu:


lib/UserInterface/ClingCodeCompleteConsumer.h
```diff
@@ -0,0 +1,38 @@
...
+public:
+  ClingCodeCompleteConsumer(const CodeCompleteOptions &Opts)
+    : CodeCompleteConsumer(Opts, false), CCTUInfo(new GlobalCodeCompletionAllocator) { }
+  ~ClingCodeCompleteConsumer() {}
+  CodeCompletionAllocator &getAllocator() override { return CCTUInfo.getAllocator();}
+  CodeCompletionTUInfo &getCodeCompletionTUInfo() override { return CCTUInfo; }
+};
+
+class ClingTabCompletion : public textinput::TabCompletion {
+  ClingCodeCompleteConsumer* fCodeCompleteConsumer;
```
pcanal: Please add an explicit comment if the pointee is not owned by the ClingTabCompletion object. thanks.

karies: Well, it's a raw pointer, so by definition for code written in 2016... :-)

pcanal: For better or worse, one year from now, I will not remember what year this code was written :)  ... and hammering the point does not hurt ... so please :)

lib/UserInterface/UserInterface.cpp
```diff
@@ -118,6 +121,16 @@ namespace cling {
     std::unique_ptr<TerminalDisplay> D(TerminalDisplay::Create());
     TextInput TI(*R, *D, histfilePath.empty() ? 0 : histfilePath.c_str());
 
+    // Inform Sema about our code complete consumer
+    const CodeCompleteOptions& ClingCodeCompleteOpts = m_MetaProcessor->getInterpreter().getCI()->getFrontendOpts().CodeCompleteOpts;
+    ClingCodeCompleteConsumer* ClingConsumer = new ClingCodeCompleteConsumer(ClingCodeCompleteOpts);
```
karies: Who is owning this? Is it ever deleted? If it's meant to be owned by the local scope than please don't use `new` but a local non-pointer variable: `ClingCodeCompleteConsumer ClingConsumer(ClingCodeCompleteOpts);`

lib/UserInterface/UserInterface.cpp
```diff
@@ -118,6 +121,16 @@ namespace cling {
...
+    m_MetaProcessor->getInterpreter().getCI()->setCodeCompletionConsumer(ClingConsumer);
+    // Inform text input about the code complete consumer
+    ClingTabCompletion* CTabCompletion = new ClingTabCompletion();
```
karies: Same here.

lib/UserInterface/ClingCodeCompleteConsumer.h
karies: More seriously: Cristina sent this as an early preview; `ClingCodeCompleteConsumer` and `ClingTabCompletion` might get merged, and if not one might own the other - we'll see. I.e. it's still a bit early for reviews :-)

CristinaCristescu:
Updated for review and discussion.

include/cling/Interpreter/CompilationOptions.h
```diff
@@ -72,6 +78,7 @@ namespace cling {
       CodeGenerationForModule = 0;
       IgnorePromptDiags = 0;
       CheckPointerValidity = 1;
+      CodeCompletionOffset = -1;
```
karies: It's enough to have the in class initializer, no need for repeating it here

include/cling/Interpreter/Interpreter.h
```diff
@@ -195,6 +195,19 @@ namespace cling {
     ///\brief Worker function, building block for interpreter's public
     /// interfaces.
     ///
+    ///\param [in] input - The input being code completed.
+    ///\param [in] CompilationOptions - The option set driving the compilation.
+    ///\param [out] T - The cling::Transaction of the compiled input.
+    ///
+    ///\returns Whether the operation was fully successful.
+    ///
+    CompilationResult CodeCompleteInternal(const std::string& input,
```
karies: Can you merge this with codeComplete? No need for an internal function here, this was just to avoid code duplication in the other cases

include/cling/Interpreter/Interpreter.h
```diff
@@ -468,6 +481,19 @@ namespace cling {
...
+    /// The interface circumvents the most of the extra work necessary to
+    /// code complete code.
+    ///
+    /// @param[in] input - The input containing the string to be completed.
+    /// @param[out] T - The cling::Transaction of the input
+    ///
+    ///\returns Whether the operation was fully successful.
+    ///
+    CompilationResult codeComplete(const std::string& input, unsigned offset,
+                                   Transaction** T = 0);
```
karies: We don't need the transaction, we don't expect anything valuable as result, given that this might not even be parsed

lib/UserInterface/UserInterface.cpp
```diff
@@ -118,6 +121,12 @@ namespace cling {
     std::unique_ptr<TerminalDisplay> D(TerminalDisplay::Create());
     TextInput TI(*R, *D, histfilePath.empty() ? 0 : histfilePath.c_str());
 
+    // Inform text input about the code complete consumer
+    ClingTabCompletion* CompletionConsumer = new ClingTabCompletion(
```
karies: Can this become a stack allocated CompletionConsumer?

lib/UserInterface/ClingCodeCompleteConsumer.h
```diff
@@ -0,0 +1,57 @@
...
+    printf("printing..");
+  }
+
+  bool Complete(Text& Line /*in+out*/,
+                          size_t& Cursor /*in+out*/,
+                          EditorRange& R /*out*/,
+                          std::vector<std::string>& DisplayCompletions /*out*/) override {
+    //Create the interpreter
+    const char * const argV = "cling";
+    Interpreter CodeCompletionInterp(1, &argV);;
```
karies: Does this not need the llvm resource dir?

karies:
Cool, congrats! I'm curious to hear tomorrow how far that gets us :-)

include/cling/Interpreter/CompilationOptions.h
CristinaCristescu: Done.

include/cling/Interpreter/Interpreter.h
CristinaCristescu: Done.

include/cling/Interpreter/Interpreter.h
CristinaCristescu: I thought so..removed now

lib/UserInterface/ClingCodeCompleteConsumer.h
CristinaCristescu: Llvmdir has default 0, I am not sure what the implications are for the correctness of code completion, though.

lib/UserInterface/ClingTabCompletion.h
```diff
@@ -6,12 +6,13 @@
...
+#include "cling/Interpreter/InterpreterCallbacks.h"
 
 #include "textinput/Text.h"
 #include "textinput/Editor.h"
 #include "textinput/Callbacks.h"
 
-namespace textinput{
+namespace cling {
 class ClingTabCompletion : public textinput::TabCompletion {
   const cling::Interpreter& ParentInterp;
```
karies: Can you take a non-const parent Interpreter instead of const-casting?

lib/Interpreter/InterpreterCallbacks.cpp
```diff
@@ -239,6 +240,17 @@ namespace cling {
     return false;
   }
 
+  void InterpreterCallbacks::CreateCodeCompleteConsumer(Interpreter* child) {
```
karies: ROOT now needs to
- create a CCC
- set it in the interpreter
- set it in Sema.

I'd prefer if ROOT only needs to do the first step, which is why I was arguing for returning a `unique_ptr` (or, if that's easier on the clang interfaces, just a CCC*) and doing the setting  on the caller side, i.e. in `ClingTabCompletion::Complete()`.

lib/UserInterface/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ endif()
 
 add_cling_library(clingUserInterface
   UserInterface.cpp
+  ClingCodeCompleteConsumer.cpp
```
karies: I think the patch does not contain ClingCodeCompleteConsumer.cpp? Did you git add it?

karies: Do we actually still need ClingCodeCompleteConsumer? Or can we just use the default clang::PrintingCCC?

CristinaCristescu: No we don't need it. I updated the CMakeList.txt

karies:
Should we close this, too?

CristinaCristescu:
Hi Axel,

Yes please close that!

Also, I have some errors on test of course, I;ll check the master’s ctest and let you know.

Cheers,
Cristina

On 15 Jul 2016, at 21:41, Axel Naumann <notifications@github.com<mailto:notifications@github.com>> wrote:


Should we close this, too?

karies:
Superseded by PR179 against my ROOT repo

PR #17, opened 2016-06-13
Remaining Windows fixes.

marsupial:


lib/Interpreter/TransactionUnloader.cpp
```diff
@@ -930,6 +936,25 @@ namespace clang {
...
 
+  bool DeclUnloader::VisitLinkageSpecDecl(LinkageSpecDecl* LSD) {
+    // LinkageSpecDecl: DeclContext
+
+    // Clang caches all references to extern "C" globals
+    if (!m_Sema->LocallyScopedExternCDecls.empty()) {
+      for (Decl * D: LSD->decls()) {
+        if (isa<NamedDecl>(D)) {
+          NamedDecl *ND = dyn_cast<NamedDecl>(D);
+          auto externC = m_Sema->LocallyScopedExternCDecls.find(ND->getDeclName());
```
vgvassilev: Not sure if this exceeds but we try to stick to 80 columns.

test/CodeUnloading/CGlobalDecl.C
```diff
@@ -0,0 +1,43 @@
+//------------------------------------------------------------------------------
```
vgvassilev: Could you squash the test into the previous commit. It is better when browsing the history.

lib/Interpreter/CIFactory.cpp
```diff
@@ -807,10 +807,20 @@ namespace {
...
+          = CI->getInvocation().getCodeGenOpts().MainFileName;
+        if (!CGOptsMainFileName.empty())
+          Filename = CGOptsMainFileName.c_str();
+
+     However when running cling on the command-line any attempt to execute
+     a file in the current working directory fails when this is done.
+     If there'se a chance clang is parsing in parallel while we're initing
+     if would be good to put this back in and figure out how to get the decls
+     in IncrementalParser::ParseInternal
+    */
```
vgvassilev: I am not sure I understand the comment. Could you add a test case, showing what is the fix about?

lib/Interpreter/Transaction.cpp
```diff
@@ -64,12 +64,23 @@ namespace cling {
   }
 
   NamedDecl* Transaction::containsNamedDecl(llvm::StringRef name) const {
-    for (auto I = decls_begin(), E = decls_end(); I < E; ++I)
+    for (auto I = decls_begin(), E = decls_end(); I < E; ++I) {
       for (auto DI : I->m_DGR) {
-        if (NamedDecl* ND = dyn_cast<NamedDecl>(DI))
+        if ( DI->getKind() == Decl::LinkageSpec ) {
```
vgvassilev: Could you remove the spaces after ( and ). Also could you use isa<LinkageSpecDecl>(DI) instead?

test/Commands/XexternC.C
```diff
@@ -0,0 +1,17 @@
+//------------------------------------------------------------------------------
```
vgvassilev: Could you meld into the actual commit, fixing the issue?

lib/Interpreter/IncrementalParser.cpp
```diff
@@ -125,7 +125,7 @@ namespace {
   if (!II)
     return;
   const clang::DefMacroDirective* MD
-    = llvm::dyn_cast<clang::DefMacroDirective>(PP.getMacroDirective(II));
+    = llvm::dyn_cast_or_null<clang::DefMacroDirective>(PP.getMacroDirective(II));
```
vgvassilev: It seems the real issue is more subtle. This allows the argument of dyn_cast_or_null to be null. I am assuming that CLING_CXXABIS is not defined in the TU. Perhaps @karies could help.

lib/Interpreter/Interpreter.cpp
```diff
@@ -1112,13 +1112,15 @@ namespace cling {
   }
 
   Interpreter::CompilationResult
-  Interpreter::loadFile(const std::string& filename,
-                        bool allowSharedLib /*=true*/,
-                        Transaction** T /*= 0*/) {
+  Interpreter::loadLibrary(const std::string& filename, bool lookup ) {
```
vgvassilev: space before ).

lib/Interpreter/Interpreter.cpp
```diff
@@ -1112,13 +1112,15 @@ namespace cling {
...
     DynamicLibraryManager* DLM = getDynamicLibraryManager();
-    std::string canonicalLib = DLM->lookupLibrary(filename);
-    if (allowSharedLib && !canonicalLib.empty()) {
-      switch (DLM->loadLibrary(canonicalLib, /*permanent*/false, /*resolved*/true)) {
+    std::string canonicalLib;
+    if ( lookup )
```
vgvassilev: spaces.

include/cling/Interpreter/Interpreter.h
```diff
@@ -540,6 +540,25 @@ namespace cling {
...
+                                  bool lookup = true);
+
+    ///\brief Loads header file
+    ///
+    ///\param [in] filename - The file to loaded.
+    ///\param [out] T -  Transaction containing the loaded file.
+    ///\returns result of the compilation.
+    ///
+    CompilationResult loadHeader(const std::string& filename,
+                               Transaction** T = 0);
```
vgvassilev: Could you add some tests. This part is going to be heavily used because it is a very visible interface.

lib/Interpreter/ValueExtractionSynthesizer.cpp
```diff
@@ -353,7 +350,9 @@ namespace {
                                              E).get();
         CallArgs.push_back(AddrOfE);
       }
-      else if (desugaredTy->isPointerType()) {
+      else if (desugaredTy->isPointerType()
```
vgvassilev: What about desugaredTy->isAnyPointerType()? Could you add the objc test case if you have one?

lib/Interpreter/ValueExtractionSynthesizer.cpp
```diff
@@ -371,11 +370,15 @@ namespace {
         // case, because of the overload resolution.
         CallArgs.push_back(E);
       }
-      Call = m_Sema->ActOnCallExpr(/*Scope*/0, m_UnresolvedNoAlloc,
+
+      if ( CallArgs.size() > nArgs ) {
```
vgvassilev: spaces.

lib/Interpreter/ValueExtractionSynthesizer.cpp
```diff
@@ -371,11 +370,15 @@ namespace {
...
+        Call = m_Sema->ActOnCallExpr(/*Scope*/0, m_UnresolvedNoAlloc,
                                    locStart, CallArgs, locEnd);
```
vgvassilev: Indent.

lib/Interpreter/CIFactory.cpp
```diff
@@ -389,7 +389,7 @@ namespace {
   }
 
   ///\brief Adds standard library -I used by whatever compiler is found in PATH.
-  static void AddHostCXXIncludes(std::vector<const char*>& args) {
+  static void AddHostCXXIncludes(std::vector<const char*>& args, const char *clang ) {
```
vgvassilev: star should to the char type.

lib/Interpreter/CIFactory.cpp
```diff
@@ -433,12 +433,14 @@ namespace {
         }
       }
 
-      static const char *CppInclQuery =
-        "echo | LC_ALL=C " LLVM_CXX " -xc++ -E -v - 2>&1 >/dev/null "
+      std::string CppInclQuery("echo | LC_ALL=C ");
+      CppInclQuery.append(clang ? clang  : LLVM_CXX);
```
vgvassilev: one too many spaces.

lib/Interpreter/CIFactory.cpp
```diff
@@ -626,7 +631,15 @@ namespace {
       argvCompile.insert(argvCompile.begin() + 2, lang);
     }
 
-    AddHostCXXIncludes(argvCompile);
+    std::string clang = llvm::sys::fs::getMainExecutable("cling", (void*)(intptr_t) locate_cling_executable);
```
vgvassilev: more than 80 cols.

lib/Interpreter/CIFactory.cpp
```diff
@@ -626,7 +631,15 @@ namespace {
...
+    {
+      clang = llvm::sys::path::parent_path(clang);
+      SmallString<128> P(clang);
+      llvm::sys::path::append(P, "clang");
+      clang.assign(&P[0], P.size());
+    }
+
+    AddHostCXXIncludes(argvCompile, llvm::sys::fs::is_regular_file(clang) ? clang.data() : NULL);
```
vgvassilev: > 80 cols.

lib/Interpreter/CIFactory.cpp
```diff
@@ -626,7 +631,15 @@ namespace {
       argvCompile.insert(argvCompile.begin() + 2, lang);
     }
 
-    AddHostCXXIncludes(argvCompile);
+    std::string clang = llvm::sys::fs::getMainExecutable("cling", (void*)(intptr_t) locate_cling_executable);
+    {
```
vgvassilev: Remove the nested scope. Please also add a comment explaining that you are looking freshly built clang sitting next to cling.

include/cling/Interpreter/DynamicLibraryManager.h
```diff
@@ -87,14 +87,17 @@ namespace cling {
...
+    ///\param [in] resolved - Whether libStem is an absolute path or resolved
+    ///               from a previous call to DynamicLibraryManager::lookupLibrary
     ///
     ///\returns kLoadLibSuccess on success, kLoadLibAlreadyLoaded if the library
     /// was already loaded, kLoadLibError if the library cannot be found or any
     /// other error was encountered.
     ///
-    LoadLibResult loadLibrary(const std::string& libStem, bool permanent);
+    LoadLibResult loadLibrary(const std::string& libStem, bool permanent,
+                              bool resolved = false );
```
vgvassilev: space.

include/cling/Interpreter/Interpreter.h
```diff
@@ -540,6 +540,25 @@ namespace cling {
     ///
     std::string lookupFileOrLibrary(llvm::StringRef file);
 
+    ///\brief Loads a shared library.
+    ///
+    ///\param [in] filename - The file to loaded.
+    ///\param [in] lookup - Whether to try to resolve the filepath
+    ///
+    ///\returns kMoreInputExpected is returned when file could not be found
+    /// otherwise kSuccess or kFailure
```
vgvassilev: missing empty ///

lib/Interpreter/DynamicLibraryManager.cpp
```diff
@@ -323,10 +323,14 @@ namespace cling {
 
   DynamicLibraryManager::LoadLibResult
   DynamicLibraryManager::loadLibrary(const std::string& libStem,
-                                     bool permanent) {
-    std::string canonicalLoadedLib = lookupLibrary(libStem);
-    if (canonicalLoadedLib.empty())
-      return kLoadLibNotFound;
+                                     bool permanent, bool resolved) {
+    std::string       lResolved;
```
vgvassilev: too many spaces :)

lib/Interpreter/DynamicLibraryManager.cpp
```diff
@@ -323,10 +323,14 @@ namespace cling {
...
+    const std::string &canonicalLoadedLib = resolved ? libStem : lResolved;
```
vgvassilev: move & to string. Check for 80 col violation.

lib/Interpreter/Interpreter.cpp
```diff
@@ -1118,7 +1118,7 @@ namespace cling {
     DynamicLibraryManager* DLM = getDynamicLibraryManager();
     std::string canonicalLib = DLM->lookupLibrary(filename);
     if (allowSharedLib && !canonicalLib.empty()) {
-      switch (DLM->loadLibrary(filename, /*permanent*/false)) {
+      switch (DLM->loadLibrary(canonicalLib, /*permanent*/false, /*resolved*/true)) {
```
vgvassilev: >80 cols?

lib/Interpreter/TransactionUnloader.cpp
```diff
@@ -1015,6 +1015,11 @@ namespace clang {
       Successful &= VisitRedeclarable(InjectedRD, InjectedRD->getDeclContext());
     }
 
+    // clang caches type_info decl. So if we're unloading it's cache version
+    // better let them know.
+    if (RD==m_Sema->CXXTypeInfoDecl)
```
vgvassilev: Could you add spaces before and after ==? Could you add a test case?

vgvassilev:
Thanks a lot for the work. It is very useful to us and if you could address the minor comments I'd be happy to merge this contribution!

marsupial:
Sorry about all the spaces, not my style!
Are you sure you want the test additions squashed into the fixes?
 It'll rewrite the branch history so I'll have to issue a fresh request.
 Isn't it better to have a state where the test is failing in the history?

vgvassilev:
@marsupial no worries I will we had something like clang-format for cling and ROOT ;).

Yes please squash the test into the fixes. Don't worry if you rewrite the branch history. You can do it in the same branch and then when pushing you should push with -f. Github is smart enough to handle the rest: it will update this branch just fine.

I'd prefer to follow clang/llvm standards and add the test with the fix. In my experience, it is much better  when browsing history to see the commits along with the fix: this way one knows what exactly failed, when and who fixed it how.

marsupial:
Ok, didn't know git hub would be smart enough to understand...but I also have some branches coming off with non-trivial history that would be effected. Unless you know a trick to rebase a branch and keep its merges in tact could we compromise and I'll squash the last two and keep the first (5b5f121) separate?

As for adding the two tests (objc, and new Interpreter::load calls):
The next pull-request I'll have changes the interface to pass FileEntries instead of strings, could the tests be added to that?
Can't add a test for obj-c until a lot of other stuff gets merged, you'll have to trust me it was crashing without the fixes.

karies:
I pushed the llvm upgrade to cling - could you rebase your changes to the current master?

lib/Interpreter/Transaction.cpp
```diff
@@ -67,11 +67,12 @@ namespace cling {
     for (auto I = decls_begin(), E = decls_end(); I < E; ++I) {
       for (auto DI : I->m_DGR) {
         if (DI->getKind() == Decl::LinkageSpec) {
-          for ( Decl *DI : LinkageSpecDecl::castToDeclContext(
-                static_cast<LinkageSpecDecl*>(DI))->decls()) {
-            if (NamedDecl* ND = dyn_cast<NamedDecl>(DI)) {
-              if (name.equals(ND->getNameAsString()))
-                return ND;
+          if (LinkageSpecDecl* LSD = cast<LinkageSpecDecl>(DI)) {
```
vgvassilev: This won't work: use `if (LinkageSpecDecl* LSD = dyn_cast<LinkageSpecDecl>(DI))`. The dyn_cast is not dynamic it is a sort of static cast.

lib/Interpreter/Transaction.cpp
```diff
@@ -67,11 +67,12 @@ namespace cling {
...
+            for ( Decl *DI : LSD->decls()) {
```
vgvassilev: spaces.

lib/Interpreter/Transaction.cpp
marsupial: We've already checked DI->getKind() == Decl::LinkageSpec, proper usage according to [http://llvm.org/docs/ProgrammersManual.html#the-isa-cast-and-dyn-cast-templates](url)

marsupial:
As you're both online now, chance I could get an answer from my comment an hour ago about keeping 5b5f121 a separate commit before going into rebase hell?
Thanks.

lib/Interpreter/Transaction.cpp
vgvassilev: Remove the first check then. and make only one if.

lib/Interpreter/Interpreter.cpp
```diff
@@ -1130,7 +1134,12 @@ namespace cling {
         return kFailure;
       }
     }
+    return kMoreInputExpected;
+  }
 
+  Interpreter::CompilationResult
+  Interpreter::loadHeader(const std::string& filename,
+                        Transaction** T /*= 0*/) {
```
vgvassilev: Indent.

vgvassilev:
@marsupial about the commit you are talking about: I am fine with that if it doesn't become a practice. Could you submit all the other changes (that you are referring to) in another PR so that we have the big picture?

marsupial:
I've added a comment that hopefully clarifies it, but if not:

The bug was not handling an obj-c pointer type, that was fixed, and you suggested isAnyPointerType which is probably better. The prior code was:
```
else if (a || b || c) {
 if (a)
 else if (b)
 else if  (c)
 else
  failure
}
```
Now its doing:
```
else {
 storedArgs
 if (a) pushArg
 else if (b) pushArg
 else if  (c) pushArg
if (argsNow > storedArgs) {
}
else
  failure
```
Which I think is more efficient (not re-comparing known information), and more maintainable as you only have to add a type check in one place.

lib/Interpreter/Transaction.cpp
marsupial: But the first check is to avoid diving of into the type-system on every decl (which will likely be many).  I think it's a bit more efficient this way, but if you think otherwise I'll concede.

lib/Interpreter/IncrementalParser.cpp
marsupial: Yes, that is the issue, but code later is prepared to handle that. The problem is llvm:: dyn_cast cannot take a null value.
If PP.getMacroDirective(II) returns dyn_cast_or_null just propagates it to MD, which is checked for NULL on the next line.

vgvassilev: On 15/06/16 19:06, marsupial wrote:
>
> In lib/Interpreter/IncrementalParser.cpp 
> <https://github.com/root-mirror/cling/pull/17#discussion_r67203107>:
>
> > @@ -125,7 +125,7 @@ namespace {
> >    if (!II)
> >      return;
> >    const clang::DefMacroDirective* MD
> > -    = llvm::dyn_cast<clang::DefMacroDirective>(PP.getMacroDirective(II));
> > +    = llvm::dyn_cast_or_null<clang::DefMacroDirective>(PP.getMacroDirective(II));
>
> Yes, that is the issue, but code later is prepared to handle that. The 
> problem is llvm:: dyn_cast cannot take a null value.
> If PP.getMacroDirective(II) returns dyn_cast_or_null just propagates 
> it to MD, which is checked for NULL on the next line.
>
I think this is hiding the real issue. PP.getMacroDirective(II) must not 
return null in this case. Could you give us the setup in which that 
happened?

lib/Interpreter/Transaction.cpp
vgvassilev: On 15/06/16 18:58, marsupial wrote:
>
> In lib/Interpreter/Transaction.cpp 
> <https://github.com/root-mirror/cling/pull/17#discussion_r67201660>:
>
> > @@ -67,11 +67,12 @@ namespace cling {
> >      for (auto I = decls_begin(), E = decls_end(); I < E; ++I) {
> >        for (auto DI : I->m_DGR) {
> >          if (DI->getKind() == Decl::LinkageSpec) {
> > -          for ( Decl *DI : LinkageSpecDecl::castToDeclContext(
> > -                static_cast<LinkageSpecDecl*>(DI))->decls()) {
> > -            if (NamedDecl* ND = dyn_cast<NamedDecl>(DI)) {
> > -              if (name.equals(ND->getNameAsString()))
> > -                return ND;
> > +          if (LinkageSpecDecl* LSD = cast<LinkageSpecDecl>(DI)) {
>
> But the first check is to avoid diving of into the type-system on 
> every decl (which will likely be many). I think it's a bit more 
> efficient this way, but if you think otherwise I'll concede.
>
I prefer the dyn_cast to go in the place of the DI->getKind() and then 
the if stmt containing the cast will go away.

vgvassilev:
On 15/06/16 18:53, marsupial wrote:
>
> I've added a comment that hopefully clarifies it, but if not:
>
> The bug was not handling an obj-c pointer type, that was fixed, and 
> you suggested isAnyPointerType which is probably better. The prior 
> code was:
>
And you got an assert because the obj-c pointer type wasn't handled, 
right? If so this is the intended behavior of the code. I'd like to keep 
it like that, because it exposes potential issues (by crashing in debug 
mode), instead of hiding them.

lib/Interpreter/IncrementalParser.cpp
marsupial: cling -x c
Really the code is already set up to handle CLING_CXXABIS not being defined (see EarlyReturnWarn struct in the same function)

marsupial:
While not printing unknown values is smart, there's no mechanism in place to report that and a hard crash seems a bit over-reaching...The obj-c pointer is in fact a pointer, and will print fine as such, so why not just do it?

This is an issue as I was getting cling to compile/run Objective-C++ code.
I'll back it out if needed, but I'm hoping to merge that as well, so it'd just come back anyway.

marsupial:
I think I've addressed, whats outstanding is:
 llvm::dyn_cast_or_null<clang::DefMacroDirective>(PP.getMacroDirective(II));
 Printing of unknown types that are known to be pointers (really only ObjC pointer)
 Using DI->getKind() == Decl::LinkageSpec instead of going through type-system on every Decl.

The first is definitely a bug in cling as it knows CLING_CXXABIS might not be defined yet would pass a NULL value to llvm::dyn_cast which is illegal.
Let me know how to proceed on the last two.

Thanks.

lib/Interpreter/IncrementalParser.cpp
vgvassilev: @karies is the expert here. Axel, what would be the right thing to do if we run in non-c++ mode here?

vgvassilev:
On 15/06/16 19:57, marsupial wrote:
>
> While not printing unknown values is smart, there's no mechanism in 
> place to report that and a hard crash seems a bit over-reaching...The 
> obj-c pointer is in fact a pointer, and will print fine as such, so 
> why not just do it?
>
I thought you fixed that. So we print now *all* pointer types. I'd keep 
the rest for now.
>
> This is an issue as I was getting cling to compile/run Objective-C++ code.
> I'll back it out if needed, but I'm hoping to merge that as well, so 
> it'd just come back anyway.
>
Overall, it seems good. Could you address the rest of my comments. Also, 
could you rebase your branch as @karies suggested?

vgvassilev:
On 15/06/16 20:04, marsupial wrote:
>
> I think I've addressed, whats outstanding is:
> llvm::dyn_cast_or_nullclang::DefMacroDirective(PP.getMacroDirective(II));
>
This is waiting for Axel's verdict.
>
> Printing of unknown types that are known to be pointers (really only 
> ObjC pointer)
>
I think that is solved, with the comments I had.
>
> Using DI->getKind() == Decl::LinkageSpec instead of going through 
> type-system on every Decl.
>
I think I replied to that. The dyn_cast pattern is used all around 
clang/llvm codebase in exactly that manner.

marsupial:
Must I rebase to latest or just merging it in is acceptable?

vgvassilev:
Please rebase.

lib/Interpreter/IncrementalParser.cpp
karies: @marsupial is suggesting the correct fix. We need a cling test with -x c :-)

marsupial:
Are you sure!? While I appreciate keeping the tree looking clean, rebasing destroys history.
There's no record of this branch every being in a functional/tested state once it's done.

karies:
@marsupial - that doesn't make sense, rebasing rewrites the history but doesn't destroy it. We're not asking you to squash this, but to adjust your branch to the /current/ master.

If you feel you cannot do that then we can take care of it. Please let us know!

marsupial:
If it has to be done, I shall but I'd like to explain my hesitance a bit.

While it doesn't remove the branch's history, it does remove one important piece of information...what the changes are based on.  There's more than 50 commits I'm rebasing too, any one of which could have done something which affects the behavior of my commits. Additionally any branches I have going on get all screwed up (not your problem but definitely a headache for me)

I appreciate a good rebase, as long as it's done in tandem with the changes occurring to master. Sadly that wasn't the case here.

vgvassilev:
On 15/06/16 22:10, marsupial wrote:
>
> For example: Part of my commits handle extern "C" linkage better. LLVM 
> used to have Sema::LocallyScopedExternCDecls to inspect these 
> declarations but it is gone now.
>
We upgraded our supported llvm/clang revision.
>
> If I do a rebase anyone who checks out 'CERN/master' between your 
> commit and the commit merged from my branch is in not only an untested 
> state, but one that is not even compilable.
>
> If I do a merge, our two parallel histories in time are recorded, and 
> one can only check out a version known to compile and has been tested, 
> either from your branch or mine.
>
This is our preferred policy, and we are fine to to do it this way. I 
think all your changes (and the one that you will submit) make sense and 
they will be accepted after a review. I wouldn't worry your repo would 
be out of sync. Please go ahead rebase, fix the issues coming from the 
new llvm/clang API changes and submit the rest of your local changes. I 
will review them asap and get the in mainline.

marsupial:
Ok. Seems it's gonna take some time though.

Out of curiosity I have to ask why this is the preferred way. Seems like your discarding one of the best features of git (merge/branch tracking).

vgvassilev:
On 15/06/16 22:41, marsupial wrote:
>
> Ok. Seems it's gonna take some time though.
>
Thanks.
>
> Out of curiosity I have to ask why this is the preferred way. Seems 
> like your discarding one of the best features of git (merge/branch 
> tracking).
>
We follow the llvm's svn approach of incremental development on the tip 
of the tree. This feature is not very important to me and it seems not 
very important to our (small) dev community. I don't see a lot of value 
to track when, what and how exactly happened in history at the cost of 
difficulties to understand the real changes (having to follow 
complicated branches and merges). This might be helpful for source code 
archeologists but in practice never happens in our context. Interesting 
discussion can be found here: 
http://lists.llvm.org/pipermail/llvm-dev/2016-May/100310.html

karies:
@marsupial - interesting view on git branches, though! But indeed and as Vassil points out, we prefer the reduced complexity over that additional time tracking info (esp as we don't always use branches, making "back then" not well-defined).

Thanks for accepting the rebase challenge! :-)

lib/Interpreter/TransactionUnloader.cpp
```diff
@@ -634,49 +699,12 @@ namespace clang {
...
-              Pos->second.remove(ND);
-        }
-        if (Pos->second.isNull() ||
-            (Pos->second.getAsVector() && !Pos->second.getAsVector()->size()))
-          Map->erase(Pos);
-      }
+    // DeclContexts like EnumDecls don't have lookup maps.
+    if (StoredDeclsMap *Map = DC->getPrimaryContext()->getLookupPtr()) {
+      eraseDeclFromMap(Map, ND);
+      checkDeclIsGone(Map, ND);  // Only #ifndef NDEBUG
```
vgvassilev: Can you wrap the function invocation here in #ifndef NDEBUG and remove it from checkDeclIsGone's definition. This makes it more clear and saves one call.

test/CodeUnloading/CGlobalDecl.C
```diff
@@ -0,0 +1,43 @@
...
+#undef _END_DECL
+
+#define BEGIN_DECL_  extern "C" {
+#define _END_DECL    }
+#include "CGlobalDecl.h"
+.undo
+#include "CGlobalDecl.h"
+
+printf("C globals included and undone\n");
+// CHECK: C globals included and undone
```
vgvassilev: This will always work, because the string "C globals included and undone" exists as part of printf already. So even if it was not printed the test will return still ok, won't it?

test/CodeUnloading/TypeInfo.C
```diff
@@ -0,0 +1,27 @@
...
+//------------------------------------------------------------------------------
+
+// RUN: cat %s | %cling -I%S -Xclang -verify  | FileCheck %s
+// Test unloadTypeInfo
+
+extern "C" int printf(const char*, ...);
+int   A, B;
+
+#include <typeinfo>
+printf("Types match: %d\n", typeid(A)==typeid(B));
```
vgvassilev: spaces before and after ==

test/CodeUnloading/TypeInfo.C
```diff
@@ -0,0 +1,27 @@
...
+// CHECK: Types match: 1
+
+.undo  // undo print
+.undo  // undo #include <typeinfo>
+
+#include <typeinfo>
+printf("Types match: %d\n", typeid(A)==typeid(B));
```
vgvassilev: Same here.

test/Driver/CurrentDir.C
```diff
@@ -0,0 +1,33 @@
...
+void CurrentDir() {
+  char thisDir[1024];
+  getcwd(thisDir, sizeof(thisDir));
+  /*
+    This would be nice, but doesn't work
+    printf("We are here: %s\n", thisDir);
+    -CHECK: We are here: %S
+  */
+
+  printf("Script ran\n");
```
vgvassilev: Same comments here. It will always succeed.

test/CodeUnloading/CGlobalDecl.C
```diff
@@ -0,0 +1,43 @@
...
+// LICENSE.TXT for details.
+//------------------------------------------------------------------------------
+
+// RUN: cat %s | %cling -I%S -Xclang -verify  | FileCheck %s
+// Test externC_GlobalDecl
+
+extern "C" int printf(const char*,...);
+
+printf("Starting C++ globals check\n");
+// CHECK: Starting C++ globals check
```
vgvassilev: Same issue here. Does it fail if printf doesn't print. Could you rely on the value printing instead. Eg.
```
int sanity = 15
// CHECK: (int) 15
```

test/CodeUnloading/TypeInfo.C
```diff
@@ -0,0 +1,27 @@
...
+// This file is dual-licensed: you can choose to license it under the University
+// of Illinois Open Source License or the GNU Lesser General Public License. See
+// LICENSE.TXT for details.
+//------------------------------------------------------------------------------
+
+// RUN: cat %s | %cling -I%S -Xclang -verify  | FileCheck %s
+// Test unloadTypeInfo
+
+extern "C" int printf(const char*, ...);
+int   A, B;
```
vgvassilev: extra spaces

lib/Interpreter/ValueExtractionSynthesizer.cpp
```diff
@@ -371,11 +369,17 @@ namespace {
         // case, because of the overload resolution.
         CallArgs.push_back(E);
       }
-      Call = m_Sema->ActOnCallExpr(/*Scope*/0, m_UnresolvedNoAlloc,
+
+      // Test CallArgs.size to make sure an additional argument (the value)
+      // has been pushed on, if not than we didn't know how to handle the type
+      if (CallArgs.size() > nArgs) {
```
vgvassilev: Could you remove the if stmt? It will still crash without even giving a nice diagnostics before upon unsupported type.

test/CodeUnloading/CGlobalDecl.C
```diff
@@ -0,0 +1,43 @@
...
+
+#define BEGIN_DECL_
+#define _END_DECL
+#include "CGlobalDecl.h"
+.undo
+#include "CGlobalDecl.h"
+.undo
+
+printf("C++ globals included and undone\n");
```
vgvassilev: see other comments.

test/CodeUnloading/CGlobalDecl.C
```diff
@@ -0,0 +1,43 @@
...
+// CHECK: C++ globals included and undone
+
+printf("Starting C globals check\n");
```
vgvassilev: see other comments

vgvassilev:
@marsupial Could you address the *last* round of comments?

test/CodeUnloading/CGlobalDecl.C
marsupial: Don't get it? But I'm not an expert in lit.

Problem was that after the undo and second #include "CGlobalDecl.h" cling would spew errors about redefinition even though the transaction was supposedly done (so the test would fail).

test/Driver/CurrentDir.C
marsupial: I't originally failed, and now it doesn't so I don't see how it will always succeed?
The failure was CurrentDir never being called after the cd %S && cling CurrentDir.C

lib/Interpreter/ValueExtractionSynthesizer.cpp
marsupial: Do you really think crashing cling because a user tried tried to evaluate a return value it doesn't handle is the right thing to do. It seems very abrasive. Why not just:
```
else
  llvm::errs() << "Unkown type "
```

marsupial:
I'm not really getting all the test will always pass comments . They all didn't pass previously.
Again I'm not an expert in lit & FileCheck.
The one thing I don't want to do is print values via cling ala:
```
 int sanity = 15
// CHECK: (int) 15
```

This insert's a ton of declarations inserted into cling (via #include<string>, #include <type_info>) when thats done and makes the undo checking impossible and it much harder to debug the declaration state upon failure.

marsupial:
I'm also not understanding what your asking about lib/Interpreter/ValueExtractionSynthesizer.cpp. There was never a nice diagnostic, only an assert (which never went away). I don't understand why you would now want to always call ActOnCallExpr (which it never did). The arguments passed to  ActOnCallExpr is part of a contract that shouldn't be broken. 

I think is what you're asking for:
```
void someExternFunc(int,int,int);
((void)(*)(int,int) someExternFunc)(arg0, arg1)
```


Again:

```
else if (a || b || c) {
 if (a)
 else if (b)
 else if  (c)
 else
  assert()
}
```
Now its doing:

```
else {
 storedArgs
 if (a) pushArg
 else if (b) pushArg
 else if  (c || d) pushArg

 if (argsNow > storedArgs) {
 }
 else
   assert()
}
```

test/CodeUnloading/CGlobalDecl.C
vgvassilev: On 16/06/16 18:44, marsupial wrote:
>
> In test/CodeUnloading/CGlobalDecl.C 
> <https://github.com/root-mirror/cling/pull/17#discussion_r67380985>:
>
> > +// CHECK: C++ globals included and undone
> > +
> > +printf("Starting C globals check\n");
> > +// CHECK: Starting C globals check
> > +
> > +#undef BEGIN_DECL_
> > +#undef _END_DECL
> > +
> > +#define BEGIN_DECL_  extern "C" {
> > +#define _END_DECL    }
> > +#include "CGlobalDecl.h"
> > +.undo
> > +#include "CGlobalDecl.h"
> > +
> > +printf("C globals included and undone\n");
> > +// CHECK: C globals included and undone
>
> Don't get it? But I'm not an expert in lit.
>
> Problem was that after the undo and second #include "CGlobalDecl.h" 
> cling would spew errors about redefinition even though the transaction 
> was supposedly done (so the test would fail).
>
In this case use //expected-no-diagnostics in the test-case. Please have 
a look at cling's test suite.

vgvassilev:
On 16/06/16 19:00, marsupial wrote:
>
> I'm not really getting all the test will always pass comments . They 
> all didn't pass previously.
> Again I'm not an expert in lit & FileCheck.
> The one thing I don't want to do is print values via cling ala:
>
> |int sanity = 15 // CHECK: (int) 15 |
>
> This insert's a ton of declarations inserted into cling (via #include, 
> #include ) when thats done and makes the undo checking impossible and 
> it much harder to debug the declaration state upon failure.
>
Ah, that is a good point. I check and I think my previous comment about 
printf's is invalid. So please ignore these, sorry.

vgvassilev:
On 16/06/16 20:05, marsupial wrote:
>
> I'm also not understanding what your asking about 
> lib/Interpreter/ValueExtractionSynthesizer.cpp. There was never a nice 
> diagnostic, only an assert (which never went away). I don't understand 
> why you would now want to always call ActOnCallExpr (which it never 
> did). The arguments passed to ActOnCallExpr is part of a contract that 
> shouldn't be broken.
>
> I think is what you're asking for:
>
> |void someExternFunc(int,int,int); ((void)(*)(int,int) 
> someExternFunc)(arg0, arg1) |
>
> Again:
>
> |else if (a || b || c) { if (a) else if (b) else if (c) else assert() } |
>
> Now its doing:
>
> |else { storedArgs if (a) pushArg else if (b) pushArg else if (c || d) 
> pushArg if (argsNow > storedArgs) { }|
|If you remove this and leave the "actOnCallExpr" will produce a 
diagnostic if no argument is pushed. I want people to see this if there 
is something unhandled and complain. I am okay if you remove this if-stmt.|

marsupial:
I don't want to belabor this but want to be clear on what happens if an argument is not pushed

> warning: expression result unused; should this cast be to 'void'? [-Wunused-value]

Is the diagnostic printed, but actOnCallExpr reutrns a valid address, that address gets called, sets up the registers and calls:

> cling::runtime::internal::setValueNoAlloc(a, b, c, d, e)

But the argument e hasn't been pushed, so I think it's undefined undefined.
Am I missing something or is this the behavior you are intending?

vgvassilev:
On 16/06/16 22:53, marsupial wrote:
>
> I don't want to belabor this but want to be clear on what happens if 
> an argument is not pushed
>
>     warning: expression result unused; should this cast be to 'void'?
>     [-Wunused-value]
>
> Is the diagnostic printed, but actOnCallExpr reutrns a valid address, 
> that address gets called, sets up the registers and calls:
>
>     cling::runtime::internal::setValueNoAlloc(a, b, c, d, e)
>
> But the argument e hasn't been pushed, so I think it's undefined 
> undefined.
>
Good point. What about keeping the assert for the debug version and 
adding a macro expansion for release transforming all asserts into 
llvm::errs()?
>
> Am I missing something or is this the behavior you are intending?
>
In general, I want cling to do only what it is designed for. IMO it 
should handle only the foreseen use cases and in any other case it 
should fail miserably (at least in debug mode). This forces users to 
request "proper" fixes in the code (I realize how harsh this might 
sound). This has always been the overall development policy that we've 
been following especially in the complex code and situations such as 
this. Otherwise, if we become more permissive many of our users consider 
bugs as features and our internal backward-compatibility policies make 
such cases very complicated to fix.

To give you an example: if your the objcptr almost worked (say printed 
the void *&, instead of void**) you wouldn't invest time in providing a 
real fix (or wouldn't even complain). Even worse, some of our 
shareholders would consider that as an intentional feature and if one 
changes it into proper void** printout we could end up with request to 
revert to the "old and well working" behavior.

lib/Interpreter/IncrementalParser.cpp
vgvassilev: On 15/06/16 21:06, Axel Naumann wrote:
>
> In lib/Interpreter/IncrementalParser.cpp 
> <https://github.com/root-mirror/cling/pull/17#discussion_r67225773>:
>
> > @@ -125,7 +125,7 @@ namespace {
> >    if (!II)
> >      return;
> >    const clang::DefMacroDirective* MD
> > -    = llvm::dyn_cast<clang::DefMacroDirective>(PP.getMacroDirective(II));
> > +    = llvm::dyn_cast_or_null<clang::DefMacroDirective>(PP.getMacroDirective(II));
>
> @marsupial <https://github.com/marsupial> is suggesting the correct 
> fix. We need a cling test with -x c :-)
>
Could you provide a test with this fix?

marsupial:
I agree with parts, but think an assert is pretty harsh. Either way it's in there as well as far more graceful behavior in release.

Tested and working, but any test I write would be pretty specific to my setup and probably fail elsewhere.

karies:
@marsupial - FYI, I'm generally representing your voice ("let's not abort a production process just because we are confused!") in discussions between Vassil and me ;-) So that point of view of yours is generally represented in cling development. But I still appreciate Vassil's strive for correctness! I like the compromise you found here.

vgvassilev:
On 17/06/16 00:12, marsupial wrote:
>
> I agree with parts, but think an assert is pretty harsh. Either way 
> it's in there as well as far more graceful behavior in release.
>
> Tested and working, but any test I write would be pretty specific to 
> my setup and probably fail elsewhere.
>
Could you be more specific. I am afraid that without tests it would be 
difficult to justify patches. I prefer external contributions to be 
well-tested because once they go in the master they will be maintained 
by us. Could you give me an example which cannot be translated into a 
lit test case?

marsupial:
I don't know the set of all types that can't be handled, I only knew one and I fixed it...
It's very hard to create an undefined type to test, and all my attempts have failed.

I think it might be enough to say the contract with SynthesizeSVRInit is that returning NULL is acceptable, but one is responsible for reporting the errors that caused that.

marsupial:
Well actually I found one more but it would be dependent on compiling with objective c support and I think only work on OS X, so I thought it useless to put it in.

And even if there were a test, per your instruction's it will fail in debug mode as the assert is triggered.

marsupial:
I found one that will work. But the assert means the test will abort in debug mode how is this handled dealt with?

vgvassilev:
On 17/06/16 14:19, marsupial wrote:
>
> I found one that will work. But the assert means the test will abort 
> in debug mode how is this handled dealt with?
>
Can I see it before answering, because I am not sure I completely 
understand.

marsupial:
It's in the last commit: [https://github.com/root-mirror/cling/pull/17/commits/2c0c174d19a31f3f5ec3c082fc52bd002ee63c90](url)

You asked for the behavior where it crashes in debug mode, but not in release.
As such any test run in debug mode will also crash.

vgvassilev:
On 17/06/16 16:27, marsupial wrote:
>
> It's in the last commit: 
> https://github.com/root-mirror/cling/pull/17/commits/2c0c174d19a31f3f5ec3c082fc52bd002ee63c90 
> <url>
>
> You asked for the behavior where it crashes in debug mode, but not in 
> release.
> As such any test run in debug mode will also crash.
>
Ok, I understand now. I think the explicit diagnostic is enough. Please 
remove the assert, I agree it better without now.

marsupial:
Curious the status of this and whether I should keep pushing additional change sets?

vgvassilev:
It looks good to me. We are waiting a green light by our buildbots after the LLVM upgrade and I can merge this in. It can take a week or two. Please go ahead and submit the other set of changes for review.

karies:
@marsupial - there is unbelievable goodness in here. Wow. Thanks *so* much for your contributions! I'll start reviewing your second set of commits. I expect that we will merge your first set of changes in independently. Again - thank you!!!

karies:
Do you think there is a chance that you can split add724b... into smaller PRs? It's a bit difficult (almost impossible) to review such a huge change set on github. If possible, please keep one PR per feature / feature class / entity that you see as a homogenous set of changes / you know what I mean :-)

(E.g. add724b LGTM but I don't know how to signal that; I can only add line notes for the overall set of changes.)

If this is too much work then I'll try to help you with that by picking the most interesting ones first (personal bias warning ;-) into separate PRs.

karies:
e67cc5d30fc2cf2549276993be16bca1b6057b46 is in (thanks!); now I need to figure out which are the other patches we have reviewed already... not easy :-) I'd really appreciate more focused PRs in the future - many PRs with few commits is a lot simpler for me than the inverse. And in general these patches are easier to integrate if you work in feature branches that I can pick up - it might have allowed me to pinpoint which commits we had reviewed...

I'll try to find the relevant commits and apply them.

karies:
I see the ones we already reviewed are in random positions in this PR's chain of commits :-( So I'll go one by one. 244b26d6021e55e214ee078bec99e7c6a43d4820 is merged, thanks!

marsupial:
I agree, and was working on separate branches as I went on, but over the month I've gotten to a point where I already merged them as I wanted to be certain everything was cooperating. I couldn't get rebasing to reliably work preserving the merges so I had to just move it all into a mainline.

That said, I have no problem re-submitting them as smaller requests, but before doing that:

 Some change-sets are seemingly unrelated but the order of commits is important. For example fixing the unloading of the runtime printer depends on a lot of commits prior in order for the fix to actually work, because <string> has to be unloaded and reloaded properly for it to be of any use.

 With what granularity? My initial thought would be these branches:
  Small fixes,  that should be one-liners/non-controversial.
  Fixes to handle failing initialization/startup.
  Improvements to DeclUnloader
  Improvements to InputValidator
  Refactor of MetaCommands system.
  OS X specific additions (.F command and objective-c).
  Changes to get cling and clang to run from the same installation/libraries
  
Also certain changes needed to occur in clang for some fixes (the message will have 'Requires changes to clang'). I'll try to upload the clang repo fork today.

karies:
Re granularity: in general, a single PR should contain only the minimal meaningful, standalone set of commits.

Now here I don't think that's a reasonable goal; instead we need a compromise between "reviewability" and "separation work". So indeed your set of PRs makes sense. Maybe you don't send them all at once, but you force the change dependencies by sequencing the PRs?

Look - in the end we're all reasonable people who like to avoid silly work. Any help from your side in making your wonderful commits more accessible for us is appreciated; in the end Vassil and I might have to invest some work, too, and that's just fine: it's worth is. So yes, please go ahead as you suggested, and if in doubt, submit things in an separate PR, and hold back with subsequent / dependent PRs until the prior PR is merged.

Thanks for bearing with us!

lib/UserInterface/textinput/Callbacks.h
```diff
@@ -30,7 +30,7 @@ namespace textinput {
                           EditorRange& R /*out*/,
                           std::vector<std::string>& DisplayCompletions /*out*/)
     = 0;
-    virtual ~TabCompletion();
+    virtual ~TabCompletion() {}
```
vgvassilev: Why did you change this? This way the vtable symbols will pollute all .o files which includes Callbacks.h instead of Editor.o.

marsupial: ExceptionRTTI.cpp throws exceptions and it's built with RTTI.
That means anything that want's to catch what it throws must also be built with RTTI.
Linux seems to not care but both OS X and Windows do.

So when building UserInterface.cpp with RTTI, unless it sees ~TabCompletion, it will try to link to it, but ~TabCompletion was not built with RTTI so link will fail.

I've worked out a better solution, where users call into a RunLoop in ExceptionRTTI.cpp.
The RunLoop will solve any problems for users of cling that don't have exception handling as well.
But that change is much further down the road.

lib/Interpreter/ValuePrinter.cpp
```diff
@@ -127,6 +127,7 @@ struct AccessCtrlRAII_t {
   AccessCtrlRAII_t(cling::Interpreter& Interp):
     LangOpts(const_cast<clang::LangOptions&>(Interp.getCI()->getLangOpts())) {
     savedAccessControl = LangOpts.AccessControl;
+    LangOpts.AccessControl = false;
```
vgvassilev: Could you come up with a test for this?

marsupial: There already is, the other half of the commit exposed the bug that was obscuring the failure.

lib/UserInterface/textinput/Callbacks.h
marsupial: I also have a hunch it will fix PowerPC exceptions failing:
https://github.com/root-mirror/cling/commit/494733cef80998a12c28a8f1475dfcb7cc250dd0

vgvassilev:
Could you rebase this PR, please?

vgvassilev:
@marsupial could you rebase this PR from time to time to see what we are left with?

marsupial:
Updated to point to remaining fixes to Windows.
Changes to COFF format in LLVM will allow the remaining 4 failing tests to pass.

bellenot:
So this gives 7 failing tests in Release mode and 20 in Debug mode...

marsupial:
According to Appveyor it is 4.
Are the additional 3 also related to NullDeref protection (or 1. below)?

You are aware there are major bugs in the LLVM that cling is being built with that will make Windows completely unstable until addressed?

1. Symbol lookup is completely unordered/undefined: https://reviews.llvm.org/D30107
2. Using COFF is missing a necessary relocation: https://reviews.llvm.org/D30709
3. Using ELF requires an __ImageBase variable during code-gen.

bellenot:
I'm trying to understand the differences between my setups and the Appveyor one. And here are the results I got. In Release mode: 
Failing Tests (7):
    Cling :: NullDeref/ExecutionTermination.C
    Cling :: NullDeref/If.C
    Cling :: NullDeref/Load.C
    Cling :: NullDeref/MethodCalls.C
    Cling :: Pragmas/add_env_path.C
    Cling :: Prompt/OutputRedirect.C
    Cling :: Utils/Transform.C

And in Debug mode:
Failing Tests (20):
    Cling :: CodeGeneration/TLSVars.C
    Cling :: CodeUnloading/AtExit.C
    Cling :: CodeUnloading/RereadFile.C
    Cling :: Extensions/Lookup/ControlFlow.C
    Cling :: Extensions/Lookup/LifetimeHandler.C
    Cling :: Interfaces/compileFunc.C
    Cling :: Interfaces/evaluate.C
    Cling :: Lookup/func.C
    Cling :: Lookup/scope.C
    Cling :: Lookup/template.C
    Cling :: Lookup/type.C
    Cling :: NullDeref/MethodCalls.C
    Cling :: Pragmas/add_env_path.C
    Cling :: Prompt/Continuation.C
    Cling :: Prompt/DontWrap.C
    Cling :: Prompt/OutputRedirect.C
    Cling :: Prompt/Regression.C
    Cling :: Prompt/globalinit.C
    Cling :: Prompt/initorder.C
    Cling :: Utils/Transform.C

And no, sorry, but I'm not fully aware of all the bugs in LLVM/Clang that are Windows related and making cling unstable. So thanks for the info and the links!

vgvassilev:
@marsupial, about 1. if you want me to review this patch please follow the comments. I am not really sure how else we could help there. I doubt somebody will look at the second patch without adding a test case...

marsupial:
I believe all the NullDeref failures are related to the missing relocation (2).
add_env_path.C & OutputRedirect.C because of unordered sym lookup (1).
Transform.C should be fixed by this (and was here & on Appveyor)

bellenot:
So applying the diff (for RuntimeDyldCOFFX86_64.h) decreases the failing test.
In Release mode:
Failing Tests (3):
    Cling :: Pragmas/add_env_path.C
    Cling :: Prompt/OutputRedirect.C
    Cling :: Utils/Transform.C
And in Debug mode:
Failing Tests (18):
    Cling :: CodeGeneration/TLSVars.C
    Cling :: CodeUnloading/AtExit.C
    Cling :: CodeUnloading/RereadFile.C
    Cling :: Extensions/Lookup/ControlFlow.C
    Cling :: Extensions/Lookup/LifetimeHandler.C
    Cling :: Interfaces/compileFunc.C
    Cling :: Interfaces/evaluate.C
    Cling :: Lookup/func.C
    Cling :: Lookup/scope.C
    Cling :: Lookup/template.C
    Cling :: Pragmas/add_env_path.C
    Cling :: Prompt/Continuation.C
    Cling :: Prompt/DontWrap.C
    Cling :: Prompt/OutputRedirect.C
    Cling :: Prompt/Regression.C
    Cling :: Prompt/globalinit.C
    Cling :: Prompt/initorder.C
    Cling :: Utils/Transform.C

marsupial:
The DynamicLibrary fixes are also here which will apply cleaner onto your LLVM:
https://github.com/marsupial/llvm/commit/6f1f1774cb17879be804b231b21318ce06bdcc09

PR #21, opened 2016-07-19
Cling UITabCompletion.

CristinaCristescu:


lib/UserInterface/UserInterface.cpp
```diff
@@ -12,6 +12,7 @@
 #include "cling/UserInterface/CompilationException.h"
 #include "cling/Interpreter/Exception.h"
 #include "cling/MetaProcessor/MetaProcessor.h"
+#include "textinput/Callbacks.h"
```
karies: What do we need this for? Can I remove it?

karies:
Merged, thanks!

PR #22, opened 2016-07-19
Code complete only input coming from tty.

CristinaCristescu:
First try.

### DIFF
--- a/lib/UserInterface/textinput/Editor.cpp
+++ b/lib/UserInterface/textinput/Editor.cpp
@@ -20,6 +20,7 @@
 #include "textinput/Callbacks.h"
 #include "textinput/History.h"
 #include "textinput/KeyBinding.h"
+#include "textinput/StreamReaderUnix.h"
 #include "textinput/TextInput.h"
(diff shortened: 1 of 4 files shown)


lib/UserInterface/textinput/Editor.cpp
```diff
@@ -449,6 +450,9 @@ namespace textinput {
         ProcessMove(kMoveEnd, R);
         std::vector<std::string> completions;
         TabCompletion* tc = fContext->GetCompletion();
+        Reader* reader = fContext->GetReaders()[0];
+        StreamReaderUnix* streamReader = (StreamReaderUnix*)(reader);
```
karies: What if the reader is a `StreamReaderWindows`? Can we make `IsFromTTY()` a pure virtual function in `Reader` and implement it for `StreamReaderWindows`, too?

CristinaCristescu: Yes, pushed!
What do we do with the vector of Readers? Is the first one good enough?

karies:
Thanks, (squashed and) merged!

PR #24, opened 2016-07-21
Cling tab comp minimal test.

CristinaCristescu:


karies:
Do we need an update of the PR after the fix to CodeComplete / resetting the Diags?

karies:
Merged! Thanks!

PR #26, opened 2016-07-22
Refactor Interpreter startup and default include paths.

marsupial:
This is the first step in getting clang & cling to run from the same installation.

### DIFF
--- a/include/cling/Interpreter/InvocationOptions.h
+++ b/include/cling/Interpreter/InvocationOptions.h
@@ -14,16 +14,44 @@
 #include <vector>
 
 namespace cling {
+
+  ///\brief Class that stores options that are used by both cling and
+  /// clang::CompilerInvocation.
+  ///
+  class CompilerOptions {
+  public:
+    /// \brief Construct CompilerOptions from given arguments. When argc & argv
+    /// are 0, all arguments are saved into \pRemaining to pass to clang. If argc
+    /// or argv is 0, caller is must fill in \pRemaining with any arguments that
+    /// should be passed to clang.
+    ///
+    CompilerOptions(int argc = 0, const char* const *argv = nullptr);
+
+    ///\brief Parse argc, and argv into our variables.
+    ///
+    ///\param [in] argc - argument count
+    ///\param [in] argv - arguments
+    ///\param [out] Inputs - save all arguments that are inputs/files here
+    ///
+    void Parse(int argc, const char* const argv[],
+               std::vector<std::string>* Inputs = nullptr);
+
+    bool Language;
+    bool ResourceDir;
+    bool SysRoot;
+    bool NoBuiltinInc;
+    bool NoCXXInc;
+    bool StdVersion;
+    bool StdLib;
+
+    ///\brief The remaining arguments to pass to clang.
+    ///
+    std::vector<const char*> Remaining;
+  };
+
   class InvocationOptions {
   public:
-    InvocationOptions():
-      ErrorOut(false), NoLogo(false), ShowVersion(false), Verbose(false),
-      Help(false), MetaString(".") {}
-    bool ErrorOut;
-    bool NoLogo;
-    bool ShowVersion;
-    bool Verbose;
-    bool Help;
+    InvocationOptions(int argc, const char* const argv[]);
 
     /// \brief A line starting with this string is assumed to contain a
     ///        directive for the MetaProcessor. Defaults to "."
@@ -32,12 +60,15 @@ namespace cling {
     std::vector<std::string> LibsToLoad;
     std::vector<std::string> LibSearchPath;
     std::vector<std::string> Inputs;
+    CompilerOptions CompilerOpts;
 
-    static InvocationOptions CreateFromArgs(int argc, const char* const argv[],
-                                            std::vector<unsigned>& leftoverArgs
-                                            /* , Diagnostic &Diags */);
+    bool ErrorOut;
+    bool NoLogo;
+    bool ShowVersion;
+    bool Verbose;
+    bool Help;
 
-    void PrintHelp();
+    static void PrintHelp();
   };
 }
 
--- a/include/cling/Interpreter/InvocationOptions.h
+++ b/include/cling/Interpreter/InvocationOptions.h
@@ -43,6 +43,7 @@ namespace cling {
     bool NoCXXInc;
     bool StdVersion;
     bool StdLib;
+    bool Verbose;
 
     ///\brief The remaining arguments to pass to clang.
     ///
@@ -65,8 +66,8 @@ namespace cling {
     bool ErrorOut;
     bool NoLogo;
     bool ShowVersion;
-    bool Verbose;
     bool Help;
+    bool Verbose() const { return CompilerOpts.Verbose; }
 
     static void PrintHelp();
   };
--- a/include/cling/Utils/Paths.h
+++ b/include/cling/Utils/Paths.h
@@ -39,13 +39,15 @@ namespace cling {
     /// \param [out] Paths - All the paths in the string that exist
     /// \param [in] Mode - If any path doesn't exist stop and return false
     /// \param [in] Delim - The delimeter to use
+    /// \param [in] Verbose - Whether to print out details as 'clang -v' would
     ///
     /// \return true if all paths existed, otherwise false
     ///
     bool SplitPaths(llvm::StringRef PathStr,
                     llvm::SmallVectorImpl<llvm::StringRef>& Paths,
                     SplitMode Mode = kPruneNonExistant,
-                    llvm::StringRef Delim = llvm::StringRef(":"));
+                    llvm::StringRef Delim = llvm::StringRef(":"),
+                    bool Verbose = false);
 
     ///\brief Adds multiple include paths separated by a delimter into the
     /// given HeaderSearchOptions.  This adds the paths but does no further
@@ -59,6 +61,11 @@ namespace cling {
     void AddIncludePaths(llvm::StringRef PathStr,
                          clang::HeaderSearchOptions& Opts,
                          const char* Delim = ":");
+
+    ///\brief Write to llvm::errs that directory does not exist in a format
+    /// matching what 'clang -v' would do
+    ///
+    void LogNonExistantDirectory(llvm::StringRef Path);
 
     ///\brief Copies the current include paths into the HeaderSearchOptions.
     ///
--- a/include/cling/Utils/Paths.h
+++ b/include/cling/Utils/Paths.h
@@ -1,0 +1,37 @@
+//--------------------------------------------------------------------*- C++ -*-
+// CLING - the C++ LLVM-based InterpreterG :)
+// author:
+//
+// This file is dual-licensed: you can choose to license it under the University
+// of Illinois Open Source License or the GNU Lesser General Public License. See
+// LICENSE.TXT for details.
+//------------------------------------------------------------------------------
+
+#ifndef CLING_UTILS_PATHS_H
+#define CLING_UTILS_PATHS_H
+
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/StringRef.h"
+
+namespace cling {
+  namespace utils {
+
+    ///\brief Collect the constituant paths from a PATH string.
+    /// /bin:/usr/bin:/usr/local/bin -> {/bin, /usr/bin, /usr/local/bin}
+    ///
+    /// All paths returned existed at the time of the call
+    /// \param [in] PathStr - The PATH string to be split
+    /// \param [out] Paths - All the paths in the string that exist
+    /// \param [in] EarlyOut - If any path doesn't exist stop and return false
+    /// \param [in] Delim - The delimeter to use
+    ///
+    /// \return true if all paths existed, otherwise false
+    ///
+    bool SplitPaths(llvm::StringRef PathStr,
+                    llvm::SmallVectorImpl<llvm::StringRef>& Paths,
+                    bool EarlyOut = false,
+                    llvm::StringRef Delim = llvm::StringRef(":"));
+  }
+}
+
+#endif // CLING_UTILS_PATHS_H
--- a/include/cling/Utils/Paths.h
+++ b/include/cling/Utils/Paths.h
@@ -12,6 +12,15 @@
 
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringRef.h"
+#include <string>
+
+namespace llvm {
+  class raw_ostream;
+}
+
+namespace clang {
+  class HeaderSearchOptions;
+}
 
 namespace cling {
   namespace utils {
@@ -31,6 +40,37 @@ namespace cling {
                     llvm::SmallVectorImpl<llvm::StringRef>& Paths,
                     bool EarlyOut = false,
                     llvm::StringRef Delim = llvm::StringRef(":"));
+
+    ///\brief Copies the current include paths into the HeaderSearchOptions.
+    ///
+    ///\param[in] Opts - HeaderSearchOptions to read from
+    ///\param[out] Paths - Vector to output elements into
+    ///\param[in] WithSystem - if true, incpaths will also contain system
+    ///       include paths (framework, STL etc).
+    ///\param[in] WithFlags - if true, each element in incpaths will be prefixed
+    ///       with a "-I" or similar, and some entries of incpaths will signal
+    ///       a new include path region (e.g. "-cxx-isystem"). Also, flags
+    ///       defining header search behavior will be included in incpaths, e.g.
+    ///       "-nostdinc".
+    ///
+    void CopyIncludePaths(const clang::HeaderSearchOptions& Opts,
+                          llvm::SmallVectorImpl<std::string>& Paths,
+                          bool WithSystem, bool WithFlags);
+
+    ///\brief Prints the current include paths into the HeaderSearchOptions.
+    ///
+    ///\param[in] Opts - HeaderSearchOptions to read from
+    ///\param[in] Out - Stream to dump to
+    ///\param[in] WithSystem - dump contain system paths (framework, STL etc).
+    ///\param[in] WithFlags - if true, each line will be prefixed
+    ///       with a "-I" or similar, and some entries of incpaths will signal
+    ///       a new include path region (e.g. "-cxx-isystem"). Also, flags
+    ///       defining header search behavior will be included in incpaths, e.g.
+    ///       "-nostdinc".
+    ///
+    void DumpIncludePaths(const clang::HeaderSearchOptions& Opts,
+                          llvm::raw_ostream& Out,
+                          bool WithSystem, bool WithFlags);
   }
 }
 
--- a/include/cling/Utils/Paths.h
+++ b/include/cling/Utils/Paths.h
@@ -25,21 +25,40 @@ namespace clang {
 namespace cling {
   namespace utils {
 
+    enum SplitMode {
+      kPruneNonExistant,  ///< Don't add non-existant paths into output
+      kFailNonExistant,   ///< Fail on any non-existant paths
+      kAllowNonExistant   ///< Add all paths whether they exist or not
+    };
+
     ///\brief Collect the constituant paths from a PATH string.
     /// /bin:/usr/bin:/usr/local/bin -> {/bin, /usr/bin, /usr/local/bin}
     ///
     /// All paths returned existed at the time of the call
     /// \param [in] PathStr - The PATH string to be split
     /// \param [out] Paths - All the paths in the string that exist
-    /// \param [in] EarlyOut - If any path doesn't exist stop and return false
+    /// \param [in] Mode - If any path doesn't exist stop and return false
     /// \param [in] Delim - The delimeter to use
     ///
     /// \return true if all paths existed, otherwise false
     ///
     bool SplitPaths(llvm::StringRef PathStr,
                     llvm::SmallVectorImpl<llvm::StringRef>& Paths,
-                    bool EarlyOut = false,
+                    SplitMode Mode = kPruneNonExistant,
                     llvm::StringRef Delim = llvm::StringRef(":"));
+
+    ///\brief Adds multiple include paths separated by a delimter into the
+    /// given HeaderSearchOptions.  This adds the paths but does no further
+    /// processing. See Interpreter::AddIncludePaths or CIFactory::createCI
+    /// for examples of what needs to be done once the paths have been added.
+    ///
+    ///\param[in] PathsStr - Path(s)
+    ///\param[in] Opts - HeaderSearchOptions to add paths into
+    ///\param[in] Delim - Delimiter to separate paths or NULL if a single path
+    ///
+    void AddIncludePaths(llvm::StringRef PathStr,
+                         clang::HeaderSearchOptions& Opts,
+                         const char* Delim = ":");
 
     ///\brief Copies the current include paths into the HeaderSearchOptions.
     ///
--- a/lib/Interpreter/CIFactory.cpp
+++ b/lib/Interpreter/CIFactory.cpp
@@ -42,9 +42,18 @@
 
 #include <memory>
 
+#ifndef _MSC_VER
+# include <unistd.h>
+# define getcwd_func getcwd
+#endif
+
+// FIXME: This code has been taken (copied from) llvm/tools/clang/lib/Driver/WindowsToolChain.cpp
+// and should probably go to some platform utils place.
+// the code for VS 11.0 and 12.0 common tools (vs110comntools and vs120comntools)
+// has been implemented (added) in getVisualStudioDir()
+#ifdef _MSC_VER
 // Include the necessary headers to interface with the Windows registry and
 // environment.
-#ifdef _MSC_VER
 # define WIN32_LEAN_AND_MEAN
 # define NOGDI
 # ifndef NOMINMAX
@@ -57,18 +66,8 @@
 # define pclose _pclose
 # define getcwd_func _getcwd
 # pragma comment(lib, "Advapi32.lib")
-#else
-# include <unistd.h>
-# define getcwd_func getcwd
-#endif
 
 using namespace clang;
-
-// FIXME: This code has been taken (copied from) llvm/tools/clang/lib/Driver/WindowsToolChain.cpp
-// and should probably go to some platform utils place.
-// the code for VS 11.0 and 12.0 common tools (vs110comntools and vs120comntools)
-// has been implemented (added) in getVisualStudioDir()
-#ifdef _MSC_VER
 
 /// \brief Read registry string.
 /// This also supports a means to look for high-versioned keys by use
@@ -338,95 +337,131 @@ static bool getVisualStudioDir(std::string &path) {
   return false;
 }
 
-#endif // _MSC_VER
+#elif defined(__APPLE__)
+
+#include <dlfcn.h> // dlopen to avoid linking with CoreServices
+#include <CoreServices/CoreServices.h>
+#include <sstream>
+
+static bool getISysRootVersion(const std::string& SDKs, int major,
+                               int minor, std::string& sysRoot) {
+  std::ostringstream os;
+  os << SDKs << "MacOSX" << major << "." << minor << ".sdk";
+
+  std::string SDKv = os.str();
+  if (llvm::sys::fs::is_directory(SDKv)) {
+    sysRoot.swap(SDKv);
+    return true;
+  }
+
+  return false;
+}
+
+static bool getISysRoot(std::string& sysRoot) {
+  using namespace llvm::sys;
+
+  // Some versions of OS X and Server have headers installed
+  if (fs::is_regular_file("/usr/include/stdlib.h"))
+    return false;
+
+  std::string SDKs("/Applications/Xcode.app/Contents/Developer");
+
+  // Is XCode installed where it usually is?
+  if (!fs::is_directory(SDKs)) {
+    // Nope, use xcode-select -p to get the path
+    if (FILE *pf = ::popen("xcode-select -p", "r")) {
+      SDKs.clear();
+      char buffer[512];
+      while (fgets(buffer, sizeof(buffer), pf) && buffer[0])
+        SDKs.append(buffer);
+
+      // remove trailing \n
+      while (!SDKs.empty() && SDKs.back() == '\n')
+        SDKs.resize(SDKs.size() - 1);
+      ::pclose(pf);
+    } else // Nothing more we can do
+      return false;
+  }
+
+  SDKs.append("/Platforms/MacOSX.platform/Developer/SDKs/");
+  if (!fs::is_directory(SDKs))
+    return false;
+
+
+  // Seems to make more sense to get the currently running SDK so any loaded
+  // libraries won't cause conflicts
+
+  // Try to get the SDK for whatever version of OS X is currently running
+  if (void *core = dlopen(
+          "/System/Library/Frameworks/CoreServices.framework/CoreServices",
+          RTLD_LAZY)) {
+    SInt32 majorVersion = -1, minorVersion = -1;
+    typedef ::OSErr (*GestaltProc)(::OSType, ::SInt32 *);
+    if (GestaltProc Gestalt = (GestaltProc)dlsym(core, "Gestalt")) {
+      Gestalt(gestaltSystemVersionMajor, &majorVersion);
+      Gestalt(gestaltSystemVersionMinor, &minorVersion);
+    }
+    ::dlclose(core);
+
+    if (majorVersion != -1 && minorVersion != -1) {
+      if (getISysRootVersion(SDKs, majorVersion, minorVersion, sysRoot))
+        return true;
+    }
+  }
+
+#define GET_ISYSROOT_VER(maj, min) \
+  if (getISysRootVersion(SDKs, maj, min, sysRoot)) \
+    return true;
+
+  // Try to get the SDK for whatever cling was compiled with
+  #if defined(MAC_OS_X_VERSION_10_11)
+    GET_ISYSROOT_VER(10, 11);
+  #elif defined(MAC_OS_X_VERSION_10_10)
+    GET_ISYSROOT_VER(10, 10);
+  #elif defined(MAC_OS_X_VERSION_10_9)
+    GET_ISYSROOT_VER(10, 9);
+  #elif defined(MAC_OS_X_VERSION_10_8)
+    GET_ISYSROOT_VER(10, 8);
+  #elif defined(MAC_OS_X_VERSION_10_7)
+    GET_ISYSROOT_VER(10, 7);
+  #elif defined(MAC_OS_X_VERSION_10_6)
+    GET_ISYSROOT_VER(10, 6);
+  #elif defined(MAC_OS_X_VERSION_10_5)
+    GET_ISYSROOT_VER(10, 5);
+  #elif defined(MAC_OS_X_VERSION_10_4)
+    GET_ISYSROOT_VER(10, 4);
+  #elif defined(MAC_OS_X_VERSION_10_3)
+    GET_ISYSROOT_VER(10, 3);
+  #elif defined(MAC_OS_X_VERSION_10_2)
+    GET_ISYSROOT_VER(10, 2);
+  #elif defined(MAC_OS_X_VERSION_10_1)
+    GET_ISYSROOT_VER(10, 1);
+  #else // MAC_OS_X_VERSION_10_0
+    GET_ISYSROOT_VER(10, 0);
+  #endif
+
+#undef GET_ISYSROOT_VER
+
+  // Nothing left to do but iterate the SDKs directory
+  // copy the paths and then sort for the latest
+  std::error_code ec;
+  std::vector<std::string> srtd;
+  for (fs::directory_iterator it(SDKs, ec), e; it != e; it.increment(ec))
+    srtd.push_back(it->path());
+  if (!srtd.empty()) {
+    std::sort(srtd.begin(), srtd.end(), std::greater<std::string>());
+    sysRoot.swap(srtd[0]);
+    return true;
+  }
+
+  return false;
+}
+
+#endif // __APPLE__, _MSC_VER
+
+using namespace clang;
 
 namespace {
-  static void SetClingCustomLangOpts(LangOptions& Opts) {
-    Opts.EmitAllDecls = 0; // Otherwise if PCH attached will codegen all decls.
-#ifdef _MSC_VER
-    Opts.Exceptions = 0;
-    if (Opts.CPlusPlus) {
-      Opts.CXXExceptions = 0;
-    }
-#else
-    Opts.Exceptions = 1;
-    if (Opts.CPlusPlus) {
-      Opts.CXXExceptions = 1;
-    }
-#endif // _MSC_VER
-    Opts.Deprecated = 1;
-    //Opts.Modules = 1;
-
-    // See test/CodeUnloading/PCH/VTables.cpp which implicitly compares clang
-    // to cling lang options. They should be the same, we should not have to
-    // give extra lang options to their invocations on any platform.
-    // Except -fexceptions -fcxx-exceptions.
-
-    Opts.Deprecated = 1;
-    Opts.GNUKeywords = 0;
-    Opts.Trigraphs = 1; // o no??! but clang has it on by default...
-
-#ifdef __APPLE__
-    Opts.Blocks = 1;
-    Opts.MathErrno = 0;
-#endif
-
-    // C++11 is turned on if cling is built with C++11: it's an interpreter;
-    // cross-language compilation doesn't make sense.
-    // Extracted from Boost/config/compiler.
-    // SunProCC has no C++11.
-    // VisualC's support is not obvious to extract from Boost...
-
-    // The value of __cplusplus in GCC < 5.0 (e.g. 4.9.3) when
-    // either -std=c++1y or -std=c++14 is specified is 201300L, which fails
-    // the test for C++14 or more (201402L) as previously specified.
-    // I would claim that the check should be relaxed to:
-
-#if __cplusplus > 201103L
-    if (Opts.CPlusPlus) Opts.CPlusPlus14 = 1;
-#endif
-#if __cplusplus >= 201103L
-    if (Opts.CPlusPlus) Opts.CPlusPlus11 = 1;
-#endif
-
-#ifdef _REENTRANT
-    Opts.POSIXThreads = 1;
-#endif
-  }
-
-  static void SetClingTargetLangOpts(LangOptions& Opts,
-                                     const TargetInfo& Target) {
-    if (Target.getTriple().getOS() == llvm::Triple::Win32) {
-      Opts.MicrosoftExt = 1;
-#ifdef _MSC_VER
-      Opts.MSCompatibilityVersion = (_MSC_VER * 100000);
-#endif
-      // Should fix http://llvm.org/bugs/show_bug.cgi?id=10528
-      Opts.DelayedTemplateParsing = 1;
-    } else {
-      Opts.MicrosoftExt = 0;
-    }
-  }
-
-  // This must be a copy of clang::getClangToolFullVersion(). Luckily
-  // we'll notice quickly if it ever changes! :-)
-  static std::string CopyOfClanggetClangToolFullVersion(StringRef ToolName) {
-    std::string buf;
-    llvm::raw_string_ostream OS(buf);
-#ifdef CLANG_VENDOR
-    OS << CLANG_VENDOR;
-#endif
-    OS << ToolName << " version " CLANG_VERSION_STRING " "
-       << getClangFullRepositoryVersion();
-
-    // If vendor supplied, include the base LLVM version as well.
-#ifdef CLANG_VENDOR
-    OS << " (based on LLVM " << PACKAGE_VERSION << ")";
-#endif
-
-    return OS.str();
-  }
-
   //
   //  Dummy function so we can use dladdr to find the executable path.
   //
@@ -434,24 +469,12 @@ namespace {
   {
   }
 
-  ///\brief Check the compile-time clang version vs the run-time clang version,
-  /// a mismatch could cause havoc. Reports if clang versions differ.
-  static void CheckClangCompatibility() {
-    if (clang::getClangToolFullVersion("cling")
-        != CopyOfClanggetClangToolFullVersion("cling"))
-      llvm::errs()
-        << "Warning in cling::CIFactory::createCI():\n  "
-        "Using incompatible clang library! "
-        "Please use the one provided by cling!\n";
-    return;
-  }
-
-  static bool strEqual(const char* a, const char* b, size_t n) {
-    return !strncmp(a, b, n) && !a[n];
-  }
-
   struct CompilerOpts {
-    bool hasMinusX, noBuiltinInc, noCXXIncludes, haveRsrcPath;
+    bool hasMinusX, noBuiltinInc, noCXXIncludes, haveRsrcPath, haveSysRoot;
+
+    static bool strEqual(const char* a, const char* b, size_t n) {
+      return !strncmp(a, b, n) && !a[n];
+    }
 
     void parse(const char* arg) {
       if (!hasMinusX && !strncmp(arg, "-x", 2))
@@ -462,14 +485,18 @@ namespace {
         noCXXIncludes = true;
       else if (!haveRsrcPath && strEqual(arg, "-resource-dir", 13))
         haveRsrcPath = true;
+#ifdef __APPLE__
+      else if (!haveSysRoot && strEqual(arg, "-isysroot", 9))
+        haveSysRoot = true;
+#endif
     }
 
     CompilerOpts(const char* const* iarg, const char* const* earg) :
       hasMinusX(false), noBuiltinInc(false), noCXXIncludes(false),
-      haveRsrcPath(false) {
+      haveRsrcPath(false), haveSysRoot(false) {
 
-      while (iarg < earg &&
-             (!hasMinusX || !noBuiltinInc || !noCXXIncludes || !haveRsrcPath)) {
+      while (iarg < earg && (!hasMinusX || !noBuiltinInc || !noCXXIncludes ||
+                            !haveRsrcPath || !haveSysRoot)) {
         if (strEqual(*iarg, "-Xclang", 7)) {
           // goto next arg if there is one
           if (++iarg < earg)
@@ -598,7 +625,7 @@ namespace {
         buffer.assign(clang);
         llvm::sys::path::append(buffer, "clang");
         clang.assign(&buffer[0], buffer.size());
-      
+
         if (llvm::sys::fs::is_regular_file(clang))
           ReadCompilerIncludePaths(clang.c_str(), buffer, sArguments);
   #endif
@@ -619,7 +646,16 @@ namespace {
           sArguments.addArgument("-nostdinc++");
       }
 
-  #if defined(__GLIBCXX__) && defined(__APPLE__)
+  #if defined(__APPLE__)
+
+      if (!opts.noBuiltinInc && !opts.haveSysRoot) {
+        std::string sysRoot;
+        if (getISysRoot(sysRoot))
+          sArguments.addArgument("-isysroot", std::move(sysRoot));
+      }
+
+    #if defined(__GLIBCXX__)
+
       // Avoid '__float128 is not supported on this target' errors
       bool haveCXXVers = false, haveCXXLib = false;
       for (const auto arg : args) {
@@ -632,7 +668,9 @@ namespace {
         sArguments.addArgument("-std=c++11");
       if (!haveCXXLib)
         sArguments.addArgument("-stdlib=libstdc++");
-  #endif
+
+    #endif //__GLIBCXX__
+  #endif // __APPLE__
 
 #endif // _MSC_VER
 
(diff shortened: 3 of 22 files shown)


test/lit.cfg
```diff
@@ -240,6 +240,13 @@ if not lit_config.quiet:
 
 #Start cling with nologo
 config.substitutions.append( ('%cling', config.cling + ' --nologo') )
+if True:
+  # Add a substitution for the builds generated include dir before install
```
karies: The comment seems to contradict the commit log - is this against picking up the include of whatever built cling or against the make-installed includes?

vgvassilev:
This can be problematic on OS X. Clang build from sources cannot find the system includes.

marsupial:
Yes, if cling is built with 'system-clang', you don't want to pick up any of the headers 'system-clang' may have installed...They are guaranteed to be binary incompatible with the ones that are used building cling (both because cling patches the headers, and because the version of clang building cling may be different).  As some tests include clang headers this can cause problems.

Additionally if only clang and std-c++ headers are installed (i.e. no dev-headers), you need to add the paths that so  that any includes generated during clangs build process are accessible.

The ultimate goal (which I've achieved over subsequent commits) is to be able to build cling & clang together so they can be installed and run from one installation point.

vgvassilev:
What I meant is:

```
echo "#include <iostream>" | clang -fsyntax-only -x c++ - -v
Apple LLVM version 7.3.0 (clang-703.0.31)
Target: x86_64-apple-darwin15.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
 "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang" -cc1 -triple x86_64-apple-macosx10.11.0 -Wdeprecated-objc-isa-usage -Werror=deprecated-objc-isa-usage -fsyntax-only -disable-free -disable-llvm-verifier -main-file-name - -mrelocation-model pic -pic-level 2 -mthread-model posix -mdisable-fp-elim -masm-verbose -munwind-tables -target-cpu core2 -target-linker-version 264.3.102 -v -dwarf-column-info -resource-dir /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../lib/clang/7.3.0 -stdlib=libc++ -fdeprecated-macro -fdebug-compilation-dir /Users/vvassilev/workspace/llvm-git/modules-selfhost/mod_issue1 -ferror-limit 19 -fmessage-length 173 -stack-protector 1 -fblocks -fobjc-runtime=macosx-10.11.0 -fencode-extended-block-signature -fcxx-exceptions -fexceptions -fmax-type-align=16 -fdiagnostics-show-option -fcolor-diagnostics -x c++ -
clang -cc1 version 7.3.0 (clang-703.0.31) default target x86_64-apple-darwin15.5.0
ignoring nonexistent directory "/usr/include/c++/v1"
#include "..." search starts here:
#include <...> search starts here:
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1
 /usr/local/include
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../lib/clang/7.3.0/include
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include
 /usr/include
 /System/Library/Frameworks (framework directory)
 /Library/Frameworks (framework directory)
End of search list.

11:14:44-vvassilev~/workspace/llvm-git/modules-selfhost/mod_issue1$ echo "#include <iostream>" | ~/workspace/llvm-git/obj/bin/clang -fsyntax-only -x c++ - -v
clang version 4.0.0 (https://github.com/llvm-mirror/clang.git 591295657ccd0b4fe27532a0bd9421caf76f0fc6) (https://github.com/llvm-mirror/llvm.git 570a51e9e7357c62486027d58fda22cd0b208ca6)
Target: x86_64-apple-darwin15.5.0
Thread model: posix
InstalledDir: /Users/vvassilev/workspace/llvm-git/obj/bin
 "/Users/vvassilev/workspace/llvm-git/obj/bin/clang-3.8" -cc1 -triple x86_64-apple-macosx10.11.0 -Wdeprecated-objc-isa-usage -Werror=deprecated-objc-isa-usage -fsyntax-only -disable-free -main-file-name - -mrelocation-model pic -pic-level 2 -mthread-model posix -mdisable-fp-elim -masm-verbose -munwind-tables -target-cpu core2 -target-linker-version 264.3.102 -v -dwarf-column-info -debugger-tuning=lldb -resource-dir /Users/vvassilev/workspace/llvm-git/obj/bin/../lib/clang/4.0.0 -stdlib=libc++ -fdeprecated-macro -fdebug-compilation-dir /Users/vvassilev/workspace/llvm-git/modules-selfhost/mod_issue1 -ferror-limit 19 -fmessage-length 173 -stack-protector 1 -fblocks -fobjc-runtime=macosx-10.11.0 -fencode-extended-block-signature -fcxx-exceptions -fexceptions -fmax-type-align=16 -fdiagnostics-show-option -fcolor-diagnostics -x c++ -
clang -cc1 version 4.0.0 based upon LLVM 4.0.0svn default target x86_64-apple-darwin15.5.0
ignoring nonexistent directory "/Users/vvassilev/workspace/llvm-git/obj/bin/../include/c++/v1"
ignoring nonexistent directory "/usr/include/c++/v1"
#include "..." search starts here:
#include <...> search starts here:
 /usr/local/include
 /Users/vvassilev/workspace/llvm-git/obj/bin/../lib/clang/4.0.0/include
 /usr/include
 /System/Library/Frameworks (framework directory)
 /Library/Frameworks (framework directory)
End of search list.
<stdin>:1:10: fatal error: 'iostream' file not found
#include <iostream>
         ^
1 error generated.
```
Freshly built clang on OSX cannot find the system headers. Perhaps I am missing something...

Also, I'd recommend to submit the other PRs. It may be quicker to review others because we have a few sets of commits which we already reviewed and I'd like to commit them as soon as possible, minimizing the rebasing efforts.

marsupial:
It's just OS X doesn't ship with system headers, right?
I have a commit that does something to fix this as well.

All of the commits still exists in at [https://github.com/root-mirror/cling/pull/17](url)

vgvassilev:
I am still lost. Is this commit part of this PR?

Are the other commits rebased?

marsupial:
> Freshly built clang on OSX cannot find the system headers
True, but if a user builds libc++ with clang then this is needed to find those headers.

> I am still lost. Is this commit that 'does something to fix this as well' part of this PR?
I have a commit in the original PR that does a better job detecting both system and c++ headers at runtime.

I think the larger issue is AddHostCXXIncludes is incredibly fragile.
It can be broken by something as simple as a user installing a new version of gcc (even while keeping the old one). Updating Xcode will break it as well. The only 'stable' solution I've found is building cling with clang and libc++.

I get you want to support system libc++ and libstdc++, but given the issues above does it make sense to discard the only stable variant?

vgvassilev:
I agree that AddHostCXXIncludes is incredibly fragile. It seems this approach will hide the problem. Why don't we use clang's driver to discover the include paths and find the most suitable version of the standard library?

marsupial:
What approach will hide what problem?
This addresses a very specific case if the user builds cling, clang, and libc++ together.
Any other use case and the behavior is the same.

vgvassilev:
The underlying issue is that cling sometimes cannot find the most compatible version of the platform header files on which it runs. To minimize interoperability issues they should be the same to the ones cling was compiled with.

Your solution to the problem, is introducing a workaround to help cling find the most suitable headers if it was built next to clang and libc++. I think this should be a part of more generic solution addressing the issues of AddHostCXXIncludes.

This would hide problems such as: https://travis-ci.org/vgvassilev/cling/jobs/150635140 around line 6080. This would make our testing infrastructure pass the tests but if cling is deployed on another platform it will fail to work.

I'd very much prefer AddHostCXXIncludes to find the best available on system library and include paths. If the user prefers to override them he should by prefixing their path in an env variable.

marsupial:
Cling will fail miserably unless it's deployed to a machine that has the exact same compiler.
Ultimately this is what I'm trying to avoid by deploying cling,clang, and libc++ together.
I can't refactor or fix every bug in the codebase in one commit...this is a step forward in my view.

The best I can offer is I know it to work across multiple systems.
[https://travis-ci.org/marsupial/cling/builds/150520519](url)
 gcc-4.7 to gcc-6 with libstdc++
 clang-3.5 to clang-3.8 with libstdc++
 clang-3.5 to clang-3.8 with libc++ built in tree
 os x xcode 6.4 to 8 with system libc++
 os x xcode 6.4 to 8 with libc++ built in tree

vgvassilev:
I'd like to discuss this more on IM. Could you drop us an email at cling-dev@cern.ch?

marsupial:
Sent, though left it too abstract and might easily get caught in spam...let me know.
While we wait
[https://github.com/root-mirror/cling/commit/b46117358a97b7f20d799d281fb0f87c4a273cd4](url), broke testing.

[https://github.com/root-mirror/cling/pull/17/commits/297694b13c02326127569fd1e2015d6cee51b661](url), fixes it

appveyor.yml
```diff
@@ -1,12 +1,21 @@
...
-  - x86
   - x64
+  - x86
+
 configuration:
   #- Debug
   - Release
 
+environment:
+  CMAKE: cmake.exe
```
vgvassilev: We depend on a patch in cmake. We need recent cmake (later than 01 Aug 2016 nightly build). Cpt installs it.

lib/Interpreter/CIFactory.cpp
```diff
@@ -492,7 +492,7 @@ namespace {
       }
 
       static const char *CppInclQuery =
-        "echo | LC_ALL=C " LLVM_CXX " -xc++ -E -v - 2>&1 >/dev/null "
+        "echo | LC_ALL=C " LLVM_CXX " -xc++ -E -v /dev/null 2>&1 "
```
vgvassilev: Why do you think this is broken? The `-` after `-v` means read output from the pipe, i.e. compile an empty cpp file.

lib/Interpreter/DynamicLibraryManager.cpp
```diff
@@ -26,7 +26,11 @@
 #include <Windows.h>
 #include <shlobj.h>
 #else
-#include <limits.h> /* PATH_MAX */
+#ifdef __APPLE__
+ #include <sys/syslimits.h> // PATH_MAX
+#else
+ #include <limits.h> // PATH_MAX
+#endif
```
vgvassilev: Could you instead use [llvm::sys::fs::make_absolute](http://llvm.org/docs/doxygen/html/namespacellvm_1_1sys_1_1fs.html#ab6c14b25e5233330251efa8e8ca0c212) to fully get rid of the PATH_MAX macro?

marsupial: You'd still need PATH_MAX to allocate a buffer to call llvm::sys::fs::make_absolute with.

vgvassilev: Yeah, you are right. I will commit as is.

lib/Interpreter/CIFactory.cpp
marsupial: It's hanging sometimes on OS X. But I see I missed the stdout redirect.
LC_ALL=C " LLVM_CXX " -xc++ -E -v /dev/null 2>&1 >/dev/null  ?

lib/Interpreter/CIFactory.cpp
```diff
@@ -338,9 +337,442 @@ static bool getVisualStudioDir(std::string &path) {
...
+#define GET_ISYSROOT_VER(maj, min) \
+  if (getISysRootVersion(SDKs, maj, min, sysRoot)) \
+    return true;
+
+  // Try to get the SDK for whatever cling was compiled with
+  #if defined(MAC_OS_X_VERSION_10_11)
+    GET_ISYSROOT_VER(10, 11);
+  #elif defined(MAC_OS_X_VERSION_10_10)
+    GET_ISYSROOT_VER(10, 10);
+  #elif defined(MAC_OS_X_VERSION_10_9)
```
vgvassilev: We support only 10.10 and 10.11. The 10.9 support is arguable not to speak earlier versions. Could you keep only these three?

lib/Interpreter/CIFactory.cpp
```diff
@@ -338,9 +337,442 @@ static bool getVisualStudioDir(std::string &path) {
...
+
+#endif
+  
+  ///\brief Adds standard library -I used by whatever compiler is found in PATH.
+  static void AddHostArguments(std::vector<const char*>& args,
+                               const char* llvmdir, const CompilerOpts& opts) {
+    static AdditionalArguments sArguments;
+    if (sArguments.empty()) {
+#ifdef _MSC_VER
+      // Honor %INCLUDE%. It should know essential search paths with vcvarsall.bat.
```
vgvassilev: @bellenot could you review?

bellenot: Looks OK with me, the _MSC_VER part looks very similar to what is already there (unless I overlooked something)...

lib/Interpreter/CIFactory.cpp
marsupial: This is for the SDK built with, not running. I can easily envision someone -needing - to compile for SDK 10.8 even though they are on 10.10.

I think it better to officially support 10.10, 10.11, 10.9 but not necessarily make it harder for someone who's not on those versions.

vgvassilev: I'd suggest to warn for the versions different than 10.11, 10.10 and 10.9.

marsupial: I could add a #warning, but:
 Its a non-portable extension.
 This is something that should be warned at configuration time to give a heads up before building.

vgvassilev: I don't think we can build cling on something earlier than 10.10. As you said, we could run cling on platforms older than the one we built on. I think we should have a runtime check.

marsupial: I build with SDK 10.9 with no problems.
I get wanting to warn the user using an SDK you feel is too outdated, but I don't get what runtime check you want to perform. What are you trying to accomplish?

I think handling the warning at config time is better (so the user can use a more recent version if they have one). The build can easily take an hour to succeed on a local machine, and giving a heads up before they've entered into that process is nicer.

vgvassilev: I agree we should have a configure time check. It covers only the cases when building cling on the same machine on which it will run.

What about if I take a prebuilt binary and run it on 10.0? How cling would warn me? There is no configure time there, right? Maybe this code should be moved in cling's RuntimeUnverse.h (which gets #included every time cling starts up).

marsupial: dyld wouldn't ever let cling run in that case. For the most part the SDK headers are stable between  releases. When it's not OS X knows how to handle it.
Otherwise everything on a users computer would have to be updated when a user did an OS update.

This is predominately to support C-level includes, not C++. Though it impacts C++ as #including \<string\> will most likely look for \<wchar.h\>

include/cling/Utils/Paths.h
```diff
@@ -0,0 +1,96 @@
+//--------------------------------------------------------------------*- C++ -*-
+// CLING - the C++ LLVM-based InterpreterG :)
+// author:  Jermaine Dupris <support@greyangle.com>
```
vgvassilev: Is that your real name? If you are uncomfortable in putting it, please omit it.

marsupial: No, just think everyone deserves more credit than they get...
If it really bothers you I'll pull it, but it makes me happy, and I'll be consistent.

vgvassilev: I am confused. Who wrote this test? You our a colleague with that name?
I am fine with keeping that name there, as long as you can guarantee the person is aware that this piece of code is donated to the cling project.

lib/Interpreter/CIFactory.cpp
```diff
@@ -446,6 +446,43 @@ namespace {
...
   }
 
+  static bool strEqual(const char* a, const char* b, size_t n) {
+    return !strncmp(a, b, n) && !a[n];
+  }
+
+  struct CompilerOpts {
+    bool hasMinusX, noBuiltinInc, noCXXIncludes, haveRsrcPath;
+
+    void parse(const char* arg) {
```
vgvassilev: How about taking `llvm::StringRef` here and using `llvm::StringRef::equals`?

vgvassilev: Sorry, that seems to be gone later.

marsupial: Yes, it gets much prettier later.

lib/Interpreter/CIFactory.cpp
vgvassilev: I would like to see a startup warning when cling runs on a platform it was not built for. It can be implemented wherever it is most convenient.

I am trying to warn the user when he is running cling on a different platform that it was built for. Eg. cling was built on 10.10 and the user is *running* the binary on 10.5. I am worried about the case where we built cling with libc++, which is not abi compatible with the one that is on the platform when cling binary runs. This is what we warn about on when running with libstdc++.

include/cling/Interpreter/InvocationOptions.h
```diff
@@ -14,16 +14,30 @@
 #include <vector>
 
 namespace cling {
+
+  class CompilerOpts {
+  public:
+    /// \brief Construct CompilerOpts from given arguments, saving them into
+    /// Remaining.
```
vgvassilev: It seems like a leftover comment.

include/cling/Interpreter/InvocationOptions.h
```diff
@@ -32,12 +46,16 @@ namespace cling {
     std::vector<std::string> LibsToLoad;
     std::vector<std::string> LibSearchPath;
     std::vector<std::string> Inputs;
+    cling::CompilerOpts CompilerOpts;
```
vgvassilev: We use the notion `cling::CompilerOptions CompilerOpts`, i.e. we shorten the variable name and its accessors (if any) and we keep the class name verbose.

include/cling/Interpreter/InvocationOptions.h
```diff
@@ -14,16 +14,30 @@
 #include <vector>
 
 namespace cling {
+
+  class CompilerOpts {
```
vgvassilev: Could you add a documentation here? How this is different from InvocationOptions, why did you decide to not extend it?

lib/Interpreter/CIFactory.cpp
marsupial: Yes the OS already handles that by refusing to let the application launch, or runtime link errors (undefined symbols). When libc++ or any system provided lib is made abi incompatible they make changes so it cannot be used by clients expecting something else.

lib/Interpreter/CIFactory.cpp
```diff
@@ -42,9 +43,19 @@
 
 #include <memory>
 
+#ifndef _MSC_VER
+#include "cling-compiledata.h"
+# include <unistd.h>
+# define getcwd_func getcwd
```
vgvassilev: Could you remove the indents here?

include/cling/Interpreter/InvocationOptions.h
marsupial: Remaining is an instance variable.

lib/Interpreter/CIFactory.cpp
vgvassilev: Ah, ok, that was the missing bit. Thanks!

include/cling/Interpreter/InvocationOptions.h
vgvassilev: Could you prefix with \p. Eg. `\p Remaining`.

include/cling/Interpreter/InvocationOptions.h
marsupial: InvocationOptions parses a lot of stuff one variant of CIFactory::createCI has no use for.

lib/Interpreter/CIFactory.cpp
marsupial: The indents were there already, you want me to match that or remove them?

vgvassilev: Remove them, please. (We really need to add our own clang-format settings.).

marsupial: Why not just use LLVM's . Wouldn't it be nice if this could be integrated into their codebase one day?

include/cling/Interpreter/InvocationOptions.h
marsupial: Is the \p because it's hanging or font styling?

include/cling/Interpreter/InvocationOptions.h
marsupial: And while I'm updating this commit, any issue with making
class CIFactory -> namespace CIFactory

vgvassilev: Not in particular, let see how it would look like.

include/cling/Interpreter/InvocationOptions.h
vgvassilev: Yes, it shows up in different font style. More info [here](https://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdp).

include/cling/Utils/Paths.h
```diff
@@ -0,0 +1,96 @@
...
+    ///
+    ///\param[in] PathsStr - Path(s)
+    ///\param[in] Opts - HeaderSearchOptions to add paths into
+    ///\param[in] Delim - Delimiter to separate paths or NULL if a single path
+    ///
+    void AddIncludePaths(llvm::StringRef PathStr,
+                         clang::HeaderSearchOptions& Opts,
+                         const char* Delim = ":");
+
+    ///\brief Prints the current include paths int the HeaderSearchOptions.
```
vgvassilev: Typo.

include/cling/Utils/Paths.h
```diff
@@ -0,0 +1,96 @@
...
+    ///       with a "-I" or similar, and some entries of incpaths will signal
+    ///       a new include path region (e.g. "-cxx-isystem"). Also, flags
+    ///       defining header search behavior will be included in incpaths, e.g.
+    ///       "-nostdinc".
+    ///
+    void CopyIncludePaths(const clang::HeaderSearchOptions& Opts,
+                          llvm::SmallVectorImpl<std::string>& Paths,
+                          bool WithSystem, bool WithFlags);
+
+    ///\brief Prints the current include paths int the HeaderSearchOptions.
```
vgvassilev: Typo.

lib/Interpreter/CIFactory.cpp
vgvassilev: It would, but we are not there yet.

lib/Interpreter/CIFactory.cpp
```diff
@@ -42,9 +43,19 @@
...
+#endif
+
+// FIXME: This code has been taken (copied from) llvm/tools/clang/lib/Driver/WindowsToolChain.cpp
```
vgvassilev: Indeed, in order to do that we could rely on `getToolChain()`. This way we could keep the implementation in one place. Last time @bellenot tried it seemed quite a bit of coding, plus eventually patching clang and we decided to postpone it.

lib/Interpreter/CIFactory.cpp
```diff
@@ -338,9 +339,367 @@ static bool getVisualStudioDir(std::string &path) {
...
+      return false;
+  }
+
+  SDKs.append("/Platforms/MacOSX.platform/Developer/SDKs/");
+  if (!fs::is_directory(SDKs))
+    return false;
+
+
+  // Seems to make more sense to get the currently running SDK so any loaded
+  // libraries won't casue conflicts
```
vgvassilev: Typo.

lib/Interpreter/CIFactory.cpp
```diff
@@ -338,9 +339,367 @@ static bool getVisualStudioDir(std::string &path) {
...
+    GET_ISYSROOT_VER(10, 1);
+  #else // MAC_OS_X_VERSION_10_0
+    GET_ISYSROOT_VER(10, 0);
+  #endif
+
+#undef GET_ISYSROOT_VER
+
+  // Scan the SDKs directory and use the latest
+  std::error_code ec;
+  std::vector<std::string> srtd;
```
vgvassilev: Can't we not allocate strings here but use llvm::StringRef/string_view. Just adding a //FIXME here is fine with me.

lib/Interpreter/CIFactory.cpp
```diff
@@ -338,9 +339,367 @@ static bool getVisualStudioDir(std::string &path) {
...
+  // without being given the address of a function in the main executable).
+  std::string GetExecutablePath(const char *Argv0) {
+    // This just needs to be some symbol in the binary; C++ doesn't
+    // allow taking the address of ::main however.
+    void *MainAddr = (void*) intptr_t(GetExecutablePath);
+    return llvm::sys::fs::getMainExecutable(Argv0, MainAddr);
+  }
+
+  class AdditionalArguments {
+    typedef std::vector< std::pair<const char*,std::string> > container_t;
```
vgvassilev: Could we use a more meaningful name here? Eg. AdditionalArgsList?

lib/Interpreter/CIFactory.cpp
```diff
@@ -338,9 +339,367 @@ static bool getVisualStudioDir(std::string &path) {
...
+      m_Saved.push_back(std::make_pair(arg,std::move(value)));
+    }
+    container_t::const_iterator begin() const { return m_Saved.begin(); }
+    container_t::const_iterator end() const { return m_Saved.end(); }
+    bool empty() const { return m_Saved.empty(); }
+  };
+
+#ifndef _MSC_VER
+
+  static void CompilerIncludePaths(const char* Compiler,
```
vgvassilev: Could you add a verb in the method name?

lib/Interpreter/CIFactory.cpp
```diff
@@ -338,9 +339,367 @@ static bool getVisualStudioDir(std::string &path) {
...
+                        "/{if (!/^#include </ && !/^End of search/){ print }}' "
+                        "| GREP_OPTIONS= grep -E \"(c|g)\\+\\+\"");
+
+    if (FILE *pf = ::popen(CppInclQuery.c_str(), "r")) {
+      Buf.resize(Buf.capacity_in_bytes());
+      while (fgets(&Buf[0], Buf.capacity_in_bytes(), pf) && Buf[0]) {
+        size_t lenbuf = strlen(&Buf[0]);
+        Buf[lenbuf - 1] = 0; // remove trailing \n
+        // Skip leading whitespace:
+        const char *start = &Buf[0];
```
vgvassilev: SmallString::str().trim() should do the trick.

lib/Interpreter/InvocationOptions.cpp
```diff
@@ -73,79 +77,104 @@ namespace {
...
-                        0, ExcludeOptionFlagMasks));
-    for (ArgList::const_iterator it = Args.begin(),
-           ie = Args.end(); it != ie; ++it) {
-      if ( (*it)->getOption().getKind() == Option::InputClass ) {
-          Opts.Inputs.push_back(std::string( (*it)->getValue() ) );
-      }
+CompilerOpts::CompilerOpts(int argc, const char* const* argv) :
+  Language(false), ResourceDir(false), SysRoot(false), NoBuiltinInc(false),
+  NoCXXInc(false), StdVersion(false), StdLib(false)
+{
```
vgvassilev: `{` should go on the previous line.

lib/Utils/Paths.cpp
```diff
@@ -0,0 +1,229 @@
...
+// This file is dual-licensed: you can choose to license it under the University
+// of Illinois Open Source License or the GNU Lesser General Public License. See
+// LICENSE.TXT for details.
+//------------------------------------------------------------------------------
+
+#include "cling/Utils/Paths.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/raw_ostream.h"
+#include "llvm/Support/ErrorHandling.h"
+#include "clang/Lex/HeaderSearchOptions.h"
```
vgvassilev: clang header files go after cling's. Please order alphabetically.

lib/Utils/Paths.cpp
```diff
@@ -0,0 +1,229 @@
+//--------------------------------------------------------------------*- C++ -*-
+// CLING - the C++ LLVM-based InterpreterG :)
+// author:  Jermaine Dupris <support@greyangle.com>
```
vgvassilev: Same question does this person know his code is donated? I'd like to have an internal discussion on this matter. For now, could you remove the authors. This way, the code will align better under our current "contribution" section. We can re-add them later when we resolve this issue. Otherwise, the PR will have to wait for the outcome of the discussion, which can be very long.

lib/Interpreter/CIFactory.cpp
marsupial: That'll copy the whole string just to trim the front.
::isspace ?

vgvassilev: According to the [documentation](http://llvm.org/docs/doxygen/html/SmallString_8h_source.html#l00253) it will not copy it.

lib/Interpreter/CIFactory.cpp
marsupial: No, I think the directory_iterator mutates, you need to copy it's current path.

include/cling/Utils/Paths.h
marsupial: Don't understand.

lib/Interpreter/CIFactory.cpp
marsupial: Don't see it.

lib/Utils/Paths.cpp
marsupial: Yes, no legal issues.

lib/Interpreter/CIFactory.cpp
vgvassilev: 'casue'->cause

include/cling/Utils/Paths.h
vgvassilev: The documentation of `CopyIncludePaths` is the same as `DumpIncludePaths`. I don't think `CopyIncludePaths` should print, right?

lib/Utils/Paths.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+    void addFlag(const char* Arg) override { m_Vector.push_back(Arg); }
+    void addPath(const std::string& Str) override {  m_Vector.push_back(Str); }
+  };
+}
+
+namespace cling {
+namespace utils {
+
+void CopyIncludePaths(const clang::HeaderSearchOptions& Opts,
```
vgvassilev: I'd move `CopyIncludePaths` and `DumpIncludePaths` into the base class; rename it to something like `IncludePathHelper`; and remove the two extra classes. Unless I am missing something, I don't see a point in introducing such a complex implementation for something that trivial.

lib/Interpreter/CIFactory.cpp
marsupial: SmallString::SmallString copies. 
SmallString::str() returns a StringRef to the copy.
http://llvm.org/docs/doxygen/html/SmallString_8h_source.html#00035

lib/Utils/Paths.cpp
marsupial: I don't get what you're suggesting.

The original implementation copied them all into a vector just to dump them out. Now it streams them out avoiding the copy.
 
CopyIncludePaths is to keep previous behavior of list:
[0] = "-I"
[1] = "Path"
[2] = "-I"
[3] = "Path"

DumpIncludePaths dumps in a more readable way that can be copy/pasted:
-I Path
-I Path

lib/Utils/Paths.cpp
vgvassilev: I'd prefer, for now, to remove the authorship. I will update the CREDITS file and I will mention him in the next release notes.

include/cling/Utils/Paths.h
```diff
@@ -0,0 +1,96 @@
...
+
+namespace clang {
+  class HeaderSearchOptions;
+}
+
+namespace cling {
+  namespace utils {
+
+    enum SplitMode {
+      kPruneNonExistant,  // Don't add non-existant paths into output
```
vgvassilev: I'd like these enum constant comments to show in doxygen too. IIRC, /// or ///< was doing the trick.

lib/Interpreter/CIFactory.cpp
vgvassilev: Ok, could you in this case add clarifying comment about that.

lib/Interpreter/CIFactory.cpp
```diff
@@ -338,9 +339,367 @@ static bool getVisualStudioDir(std::string &path) {
...
+  // without being given the address of a function in the main executable).
+  std::string GetExecutablePath(const char *Argv0) {
+    // This just needs to be some symbol in the binary; C++ doesn't
+    // allow taking the address of ::main however.
+    void *MainAddr = (void*) intptr_t(GetExecutablePath);
+    return llvm::sys::fs::getMainExecutable(Argv0, MainAddr);
+  }
+
+  class AdditionalArgList {
+    typedef std::vector< std::pair<const char*,std::string> > container_t;
```
vgvassilev: I left a comment here but it seems gone now... Could you rename this to something like `AdditionalArgList`.

lib/Utils/Paths.cpp
vgvassilev: I don't see why we need class hierarchy to implement this.

lib/Interpreter/InvocationOptions.cpp
```diff
@@ -19,13 +19,17 @@
 #include "llvm/Support/raw_ostream.h"
 
 #include <memory>
+#include <algorithm>
```
vgvassilev: Please order alphabetically.

lib/Utils/Paths.cpp
marsupial: Well the behavior is different, one calls push_back the other << 
Whats the issue with a hierarchy and more general?

vgvassilev: I think this can be implemented by adding the 2 functions in the base class and remove the derived classes. Makes it more maintainable and consistent to the rest of the code base. I don't have issues with java-style decorators/hierarchies but we are trying to avoid them as much as possible.

lib/Interpreter/CIFactory.cpp
vgvassilev: On 16/08/16 13:01, marsupial wrote:
>
> In lib/Interpreter/CIFactory.cpp 
> <https://github.com/root-mirror/cling/pull/26#discussion_r74914489>:
>
> > +                                   llvm::SmallVectorImpl<char>& Buf,
> > +                                   AdditionalArguments& Args) {
> > +    std::string CppInclQuery("LC_ALL=C ");
> > +    CppInclQuery.append(Compiler);
> > +    CppInclQuery.append(" -xc++ -E -v /dev/null 2>&1 >/dev/null "
> > +                        "| awk '/^#include </,/^End of search"
> > +                        "/{if (!/^#include </ && !/^End of search/){ print }}' "
> > +                        "| GREP_OPTIONS= grep -E \"(c|g)\\+\\+\"");
> > +
> > +    if (FILE *pf = ::popen(CppInclQuery.c_str(), "r")) {
> > +      Buf.resize(Buf.capacity_in_bytes());
> > +      while (fgets(&Buf[0], Buf.capacity_in_bytes(), pf) && Buf[0]) {
> > +        size_t lenbuf = strlen(&Buf[0]);
> > +        Buf[lenbuf - 1] = 0; // remove trailing \n
> > +        // Skip leading whitespace:
> > +        const char *start = &Buf[0];
>
> SmallString::SmallString copies.
> SmallString::str() returns a StringRef to the copy.
> http://llvm.org/docs/doxygen/html/SmallString_8h_source.html#00035
>

Not sure I understand. I am suggesting to replace Buf's type from 
SmallVectorImpl to SmallString.

lib/Utils/Paths.cpp
vgvassilev: On 16/08/16 11:18, marsupial wrote:
>
> In lib/Utils/Paths.cpp 
> <https://github.com/root-mirror/cling/pull/26#discussion_r74901399>:
>
> > @@ -0,0 +1,229 @@
> > +//--------------------------------------------------------------------*- C++ -*-
> > +// CLING - the C++ LLVM-based InterpreterG :)
> > +// author:  Jermaine Dupris <support@greyangle.com>
>
> Yes, no legal issues.
>
Please, remove authors. They will be mentioned in the CREDITS.

lib/Utils/Paths.cpp
marsupial: The 2 functions are in the base class, but virtual and overridden (so it's actually 4 functions)

The only thing I can think of that accomplishes what you're saying is:
union {
  llvm::raw_ostream* strm;
 std::vector<std::string> *vec;
} U;
bool IsVec;
if (IsVec) U->vec->push_back(); else (*U->strm) << ;

Which seems a way weirder construct (to me), much harder to maintain, and what parts of C++ were designed to avoid.

lib/Interpreter/CIFactory.cpp
marsupial: Ah, that wasn't clear to me. Though you wanted SmallString(&Buf[0], len).str()

marsupial: Buf is already a SmallString, which means adding it's size explicitly in the argument list:

  static void ReadCompilerIncludePaths(const char* Compiler,
                                       llvm::SmallString<2048>& Buf,
                                       AdditionalArgList& Args) {

marsupial: Or I can just construct a StringRef from Buf!

lib/Utils/Paths.cpp
marsupial: I did that already.

lib/Utils/Paths.cpp
vgvassilev: Put the `SmallVectorImpl<std::string>& m_Vector` in `FormatIncludes`. Rename `FormatIncludes::dump` into `FormatIncludes::collect`. Add a routine `FormatIncludes::dump(llvm::raw_ostream* Out = nullptr);` and a routine `FormatIncludes::copy(llvm::SmallVectorImpl<std::string>& Paths);`.

This greatly simplifies the implementation at the cost of filling an vector with a few strings upon calling `DumpIncludePaths`.

marsupial: The way it was you that you want to go back to makes it incredibly difficult to parse the output. The way it is now each line is a discrete unit of flags / arguments.

Is it possible to change the way the Copy works, so each element is also discreet. I could probably simplify it that way and we'd both be happy.

[0] -index-header-map
[1] -I
[2] PATH

would become:

[0] -index-header-map -I PATH
[1] -I PATH

vgvassilev: I think changing that would cause a breakage in ROOT (the main user of cling). What about implementing what I suggested and moving the more intrusive change in a separate PR. If we solve this one I can integrate this PR tomorrow.

vgvassilev:
I started integrating this PR and I get 
```
$ core/utils/src/rootcling_tmp -f core/base/src/G__Core.cxx -s lib/libCore.so -c     /Users/vvassilev/workspace/root/core/base/inc/TApplication.h /Users/vvassilev/workspace/root/core/base/inc/TApplicationImp.h /Users/vvassilev/workspace/root/core/base/inc/TAtt3D.h /Users/vvassilev/workspace/root/core/base/inc/TAttAxis.h /Users/vvassilev/workspace/root/core/base/inc/TAttBBox.h /Users/vvassilev/workspace/root/core/base/inc/TAttBBox2D.h /Users/vvassilev/workspace/root/core/base/inc/TAttFill.h /Users/vvassilev/workspace/root/core/base/inc/TAttLine.h /Users/vvassilev/workspace/root/core/base/inc/TAttMarker.h /Users/vvassilev/workspace/root/core/base/inc/TAttPad.h /Users/vvassilev/workspace/root/core/base/inc/TAttText.h /Users/vvassilev/workspace/root/core/base/inc/TBenchmark.h /Users/vvassilev/workspace/root/core/base/inc/TBrowser.h /Users/vvassilev/workspace/root/core/base/inc/TBrowserImp.h /Users/vvassilev/workspace/root/core/base/inc/TBuffer.h /Users/vvassilev/workspace/root/core/base/inc/TBuffer3D.h /Users/vvassilev/workspace/root/core/base/inc/TBuffer3DTypes.h /Users/vvassilev/workspace/root/core/base/inc/TCanvasImp.h /Users/vvassilev/workspace/root/core/base/inc/TColor.h /Users/vvassilev/workspace/root/core/base/inc/TColorGradient.h /Users/vvassilev/workspace/root/core/base/inc/TContextMenu.h /Users/vvassilev/workspace/root/core/base/inc/TContextMenuImp.h /Users/vvassilev/workspace/root/core/base/inc/TControlBarImp.h /Users/vvassilev/workspace/root/core/base/inc/TDatime.h /Users/vvassilev/workspace/root/core/base/inc/TDirectory.h /Users/vvassilev/workspace/root/core/base/inc/TEnv.h /Users/vvassilev/workspace/root/core/base/inc/TError.h /Users/vvassilev/workspace/root/core/base/inc/TException.h /Users/vvassilev/workspace/root/core/base/inc/TExec.h /Users/vvassilev/workspace/root/core/base/inc/TFolder.h /Users/vvassilev/workspace/root/core/base/inc/TGuiFactory.h /Users/vvassilev/workspace/root/core/base/inc/TInspectorImp.h /Users/vvassilev/workspace/root/core/base/inc/TMD5.h /Users/vvassilev/workspace/root/core/base/inc/TMacro.h /Users/vvassilev/workspace/root/core/base/inc/TMathBase.h /Users/vvassilev/workspace/root/core/base/inc/TMemberInspector.h /Users/vvassilev/workspace/root/core/base/inc/TMessageHandler.h /Users/vvassilev/workspace/root/core/base/inc/TNamed.h /Users/vvassilev/workspace/root/core/base/inc/TObjString.h /Users/vvassilev/workspace/root/core/base/inc/TObject.h /Users/vvassilev/workspace/root/core/base/inc/TPRegexp.h /Users/vvassilev/workspace/root/core/base/inc/TPluginManager.h /Users/vvassilev/workspace/root/core/base/inc/TPoint.h /Users/vvassilev/workspace/root/core/base/inc/TProcessID.h /Users/vvassilev/workspace/root/core/base/inc/TProcessUUID.h /Users/vvassilev/workspace/root/core/base/inc/TQClass.h /Users/vvassilev/workspace/root/core/base/inc/TQCommand.h /Users/vvassilev/workspace/root/core/base/inc/TQConnection.h /Users/vvassilev/workspace/root/core/base/inc/TQObject.h /Users/vvassilev/workspace/root/core/base/inc/TQObjectEmitVA.h /Users/vvassilev/workspace/root/core/base/inc/TROOT.h /Users/vvassilev/workspace/root/core/base/inc/TRef.h /Users/vvassilev/workspace/root/core/base/inc/TRefCnt.h /Users/vvassilev/workspace/root/core/base/inc/TRegexp.h /Users/vvassilev/workspace/root/core/base/inc/TRemoteObject.h /Users/vvassilev/workspace/root/core/base/inc/TRootIOCtor.h /Users/vvassilev/workspace/root/core/base/inc/TStopwatch.h /Users/vvassilev/workspace/root/core/base/inc/TStorage.h /Users/vvassilev/workspace/root/core/base/inc/TString.h /Users/vvassilev/workspace/root/core/base/inc/TStringLong.h /Users/vvassilev/workspace/root/core/base/inc/TStyle.h /Users/vvassilev/workspace/root/core/base/inc/TSysEvtHandler.h /Users/vvassilev/workspace/root/core/base/inc/TSystem.h /Users/vvassilev/workspace/root/core/base/inc/TSystemDirectory.h /Users/vvassilev/workspace/root/core/base/inc/TSystemFile.h /Users/vvassilev/workspace/root/core/base/inc/TTask.h /Users/vvassilev/workspace/root/core/base/inc/TThreadSlots.h /Users/vvassilev/workspace/root/core/base/inc/TTime.h /Users/vvassilev/workspace/root/core/base/inc/TTimer.h /Users/vvassilev/workspace/root/core/base/inc/TUUID.h /Users/vvassilev/workspace/root/core/base/inc/TVersionCheck.h /Users/vvassilev/workspace/root/core/base/inc/TVirtualFFT.h /Users/vvassilev/workspace/root/core/base/inc/TVirtualGL.h /Users/vvassilev/workspace/root/core/base/inc/TVirtualPS.h /Users/vvassilev/workspace/root/core/base/inc/TVirtualPad.h /Users/vvassilev/workspace/root/core/base/inc/TVirtualPadEditor.h /Users/vvassilev/workspace/root/core/base/inc/TVirtualPadPainter.h /Users/vvassilev/workspace/root/core/base/inc/TVirtualViewer3D.h /Users/vvassilev/workspace/root/core/base/inc/GuiTypes.h /Users/vvassilev/workspace/root/core/base/inc/KeySymbols.h /Users/vvassilev/workspace/root/core/base/inc/Buttons.h /Users/vvassilev/workspace/root/core/base/inc/TTimeStamp.h /Users/vvassilev/workspace/root/core/base/inc/TVirtualMutex.h /Users/vvassilev/workspace/root/core/base/inc/TVirtualPerfStats.h /Users/vvassilev/workspace/root/core/base/inc/TVirtualX.h /Users/vvassilev/workspace/root/core/base/inc/TParameter.h /Users/vvassilev/workspace/root/core/base/inc/TVirtualAuth.h /Users/vvassilev/workspace/root/core/base/inc/TFileInfo.h /Users/vvassilev/workspace/root/core/base/inc/TFileCollection.h /Users/vvassilev/workspace/root/core/base/inc/TRedirectOutputGuard.h /Users/vvassilev/workspace/root/core/base/inc/TVirtualMonitoring.h /Users/vvassilev/workspace/root/core/base/inc/TObjectSpy.h /Users/vvassilev/workspace/root/core/base/inc/TUri.h /Users/vvassilev/workspace/root/core/base/inc/TUrl.h /Users/vvassilev/workspace/root/core/base/inc/TInetAddress.h /Users/vvassilev/workspace/root/core/base/inc/TVirtualTableInterface.h /Users/vvassilev/workspace/root/core/base/inc/TBase64.h /Users/vvassilev/workspace/root/core/cont/inc/TArray.h /Users/vvassilev/workspace/root/core/cont/inc/TArrayC.h /Users/vvassilev/workspace/root/core/cont/inc/TArrayD.h /Users/vvassilev/workspace/root/core/cont/inc/TArrayF.h /Users/vvassilev/workspace/root/core/cont/inc/TArrayI.h /Users/vvassilev/workspace/root/core/cont/inc/TArrayL.h /Users/vvassilev/workspace/root/core/cont/inc/TArrayL64.h /Users/vvassilev/workspace/root/core/cont/inc/TArrayS.h /Users/vvassilev/workspace/root/core/cont/inc/TBits.h /Users/vvassilev/workspace/root/core/cont/inc/TBtree.h /Users/vvassilev/workspace/root/core/cont/inc/TClassTable.h /Users/vvassilev/workspace/root/core/cont/inc/TClonesArray.h /Users/vvassilev/workspace/root/core/cont/inc/TCollection.h /Users/vvassilev/workspace/root/core/cont/inc/TCollectionProxyInfo.h /Users/vvassilev/workspace/root/core/cont/inc/TExMap.h /Users/vvassilev/workspace/root/core/cont/inc/THashList.h /Users/vvassilev/workspace/root/core/cont/inc/THashTable.h /Users/vvassilev/workspace/root/core/cont/inc/TIterator.h /Users/vvassilev/workspace/root/core/cont/inc/TList.h /Users/vvassilev/workspace/root/core/cont/inc/TMap.h /Users/vvassilev/workspace/root/core/cont/inc/TObjArray.h /Users/vvassilev/workspace/root/core/cont/inc/TObjectTable.h /Users/vvassilev/workspace/root/core/cont/inc/TOrdCollection.h /Users/vvassilev/workspace/root/core/cont/inc/TRefArray.h /Users/vvassilev/workspace/root/core/cont/inc/TRefTable.h /Users/vvassilev/workspace/root/core/cont/inc/TSeqCollection.h /Users/vvassilev/workspace/root/core/cont/inc/TSortedList.h /Users/vvassilev/workspace/root/core/cont/inc/TVirtualCollectionProxy.h /Users/vvassilev/workspace/root/core/cont/inc/ROOT/TSeq.h /Users/vvassilev/workspace/root/core/meta/inc/TBaseClass.h /Users/vvassilev/workspace/root/core/meta/inc/TClass.h /Users/vvassilev/workspace/root/core/meta/inc/TClassGenerator.h /Users/vvassilev/workspace/root/core/meta/inc/TClassMenuItem.h /Users/vvassilev/workspace/root/core/meta/inc/TClassRef.h /Users/vvassilev/workspace/root/core/meta/inc/TClassStreamer.h /Users/vvassilev/workspace/root/core/meta/inc/TDataMember.h /Users/vvassilev/workspace/root/core/meta/inc/TDataType.h /Users/vvassilev/workspace/root/core/meta/inc/TDictAttributeMap.h /Users/vvassilev/workspace/root/core/meta/inc/TDictionary.h /Users/vvassilev/workspace/root/core/meta/inc/TEnum.h /Users/vvassilev/workspace/root/core/meta/inc/TEnumConstant.h /Users/vvassilev/workspace/root/core/meta/inc/TFileMergeInfo.h /Users/vvassilev/workspace/root/core/meta/inc/TFunction.h /Users/vvassilev/workspace/root/core/meta/inc/TFunctionTemplate.h /Users/vvassilev/workspace/root/core/meta/inc/TGenericClassInfo.h /Users/vvassilev/workspace/root/core/meta/inc/TGlobal.h /Users/vvassilev/workspace/root/core/meta/inc/TInterpreter.h /Users/vvassilev/workspace/root/core/meta/inc/TInterpreterValue.h /Users/vvassilev/workspace/root/core/meta/inc/TIsAProxy.h /Users/vvassilev/workspace/root/core/meta/inc/TListOfDataMembers.h /Users/vvassilev/workspace/root/core/meta/inc/TListOfEnums.h /Users/vvassilev/workspace/root/core/meta/inc/TListOfEnumsWithLock.h /Users/vvassilev/workspace/root/core/meta/inc/TListOfFunctionTemplates.h /Users/vvassilev/workspace/root/core/meta/inc/TListOfFunctions.h /Users/vvassilev/workspace/root/core/meta/inc/TMemberStreamer.h /Users/vvassilev/workspace/root/core/meta/inc/TMethod.h /Users/vvassilev/workspace/root/core/meta/inc/TMethodArg.h /Users/vvassilev/workspace/root/core/meta/inc/TMethodCall.h /Users/vvassilev/workspace/root/core/meta/inc/TProtoClass.h /Users/vvassilev/workspace/root/core/meta/inc/TRealData.h /Users/vvassilev/workspace/root/core/meta/inc/TSchemaHelper.h /Users/vvassilev/workspace/root/core/meta/inc/TSchemaRule.h /Users/vvassilev/workspace/root/core/meta/inc/TSchemaRuleSet.h /Users/vvassilev/workspace/root/core/meta/inc/TStreamer.h /Users/vvassilev/workspace/root/core/meta/inc/TStreamerElement.h /Users/vvassilev/workspace/root/core/meta/inc/TToggle.h /Users/vvassilev/workspace/root/core/meta/inc/TToggleGroup.h /Users/vvassilev/workspace/root/core/meta/inc/TVirtualIsAProxy.h /Users/vvassilev/workspace/root/core/meta/inc/TVirtualRefProxy.h /Users/vvassilev/workspace/root/core/meta/inc/TVirtualStreamerInfo.h -DSYSTEM_TYPE_macosx /Users/vvassilev/workspace/root/core/unix/inc/TUnixSystem.h /Users/vvassilev/workspace/root/core/macosx/inc/TMacOSXSystem.h /Users/vvassilev/workspace/root/core/zip/inc/Compression.h /Users/vvassilev/workspace/root/core/clib/inc/strlcpy.h /Users/vvassilev/workspace/root/core/clib/inc/snprintf.h  /Users/vvassilev/workspace/root/core/metautils/inc/ESTLType.h /Users/vvassilev/workspace/root/core/metautils/inc/RConversionRuleParser.h /Users/vvassilev/workspace/root/core/metautils/inc/RStringView.h /Users/vvassilev/workspace/root/core/metautils/inc/TClassEdit.h /Users/vvassilev/workspace/root/core/metautils/inc/TClingRuntime.h /Users/vvassilev/workspace/root/core/metautils/inc/TSchemaRuleProcessor.h /Users/vvassilev/workspace/root/core/metautils/inc/TSchemaType.h /Users/vvassilev/workspace/root/core/metautils/inc/root_std_complex.h /Users/vvassilev/workspace/root/core/textinput/inc/Getline.h -I/Users/vvassilev/workspace/root /Users/vvassilev/workspace/root/core/base/inc/LinkDef.h 
ERROR in cling::CIFactory::createCI(): cannot extract standard library include paths!
Invoking:
  /Users/vvassilev/workspace/root_urgent/core/utils/src/clang
Results was:
sh: /Users/vvassilev/workspace/root_urgent/core/utils/src/clang: No such file or directory
With exit code 32512
input_line_1:1:10: fatal error: 'new' file not found
#include <new>
         ^
Warning in cling::IncrementalParser::CheckABICompatibility():
  Possible C++ standard library mismatch, compiled with _LIBCPP_VERSION v3700 but extraction of runtime standard library version failed.
In file included from input_line_4:1:
In file included from /Users/vvassilev/workspace/root_urgent/etc/cling/lib/clang/3.9.0/include/assert.h:8:
In file included from /usr/include/assert.h:42:
/usr/include/sys/cdefs.h:81:2: warning: "Unsupported compiler detected" [-W#warnings]
#warning "Unsupported compiler detected"
 ^
/usr/include/sys/cdefs.h:707:2: error: Unsupported architecture
#error Unsupported architecture
 ^
In file included from input_line_4:2:
In file included from /Users/vvassilev/workspace/root_urgent/etc/cling/lib/clang/3.9.0/include/stdlib.h:8:
In file included from /usr/include/stdlib.h:63:
In file included from /usr/include/_types.h:27:
In file included from /usr/include/sys/_types.h:33:
/usr/include/machine/_types.h:34:2: error: architecture not supported
#error architecture not supported
 ^
In file included from input_line_4:2:
In file included from /Users/vvassilev/workspace/root_urgent/etc/cling/lib/clang/3.9.0/include/stdlib.h:8:
In file included from /usr/include/stdlib.h:63:
In file included from /usr/include/_types.h:27:
/usr/include/sys/_types.h:55:9: error: unknown type name '__int64_t'; did you mean '__int128_t'?
typedef __int64_t       __darwin_blkcnt_t;      /* total blocks */
        ^
note: '__int128_t' declared here
/usr/include/sys/_types.h:56:9: error: unknown type name '__int32_t'; did you mean '__int128_t'?
typedef __int32_t       __darwin_blksize_t;     /* preferred block size */
        ^
note: '__int128_t' declared here
/usr/include/sys/_types.h:57:9: error: unknown type name '__int32_t'; did you mean '__int128_t'?
typedef __int32_t       __darwin_dev_t;         /* dev_t */
        ^
note: '__int128_t' declared here
/usr/include/sys/_types.h:60:9: error: unknown type name '__uint32_t'; did you mean '__uint128_t'?
typedef __uint32_t      __darwin_gid_t;         /* [???] process and group IDs */
        ^
note: '__uint128_t' declared here
/usr/include/sys/_types.h:61:9: error: unknown type name '__uint32_t'; did you mean '__uint128_t'?
typedef __uint32_t      __darwin_id_t;          /* [XSI] pid_t, uid_t, or gid_t*/
        ^
note: '__uint128_t' declared here
/usr/include/sys/_types.h:62:9: error: unknown type name '__uint64_t'; did you mean '__uint128_t'?
typedef __uint64_t      __darwin_ino64_t;       /* [???] Used for 64 bit inodes */
        ^
note: '__uint128_t' declared here
/usr/include/sys/_types.h:68:9: error: unknown type name '__darwin_natural_t'
typedef __darwin_natural_t __darwin_mach_port_name_t; /* Used by mach */
        ^
/usr/include/sys/_types.h:70:9: error: unknown type name '__uint16_t'; did you mean '__uint128_t'?
typedef __uint16_t      __darwin_mode_t;        /* [???] Some file attributes */
        ^
note: '__uint128_t' declared here
/usr/include/sys/_types.h:71:9: error: unknown type name '__int64_t'; did you mean '__int128_t'?
typedef __int64_t       __darwin_off_t;         /* [???] Used for file sizes */
        ^
note: '__int128_t' declared here
/usr/include/sys/_types.h:72:9: error: unknown type name '__int32_t'; did you mean '__int128_t'?
typedef __int32_t       __darwin_pid_t;         /* [???] process and group IDs */
        ^
note: '__int128_t' declared here
/usr/include/sys/_types.h:73:9: error: unknown type name '__uint32_t'; did you mean '__uint128_t'?
typedef __uint32_t      __darwin_sigset_t;      /* [???] signal set */
        ^
note: '__uint128_t' declared here
/usr/include/sys/_types.h:74:9: error: unknown type name '__int32_t'; did you mean '__int128_t'?
typedef __int32_t       __darwin_suseconds_t;   /* [???] microseconds */
        ^
note: '__int128_t' declared here
/usr/include/sys/_types.h:75:9: error: unknown type name '__uint32_t'; did you mean '__uint128_t'?
typedef __uint32_t      __darwin_uid_t;         /* [???] user IDs */
        ^
note: '__uint128_t' declared here
/usr/include/sys/_types.h:76:9: error: unknown type name '__uint32_t'; did you mean '__uint128_t'?
typedef __uint32_t      __darwin_useconds_t;    /* [???] microseconds */
        ^
note: '__uint128_t' declared here
In file included from input_line_4:2:
In file included from /Users/vvassilev/workspace/root_urgent/etc/cling/lib/clang/3.9.0/include/stdlib.h:8:
In file included from /usr/include/stdlib.h:65:
In file included from /usr/include/sys/wait.h:109:
In file included from /usr/include/sys/signal.h:82:
/usr/include/machine/signal.h:34:2: error: architecture not supported
#error architecture not supported
 ^
In file included from input_line_4:2:
In file included from /Users/vvassilev/workspace/root_urgent/etc/cling/lib/clang/3.9.0/include/stdlib.h:8:
In file included from /usr/include/stdlib.h:65:
In file included from /usr/include/sys/wait.h:109:
In file included from /usr/include/sys/signal.h:146:
/usr/include/machine/_mcontext.h:31:2: error: architecture not supported
#error architecture not supported
 ^
In file included from input_line_4:2:
In file included from /Users/vvassilev/workspace/root_urgent/etc/cling/lib/clang/3.9.0/include/stdlib.h:8:
In file included from /usr/include/stdlib.h:65:
In file included from /usr/include/sys/wait.h:109:
In file included from /usr/include/sys/signal.h:150:
/usr/include/sys/_types/_sigaltstack.h:39:2: error: unknown type name '__darwin_size_t'
        __darwin_size_t ss_size;        /* signal stack length */
        ^
fatal error: too many errors emitted, stopping now [-ferror-limit=]
Error: Error loading the default header files.
```

Could you try to compile ROOT on OSX and fix this issue. rootcling_tmp uses the CIFactory created compiler.

vgvassilev:
It seems that the content of `interpreter/cling//lib/Interpreter/cling-compiledata.h` differs from `interpreter/cling//lib/Interpreter/cling-compiledata.h.in`

I also get some warnings:
```
clang++ -g -MMD -MP -I/Users/vvassilev/workspace/root/interpreter/cling/include -I/Users/vvassilev/workspace/root_urgent/interpreter/llvm/inst/include -stdlib=libc++ -fPIC -pedantic -std=c++11 -DNDEBUG -fno-exceptions -fno-rtti -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -fno-strict-aliasing -DNDEBUG -Iinterpreter/cling/lib/Interpreter/ -pthread -o interpreter/cling/lib/Interpreter/CIFactory.o -c /Users/vvassilev/workspace/root/interpreter/cling/lib/Interpreter/CIFactory.cpp
/Users/vvassilev/workspace/root/interpreter/cling/lib/Interpreter/CIFactory.cpp:403:15: warning: 
      'gestaltSystemVersionMajor' is deprecated: first deprecated in OS X 10.8 - Use NSProcessInfo's
      operatingSystemVersion property instead. [-Wdeprecated-declarations]
      Gestalt(gestaltSystemVersionMajor, &majorVersion);
              ^
/System/Library/Frameworks/CoreServices.framework/Frameworks/CarbonCore.framework/Headers/Gestalt.h:2077:3: note: 
      'gestaltSystemVersionMajor' has been explicitly marked deprecated here
  gestaltSystemVersionMajor     = 'sys1', /* The major system version number; in 10.4.17 this would be the de...
  ^
/Users/vvassilev/workspace/root/interpreter/cling/lib/Interpreter/CIFactory.cpp:404:15: warning: 
      'gestaltSystemVersionMinor' is deprecated: first deprecated in OS X 10.8 - Use NSProcessInfo's
      operatingSystemVersion property instead. [-Wdeprecated-declarations]
      Gestalt(gestaltSystemVersionMinor, &minorVersion);
              ^
/System/Library/Frameworks/CoreServices.framework/Frameworks/CarbonCore.framework/Headers/Gestalt.h:2078:3: note: 
      'gestaltSystemVersionMinor' has been explicitly marked deprecated here
  gestaltSystemVersionMinor     = 'sys2', /* The minor system version number; in 10.4.17 this would be the de...
  ^
2 warnings generated.
```

vgvassilev:
We need to update Module.mk's CLINGCOMPDH target, too.

vgvassilev:
I also get similar behavior using the cmake build of ROOT.

marsupial:
Are you building from a previous build? I've had to blow away the CMake caches for cling.

vgvassilev:
Integrated. Thank you for your contribution!

PR #30, opened 2016-08-13
Don't add -nostdinc flag just because popen succeeded, make sure we g…

marsupial:
…ot what we were looking for first.


vgvassilev:
It seems still failing.

marsupial:
You're Travis builds only master...and thats what's failing.
This fixes that.

vgvassilev:
Thanks! Integrated. Could you in future split the subject line? More info could be found [here](http://chris.beams.io/posts/git-commit/).

PR #31, opened 2016-08-15
Check HostCXXI is empty rather than size() < 3 and fix git log hanging.

marsupial:


vgvassilev:
Integrated. It seems that our internal infrastructure is happy, too. Thanks for your contribution!

PR #32, opened 2016-08-15
Fix cling when built for 32 bit architectures.

marsupial:


vgvassilev:
Integrated. Thanks a lot for your contribution!

PR #33, opened 2016-08-16
Changes to build an expanded matrix.

marsupial:


.travis.yml
```diff
@@ -47,62 +36,145 @@ matrix:
...
+      addons:
+        apt:
+          sources: ['ubuntu-toolchain-r-test']
+          packages: ['g++-4.7']
+      compiler: gcc-4.7
 
-  # 1. Linux Clang Builds
+    - os: linux
+      env: DIST=pangolin COMPILER=g++-4.8 CCOMPILER=gcc-4.8
+      addons:
```
vgvassilev: Could you use the shorthand syntax here, the way it was before?
Eg. `addons: &gcc48` (this names a variable, which gets expanded if used as `*gcc48`). This will shorten the file and reduce the copy-paste-s.

.travis.yml
```diff
@@ -159,13 +224,19 @@ script:
       git push --force https://$OAUTH_TOKEN@github.com/${TRAVIS_REPO_SLUG} cling-nightlies
     fi
 
-    timeout $TIMEOUT tools/packaging/cpt.py --current-dev=tar --with-cling-url=https://github.com/$TRAVIS_REPO_SLUG --skip-cleanup
+    timeout $TIMEOUT tools/packaging/cpt.py --current-dev=tar --with-cling-url=https://github.com/$TRAVIS_REPO_SLUG --verbose --skip-cleanup
```
vgvassilev: Why do we need verbose?

.travis.yml
```diff
@@ -127,30 +199,23 @@ install:
...
-  chmod +x $CLING_LOCAL_BIN/$CC_COMPILER
-
-- export PATH=$CLING_LOCAL_BIN:$PATH
-- eval $EXECUTE_COMMAND
+ - $CMAKE --version
+ - export CXX="ccache $COMPILER"
+ - export CC="ccache $CCOMPILER"
 
 script:
+- $CXX --version
```
vgvassilev: Why do we need this here, it is printed at configuration time and upon a failure.

marsupial: I'm gonna cancel this pull request....
I think there's a way to get travis to use the cache I already have rather than rebuilding.

marsupial:
I'm gonna cancel this pull request....
I think there's a way to get travis to use the cache I already have rather than rebuilding.

marsupial:
Could you cancel the build in Travis if it doesn't do it automatically?

vgvassilev:
Sure. However I think that PR was good and going in a good direction.

marsupial:
It'll be back

marsupial:
Its dependent on https://github.com/root-mirror/cling/pull/26/commits/894d3143f79cec48a9aec1c7b4c4c70c89ee741d

CMake isn't finding the right compiler with ccache.

marsupial:
No really its best to do this later. The CMake issue causes all tests to fail at this point.
Theres a problem with InterpreterException that also must get resolved.
And its not pulling the from the best branch in my tree so the caches aren't being shared.
You'll have to babysit each build and restart it after timeout.

.travis.yml
marsupial: Must have slipped through conflict resolution.

.travis.yml
marsupial: Yes, but the referencing should go at the top of the file, not in individual entries...but I haven't had time to do that.

When builds fail it may only be a tiny number of entries in the matrix. Being able to snip out all of the passing tests and start the next build on only failing ones is easier when each entry is self contained. If the reference is tied to a specific entry I'd have to cut paste even more.

vgvassilev:
Integrated. Thank you for your contribution!

PR #34, opened 2016-08-16
Match exception specification for InterpreterException.

marsupial:
Some compilers are complaining about the mismatching exception specifications
between destructors of InterpreterException and its subclasses.


vgvassilev:
Integrated. Thank you for your contribution.

PR #38, opened 2016-08-20
Handle failures in initialization and add cling output features.

marsupial:
I can't split these as the second relies on the first.

### DIFF
--- a/lib/Interpreter/CIFactory.cpp
+++ b/lib/Interpreter/CIFactory.cpp
@@ -24,8 +24,11 @@
 #include "clang/Driver/Driver.h"
 #include "clang/Driver/Job.h"
 #include "clang/Driver/Tool.h"
+#include "clang/Frontend/FrontendDiagnostic.h"
 #include "clang/Frontend/TextDiagnosticPrinter.h"
 #include "clang/Frontend/VerifyDiagnosticConsumer.h"
+#include "clang/FrontendTool/Utils.h"
+#include "clang/Serialization/SerializationDiagnostic.h"
 #include "clang/Lex/Preprocessor.h"
 #include "clang/Sema/Sema.h"
 #include "clang/Sema/SemaDiagnostic.h"
@@ -597,6 +600,70 @@ static void stringifyPreprocSetting(PreprocessorOptions& PPOpts,
     return Diags;
   }
 
+  static bool SetupCompiler(CompilerInstance* CI, bool full) {
+    // Set the language options, which cling needs.
+    // This may have already been done via a precompiled header
+    if (full)
+      SetClingCustomLangOpts(CI->getLangOpts());
+
+    PreprocessorOptions& PPOpts = CI->getInvocation().getPreprocessorOpts();
+    SetPreprocessorFromBinary(PPOpts);
+
+    PPOpts.addMacroDef("__CLING__");
+    if (CI->getLangOpts().CPlusPlus11 == 1) {
+      // http://llvm.org/bugs/show_bug.cgi?id=13530
+      PPOpts.addMacroDef("__CLING__CXX11");
+    }
+
+    if (CI->getDiagnostics().hasErrorOccurred()) {
+      cling::errs() << "Compiler error to early in initialization.\n";
+      return false;
+    }
+
+    CI->setTarget(TargetInfo::CreateTargetInfo(CI->getDiagnostics(),
+                                               CI->getInvocation().TargetOpts));
+    if (!CI->hasTarget()) {
+      cling::errs() << "Could not determine compiler target.\n";
+      return false;
+    }
+
+    CI->getTarget().adjust(CI->getLangOpts());
+
+    // This may have already been done via a precompiled header
+    if (full)
+      SetClingTargetLangOpts(CI->getLangOpts(), CI->getTarget());
+
+    SetPreprocessorFromTarget(PPOpts, CI->getTarget().getTriple());
+    return true;
+  }
+
+  class ActionScan {
+    std::set<const clang::driver::Action*> m_Visited;
+    const clang::driver::Action::ActionClass m_Kind;
+
+    bool find (const clang::driver::Action* A) {
+      if (A && !m_Visited.count(A)) {
+        if (A->getKind() == m_Kind)
+          return true;
+
+        m_Visited.insert(A);
+        return find(*A->input_begin());
+      }
+      return false;
+    }
+
+  public:
+    ActionScan(clang::driver::Action::ActionClass k) : m_Kind(k) {}
+
+    bool find (clang::driver::Compilation* C) {
+      for (clang::driver::Action* A : C->getActions()) {
+        if (find(A))
+          return true;
+      }
+      return false;
+    }
+  };
+  
   static CompilerInstance*
   createCIImpl(std::unique_ptr<llvm::MemoryBuffer> Buffer,
                const CompilerOptions& COpts, const char* LLVMDir,
@@ -645,8 +712,11 @@ static void stringifyPreprocSetting(PreprocessorOptions& PPOpts,
       }
 #endif
 
-    argvCompile.push_back("-c");
-    argvCompile.push_back("-");
+    const bool genOutput = COpts.HasOutput && !OnlyLex;
+    if (!genOutput) {
+      argvCompile.push_back("-c");
+      argvCompile.push_back("-");
+    }
 
     std::unique_ptr<clang::CompilerInvocation>
       InvocationPtr(new clang::CompilerInvocation);
@@ -691,6 +761,20 @@ static void stringifyPreprocSetting(PreprocessorOptions& PPOpts,
     CI->setInvocation(InvocationPtr.get());
     InvocationPtr.release();
     CI->setDiagnostics(Diags.get()); // Diags is ref-counted
+
+    if (genOutput) {
+      ActionScan scan(clang::driver::Action::PrecompileJobClass);
+      if (!scan.find(Compilation.get())) {
+        cling::errs() << "Only output to precompiled header is supported.\n";
+        return nullptr;
+      }
+      if (!SetupCompiler(CI.get(), true))
+        return nullptr;
+
+      ProcessWarningOptions(*Diags, DiagOpts);
+      ExecuteCompilerInvocation(CI.get());
+      return CI.release();
+    }
 
     CI->createFileManager();
     clang::CompilerInvocation& Invocation = CI->getInvocation();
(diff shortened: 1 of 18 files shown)


karies:
That's an awesome set of features! Thank you!

lib/Interpreter/CIFactory.cpp
```diff
@@ -928,29 +961,132 @@ namespace {
   }
 
   static void AddRuntimeIncludePaths(llvm::StringRef ClingBin,
-                                     clang::HeaderSearchOptions& HOpts) {
-    if (HOpts.Verbose)
-      llvm::errs() << "Adding runtime include paths:\n";
-    // Add configuration paths to interpreter's include files.
+                                    std::vector<const char*>& args,
+                                    bool Verbose) {
+    static std::vector<std::string> sArguments;
```
vgvassilev: Could you implement that without statics? Statics are troublesome for multi interpreter setups.

lib/Interpreter/CIFactory.cpp
```diff
@@ -928,29 +961,132 @@ namespace {
...
+    }
+
+    if (Verbose) {
+      llvm::errs() << "Added runtime include paths:\n";
+      for (const std::string& arg : sArguments)
+        llvm::errs() << "  " << arg << "\n";
+    }
+  }
+
+  static llvm::IntrusiveRefCntPtr<DiagnosticsEngine>
```
vgvassilev: Could you implement that without statics? Statics are troublesome for multi interpreter setups.

lib/Interpreter/CIFactory.cpp
```diff
@@ -999,35 +1137,46 @@ namespace {
...
+    }
+
+    std::unique_ptr<clang::CompilerInvocation>
+      InvocationPtr(new clang::CompilerInvocation);
+    if (!InvocationPtr)
+      return nullptr;
 
-    clang::CompilerInvocation*
-      Invocation = new clang::CompilerInvocation;
+    using llvm::IntrusiveRefCntPtr;
```
vgvassilev: I'd prefer the verbosity of `llvm::IntrusiveRefCntPtr`.

lib/Interpreter/IncrementalParser.cpp
```diff
@@ -182,22 +192,32 @@ namespace cling {
...
+      llvm::errs() << "TransactionPool could not be created.\n";
+      return false;
+    }
+
     if (hasCodeGenerator()) {
       getCodeGenerator()->Initialize(getCI()->getASTContext());
       m_BackendPasses.reset(new BackendPasses(getCI()->getCodeGenOpts(),
                                               getCI()->getTargetOpts(),
                                               getCI()->getLangOpts()));
+      if (!m_BackendPasses) {
```
vgvassilev: Let's turn these into asserts. If we cannot create `BackendPasses` and `m_TransactionPool` we are doomed anyway.

lib/Interpreter/IncrementalParser.cpp
```diff
@@ -206,46 +226,65 @@ namespace cling {
...
     if (!PCHFileName.empty()) {
       Transaction* CurT = beginTransaction(CO);
+      DiagnosticErrorTrap Trap(m_CI->getSema().getDiagnostics());
       m_CI->createPCHExternalASTSource(PCHFileName,
                                        true /*DisablePCHValidation*/,
                                        true /*AllowPCHWithCompilerErrors*/,
                                        0 /*DeserializationListener*/,
                                        true /*OwnsDeserializationListener*/);
+
+      // TODO: Better diagnosis of whether the error is safe to ignore.
```
vgvassilev: Could you transform that into FIXME, we should be using FIXME all over for easier grepping. Also, it seems this check is enough. If the error recovery cannot unload a transaction it should set this bit to bail out.

lib/Interpreter/IncrementalParser.cpp
```diff
@@ -206,46 +226,65 @@ namespace cling {
...
+        if (External)
+          External->StartTranslationUnit(m_Consumer);
+
+        // If I belong to the parent Interpreter, only then do
+        // the #include <new>
+        if (!isChildInterpreter && m_CI->getLangOpts().CPlusPlus) {
+          // <new> is needed by the ValuePrinter so it's a good thing to include it.
+          // We need to include it to determine the version number of the standard
+          // library implementation.
+          ParseInternal("#include <new>");
```
vgvassilev: Why don't we include the header with the version only?

lib/Interpreter/Interpreter.cpp
```diff
@@ -163,8 +171,15 @@ namespace cling {
...
     m_LLVMContext.reset(new llvm::LLVMContext);
+    if (!m_LLVMContext)
+      return;
     m_DyLibManager.reset(new DynamicLibraryManager(getOptions()));
+    if (!m_DyLibManager)
+      return;
+
     m_IncrParser.reset(new IncrementalParser(this, llvmdir));
+    if (!m_IncrParser || !m_IncrParser->isValid(false))
+      return;
```
vgvassilev: Please turn those into asserts, too.

vgvassilev:
Could you also add the pch and `-E` support in the release notes?

lib/Interpreter/Interpreter.cpp
marsupial: That is now a valid state, for output to work we need to support some level of the Interpreter returning from it's constructor gracefully when it wasn't fully initialized.

marsupial:
I'm not sure I'm understanding the logic behind replacing actual error handling with asserts.
It only does something in Debug, in Release it's totally silent. 
If you use cling as a library, it aborts the running process.
There are valid cases for process to be fine if cling couldn't initialize an Interpreter fully, writing output through clang is one of them.

I understand ROOT is the primary client and you want to keep it happy, but I think it's the hosts responsibility to make sure it is setup went fine.  Hence the isValid methods.

Is it too problematic to add assert(m_Interp->isValid() && "Initialization failed") in ROOT?

lib/Interpreter/IncrementalParser.cpp
marsupial: I'm sorry, I don't understand this.

lib/Interpreter/CIFactory.cpp
marsupial: I generally agree, but was following the pattern of AddHostArguments.
I thought it did statics for two reasons:
 Avoid redoing work.
 Make sure all children get the same flags as parent. In an interpreted environment there's no guarantee that would be true across calls if they are not static.

vgvassilev: Ok, that makes sense. In this case we force slave interpreters to be with the same invocation flags. Could you explain this constraint in a comment.

lib/Interpreter/IncrementalParser.cpp
vgvassilev: `<new>` is a c++ header. It would break (even more) cling run in non-c++ mode. I was suggesting instead to include <new> to include c++config (defining the version macro). This would improve the construction time of the interpreter (we don't parse extra things unless we want to print values out). We probably should do that only if we are in c++ mode.

Shouldn't we be doing something like that for __GLIBC__, too?

lib/Interpreter/Interpreter.cpp
vgvassilev: I am not sure I understand 'for output to work'. Is there a particular use case for this?

lib/Interpreter/Interpreter.cpp
```diff
@@ -175,18 +190,29 @@ namespace cling {
...
+    }
 
     // Tell the diagnostic client that we are entering file parsing mode.
     DiagnosticConsumer& DClient = getCI()->getDiagnosticClient();
     DClient.BeginSourceFile(getCI()->getLangOpts(), &PP);
 
     llvm::SmallVector<IncrementalParser::ParseResultTransaction, 2>
       IncrParserTransactions;
-    m_IncrParser->Initialize(IncrParserTransactions, isChildInterp);
+    if (!m_IncrParser->Initialize(IncrParserTransactions, isChildInterp)) {
```
vgvassilev: I am opposed to this approach in general. I think we should always have a well-constructed and initialized interpreter, this reduces a lot of debugging. What would be the use case here?

vgvassilev:
Indeed we could add this assertion on the user side. One of the issues would be that we don't really know where exactly it failed. While I am inclined to allow this in some places in cling, I don't feel that the interpreter initialization time should be one of them. Unless there is some very obvious reason or a use case which I am missing.

For the library case: Aborting the process is not always a bad thing, especially when we know we are not in a valid state, thus most likely the produced results would be bogus. If we want to continue no matter what, think we need to implement a macro that expands asserts to printing errors and continuing the execution (where possible). This way we would be able to survive asserts in llvm and clang, too. I'd prefer this approach.

lib/Interpreter/IncrementalParser.cpp
marsupial: I think what you want is ciso646.
I know it currently works for libstdc++ and libc++, but that wasn't always the case and I don;t know if this is true on Windows either.

lib/Interpreter/Interpreter.cpp
marsupial: These commits add features where cling now has a mode to output files, to accomplish that without a lot of interface changes there needs to be a way to deal with the constructors understanding initialization of a full interpreter may not have occurred...more on that below.

marsupial:
> One of the issues would be that we don't really 

Well I've been logging to llvm::errs() in the places where failure can be detected/logged.

> I don't feel that the interpreter initialization time should be one of them. Unless there is some very obvious reason or a use case which I am missing.

I'm of the opinion initialization has to be one of the cases where failure can occur without aborting.

Think of a web-browser, would it be ok to crash the whole thing just because a JavaScript engine wasn't properly initialized, or should that failure be detected and the page rendered differently?

Even more important to me is that users will have work (possibly unsaved) before libCling is invoked.  If libCling fails: throwing a warning is fine, but flat out abort is unacceptable. I think beyond cling & ROOT this would be the majority of use cases too, you're a subsystem not the owning process.

I do agree that aborting is acceptable in some case, but see them as extreme cases where any consistancy can no longer be guaranteed.

I see the real issue begin that construction & initialization is intertwined, not only between one object, but several.  But any change to this would be a much larger interface change than proposed here.

lib/Interpreter/IncrementalParser.cpp
pcanal: In my opinion, assert should be reserved to the case where a contract is broken (user is passing *unallowed/unsupported* values or something *impossible* (i.e. as far as the intent of the code is concerned the code path is thought to not be reacheable).

Another consideration is that if a hard expectation is met (i.e. with the new code, m_BackendPasses can be null, the rest of the code needs to be updated to support this case by failing elegantly)

In the code here, the only case where 'if (!m_BackendPasses)' could be true is if the system is running out of memory.  So another option here is to throw an exception.

marsupial:
Looking at it the changes you're requesting would require:
1. Another CIFactory::createCI variant
2. createCIImpl being split into two functions.
3. Movement of InvocationOptions out of Interpreter
4. Addition of canAbort argument (could default to true) to both Interpeter and IncrementalParser constructors

While I generally agree with 1-3:
3 would be a larger interface change.
4 is a bit ugly.

lib/Interpreter/IncrementalParser.cpp
marsupial: I generally agree with all of this. My only issue is that if Interpreter::Interpreter is the only user of IncrementalParser::Initialize I know of, and if Interpreter::Interpreter fails for whatever reason, I need users to be able to continue working until they decide to quit, no segfault, no aborting.

Unless libstd++ or libc++ was built with -f-no-exceptions, then any failure of new should be throwing std::bad_alloc anyway.

As the code was it was blindly assuming everything succeeded, so could wind up dereference null pointer which is a crash. And asserting won't change this behavior when built in Release.

lib/Interpreter/CIFactory.cpp
```diff
@@ -999,35 +1140,46 @@ namespace {
...
     Driver.setCheckInputsExist(false); // think foo.C(12)
     llvm::ArrayRef<const char*>RF(&(argvCompile[0]), argvCompile.size());
     std::unique_ptr<clang::driver::Compilation>
       Compilation(Driver.BuildCompilation(RF));
-    const clang::driver::ArgStringList* CC1Args
-      = GetCC1Arguments(Diags.get(), Compilation.get());
-    if (CC1Args == NULL) {
-      delete Invocation;
-      return 0;
+    if (!Compilation) {
```
pcanal: Can we use a variable name that is 'distinct' from a class name (to avoid confusion between constructor and variable)?

vgvassilev:
Could you rebase this PR, please?

Would this be able to handle:
```
clang ...opts... -E > PP.out
cling -include PP.out
```
This would greatly help reducing cling-related issues.

marsupial:
I'm not sure if you asking whether it records the session or output can be fed back in.
Former no, later yes.
Which also makes it possible to use the preprocessor in tests.
`cling -C -E File.C > PPFile.C`
`cat PPFile.C | cling | FileCheck PPFile.C`

vgvassilev:
@marsupial that should be good enough and it could be improved in future. Please rebase to continue the review and merge it in.

marsupial:
This has been rebased to master.
 I'd rather not have to go through that again so at a minimum there should be a definitive decision whether it will get in or not.

karies:
Thanks. I have discussed with Vassil; this is good to go. I'll test before pushing. If I have any further comments it'll be either due to a test failure or after the merge.

karies:
Merged, thanks a lot!

Could you provide a couple of lines describing this PR's features, for the announcement of  the next cling release?

marsupial:
Are the additions in the ReleaseNotes not enough?

karies:
Yes they are - and that's probably also why I subconsciously thought of it. Thanks!

PR #39, opened 2016-08-23
CMake: Fix --gcc-toolchain flag not being propagated

marsupial:
This was broken in b8b4bec (ROOT: 965fb30) when support for ccache and distcc was added,
and affected cling being compiled with clang.


vgvassilev:
Integrated. Thanks for your contribution!

PR #41, opened 2016-09-01
Don't wrap function, method, constructor, or destructor definitions.

marsupial:


karies:
This PR triggers a failure in ROOT, reproducible like this on the cling prompt:
```
#include "cling/Interpreter/Interpreter.h"
gCling->process("#ifdef __UNDEFINED__\n42\n#endif")
```

which gives happiness in the master, and with this PR
```
input_line_5:6:1: error: extraneous closing brace ('}')
}
^
(cling::Interpreter::CompilationResult) (cling::Interpreter::CompilationResult::kFailure) : (unsigned int) 1
```

Could you have a look at it?

marsupial:
Fixed and added the test case.

karies:
That seems to make ROOT happy - thanks a *lot* for this dramatic improvement for many cling use cases!

Merged!

PR #43, opened 2016-09-05
Windows Fixes

marsupial:


lib/Interpreter/LookupHelper.cpp
```diff
@@ -387,9 +387,9 @@ namespace cling {
                                 quickTypeName.size()-6);
       innerConst = true;
     }
-    constexpr int pointerType = 0;
-    constexpr int lrefType = 1;
-    constexpr int rrefType = 2;
+
+    enum PointerType { kPointerType, kLRefType, kRRefType, };
```
pcanal: How is this change related to noexcept/constexpr?  If it isn't, can we split it into its own commit?  Thanks.

marsupial: Only related in prior history of changes that introduced the use of constexpr.
Splitting it out should be easy enough.

PR #45, opened 2016-09-07
Windows Fixes

marsupial:


include/cling/Interpreter/Exception.h
```diff
@@ -10,15 +10,8 @@
...
-
-#ifndef CLING_NOEXCEPT
-# if defined(_MSC_VER) && (_MSC_VER < 1900)
-#  define CLING_NOEXCEPT
-# else
-#  define CLING_NOEXCEPT noexcept
-# endif
-#endif
+#include "llvm/Support/Compiler.h"
+#include <stdexcept>
```
karies: Good catch, thanks!

karies:
Wonderful, thanks! Merged!

lib/Interpreter/CIFactory.cpp
```diff
@@ -240,102 +241,103 @@ static bool getUniversalCRTSdkDir(std::string &Path,
...
+static void logSearch(const char* Name, const std::string& Value,
+                      const char* Found = nullptr) {
+  if (Found)
+    llvm::errs() << "Found " << Name << " '" << Value << "' that matches "
+                 << Found << " version\n";
+  else
+    llvm::errs() << Name << " '" << Value << "' not found.\n";
+}
+
+static void trimString(const char* Value, const char* SubStr,
```
vgvassilev: How is this different from llvm::StringRef::ltrim?

marsupial: llvm::StringRef::ltrim does individual and consecutive characters.
This is more like (which does look nicer, but has the overhead of two strlens):

```
llvm::StringRef ValueRef(Value);
Path = ValueRef.substr(0, ValueRef.find(SubStr)).str();
```

PR #46, opened 2016-09-07
Fallback to getting system version with sw_vers

marsupial:


lib/Interpreter/CIFactory.cpp
```diff
@@ -893,16 +902,64 @@ namespace {
...
 #endif
 
-// https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html
 #ifdef _GLIBCXX_USE_CXX11_ABI
+  // https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html
     PPOpts.addMacroDef("_GLIBCXX_USE_CXX11_ABI="
                        ClingStringify(_GLIBCXX_USE_CXX11_ABI));
 #endif
+#ifdef _GLIBCXX_USE_NANOSLEEP
+  // http://stackoverflow.com/questions/12523122/what-is-glibcxx-use-nanosleep-all-about
```
karies: I find that very specific. Why should this be in cling's setup (in contrast to user code)?

lib/Interpreter/CIFactory.cpp
```diff
@@ -893,16 +902,64 @@ namespace {
     // of the two is the underlying compiler.
 
 #ifdef __clang__
-    PPOpts.addMacroDef("__CLING__clang__=" ClingStringify(__clang__));
+  #if defined(__apple_build_version__)
+     // __apple_build_version__ >= 8000031 Xcode 8
+     #if __apple_build_version__ >= 7030029
```
karies: Why can we not just #define `__CLING__clang__` as `__clang__`? I don't see it as cling's responsibility to bring sanity in version macros...

lib/Interpreter/CIFactory.cpp
marsupial: It can wind up blocking std::this_thread::sleep_for when not defined.
Ran into this issue and thought it helpful that other's don't have to.
Don't really care if it's in or out.
Let me know.

karies: As cling doesn't seem to use `std::this_thread::sleep_for` itself I'd argue that it is the user's responsibility to handle this.  Cling shouldn't try to fix people's (use of) libstdc++. Based on my current understanding (which might be partial / wrong) I'd prefer to have it out.

lib/Interpreter/CIFactory.cpp
marsupial: Because of Apple's macro's there's no good way to get the version on clang compiled with.  It's a nice opportunity to be more specific, and this makes it much easier to get the version in cling. __clang__ is already defined as __clang__ so it's not losing any information.

karies: In cling, `__clang__` is `1`, because cling uses clang as runtime-compiler.

We should define `__CLING__clang__` to be the value of `__clang__` when compiling CIFactory.cxx. We can define `__CLING__apple_build_version__` to the value of `__apple_build_version__` if that's really really useful when using cling.

But from what I see, you are changing the meaning of `__CLING__clang__`; you are providing information (setting the value to `__apple_build_version__`) that is not used by cling itself.

I understand that it would have been nice by Apple to come up with a more useful versioning scheme - but is it cling's job to repair that?

karies: More arguments against this:
- nobody should check against the clang version, but against compiler features
- I can almost guarantee that we will forget to maintain this in the future.

I see that Xcode 8 defines the clang version as 8.0.0.38, not mentioning any version number of the underlying clang. That's the versioning mess that I was referring to - and I do not see it as cling's task to fix that :-)

marsupial: The changes to the macros are here because the commit for falling back if a deprecated method is removed wouldn't apply cleanly to master without that commit which I authored because I needed the info.  If you don't want it thats fine.

karies:
Thanks, much appreciated! It's now merged! (Module small whitespace corrections.)

PR #47, opened 2016-09-09
Fix missing call to pclose, introduced in 877a1f4.

marsupial:
Sorry, would have been nice to squash this, but it was hiding further up the tree than I was looking.
At least conflicts are useful sometimes!


karies:
Yeah sorry here, too - I should have actually noticed that during review :-( Anyway - thanks for the fix, merged!

PR #48, opened 2016-09-11
Fix TransactionPool

marsupial:
Transaction::~Transaction being called twice.
Remove extra Transaction allocation/leak per Transaction.

### DIFF
--- a/lib/Interpreter/TransactionPool.h
+++ b/lib/Interpreter/TransactionPool.h
@@ -30,62 +30,48 @@ namespace cling {
     //
     llvm::SmallVector<Transaction*, POOL_SIZE>  m_Transactions;
 
-    ///\brief The Sema required by cling::Transactions' ctor.
-    ///
-    clang::Sema& m_Sema;
-
-    // We need to free them in blocks.
-    //
-    //llvm::SmallVector<Transaction*, 64> m_TransactionBlocks;
-#ifndef NDEBUG
-    bool m_Debug;
-#endif
-
-  private:
   public:
-    TransactionPool(clang::Sema& S) : m_Sema(S) {
-#ifndef NDEBUG
-      m_Debug = false;
-#endif
-    }
-
+    TransactionPool() {}
     ~TransactionPool() {
-      for (size_t i = 0, e = m_Transactions.size(); i < e; ++i)
-        delete m_Transactions[i];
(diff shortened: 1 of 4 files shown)


lib/Interpreter/TransactionPool.h
```diff
@@ -20,74 +20,59 @@ namespace clang {
...
 
-      T->m_State = Transaction::kCollecting;
       return T;
     }
 
-    void releaseTransaction(Transaction* T) {
-      assert((T->getState() == Transaction::kCompleted ||
-              T->getState() == Transaction::kRolledBack)
-             && "Transaction must completed!");
+    void releaseTransaction(Transaction* T, bool reuse = true) {
```
karies: The original code had an `#ifndef NDEBUG`, to force allocation of new transactions. Can we keep this in some way, maybe setting `reuse` to false?

marsupial: I can.
Out of curiosity what is the reasoning?

marsupial: For the comment.

karies: We cat set watch points if transaction number N always ends up at a given address. If they are re-used then the transaction N is at the same memory address as a gazillion transactions before. That makes watch points much more difficult to use...

marsupial: Added.

lib/Interpreter/TransactionPool.h
```diff
@@ -52,7 +57,7 @@ namespace cling {
       return T;
     }
 
-    void releaseTransaction(Transaction* T, bool reuse = true) {
+    void releaseTransaction(Transaction* T, bool reuse = !kDebugMode) {
```
karies: For my brain, that makes the logic more complex than needed: this setting should override the value passed as `reuse`  if `reuse` is `true` and `NDEBUG` is off. I.e. in `#ifndef NDEBUG`, `reuse` shall be `false`, whatever the argument passed. With your patch I'll need to check that no caller calls it with an explicit `true`, now or in the future.

Can we "hard-wire" this by overriding the value of `reuse = false` in `releaseTransaction`'s body for `#ifndef NDEBUG`?

marsupial: Ok?

karies:
Perfect, thanks! Merged!

PR #49, opened 2016-09-12
Move DeclUnloader and ParserStateRAII into cling namespace.

marsupial:
Requires changes to clang here:
https://github.com/marsupial/clang/commit/bb84e42bcd751fdc10024b9e2a0c66c3c7afadc1

### DIFF
--- a/lib/Interpreter/DeclUnloader.cpp
+++ b/lib/Interpreter/DeclUnloader.cpp
@@ -23,9 +23,8 @@
 
 #include "llvm/IR/Constants.h"
 
-// FIXME: rename back to cling when gcc fix the
-// namespace cling { using cling::DeclUnloader DeclUnloader} bug
-namespace clang {
+namespace cling {
+using namespace clang;
 
 bool DeclUnloader::isDefinition(TagDecl* R) {
   return R->isCompleteDefinition() && isa<CXXRecordDecl>(R);
@@ -173,8 +172,6 @@ bool DeclUnloader::VisitRedeclarable(clang::Redeclarable<T>* R, DeclContext* DC)
 #endif
   return true;
 }
-
- using namespace cling;
 
   // Copied and adapted from GlobalDCE.cpp
   class GlobalValueEraser {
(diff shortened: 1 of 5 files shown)


marsupial:
Added additional commit in front to keep recent CMake changes closer/nicer in repo.

marsupial:
Added additional commits for comments, and conflict resolution.

marsupial:
But who knows if or when that would get in?
I'm of the opinion things should be in their proper namespace, both for the obvious reasons and it makes it clearer for new authors to understand which code is cling specific.

karies:
@vgvassilev "Can we simplify the patch in clang by reverting +friend class ::cling::ParserStateRAII;? We should submit the ParserStateRAII to clang's mainline." - yes, but should we block further development for something we plan to do since what - three years? I.e. we should do what you suggest but until it's in we should keep going.

vgvassilev:
@marsupial I am positive that I could get it on the fast lane for a review and check it in.
@karies How keeping ParserStateRAII where it is blocks you. It seems we are generating extra work from refactorings which we agree are temporary.

marsupial:
Even if it makes it into clang at most you'd have avoided editing all of 3 lines...this conversation has taken longer than that would have!

If it makes into into clang I have no problem fixing this if it made it into cling or still a PR, but I'm going to have to decline doing the extra work on a possibility it gets in. 

FWIW I submitted this because unless these changes make it into master I have to continually cherry-pick it onto the tip of every branch so I can test locally before submitting PRs. Having them bundled into one commit is just much easier to manage.

vgvassilev:
I'd have really preferred our patches to clang to become less not more. Piling up patches makes upgrades to newer llvm/clang versions really hard.

If this makes things really more convenient for you I am inclined to accept it.

karies:
...and just a side note: once it's accepted into clang we'll still need to bump cling's llvm/clang revision to benefit from it. So the turn-around even for fast-path is weeks.

karies:
Sorry think fingers didn't mean to close...

vgvassilev:
Ok, let's integrate it and I will try to offload this patch in clang. @karies could you take over this PR while I am working on submitting the patch to clang?

karies:
Taken. LGT*Y*, @vgvassilev ?

vgvassilev:
Yes.

karies:
Merged, together with your clang patch marsupial/clang@bb84e42 - thanks!

PR #52, opened 2016-09-16
extern "C" declarations & cruft removal

marsupial:
Allow extern "C" declarations to be called from .x command.
Fix some cruft and now invalid comments built up from commits earlier this week.

### DIFF
--- a/lib/Interpreter/Transaction.cpp
+++ b/lib/Interpreter/Transaction.cpp
@@ -62,12 +62,22 @@ namespace cling {
   }
 
   NamedDecl* Transaction::containsNamedDecl(llvm::StringRef name) const {
-    for (auto I = decls_begin(), E = decls_end(); I < E; ++I)
+    for (auto I = decls_begin(), E = decls_end(); I < E; ++I) {
       for (auto DI : I->m_DGR) {
-        if (NamedDecl* ND = dyn_cast<NamedDecl>(DI))
+        if (NamedDecl* ND = dyn_cast<NamedDecl>(DI)) {
           if (name.equals(ND->getNameAsString()))
             return ND;
+        }
+        else if (LinkageSpecDecl* LSD = dyn_cast<LinkageSpecDecl>(DI)) {
+          for (Decl* DI : LSD->decls()) {
+            if (NamedDecl* ND = dyn_cast<NamedDecl>(DI)) {
+              if (name.equals(ND->getNameAsString()))
+                return ND;
+            }
+          }
+        }
       }
+    }
     return 0;
   }
 
--- a/include/cling/Interpreter/Transaction.h
+++ b/include/cling/Interpreter/Transaction.h
@@ -185,12 +185,7 @@ namespace cling {
     ///
     clang::FileID m_BufferFID;
 
-    ///\brief This is all to support allocation via TransactionPool.
-    /// There is currently no way for TransactionPool to mark if a Transaction
-    /// originated there or elsewhere, so if this interface is needed
-    /// then TransactionPool will have to be rewritten (or even removed
-    /// as the performance gains seem negligable)
-    ///
+    /// TransactionPool needs direct access to m_State as setState asserts
     friend class TransactionPool;
 
     void Initialize(clang::Sema& S);
--- a/lib/Interpreter/InvocationOptions.cpp
+++ b/lib/Interpreter/InvocationOptions.cpp
@@ -65,11 +65,8 @@ namespace {
     Opts.NoLogo = Args.hasArg(OPT__nologo);
     Opts.ShowVersion = Args.hasArg(OPT_version);
     Opts.Help = Args.hasArg(OPT_help);
-    if (Args.hasArg(OPT__metastr, OPT__metastr_EQ)) {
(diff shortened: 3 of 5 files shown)


karies:
Thanks, merged!

lib/Interpreter/Transaction.cpp
```diff
@@ -62,12 +62,22 @@ namespace cling {
...
   NamedDecl* Transaction::containsNamedDecl(llvm::StringRef name) const {
-    for (auto I = decls_begin(), E = decls_end(); I < E; ++I)
+    for (auto I = decls_begin(), E = decls_end(); I < E; ++I) {
       for (auto DI : I->m_DGR) {
-        if (NamedDecl* ND = dyn_cast<NamedDecl>(DI))
+        if (NamedDecl* ND = dyn_cast<NamedDecl>(DI)) {
           if (name.equals(ND->getNameAsString()))
             return ND;
+        }
+        else if (LinkageSpecDecl* LSD = dyn_cast<LinkageSpecDecl>(DI)) {
```
vgvassilev: @marsupial why don't we do a DeclContext::lookup here? This would make the operation O(1).

marsupial: That's where the extern "C" decls be hiding.
I guess you could do a separate loop after the first one has failed?

vgvassilev: I don't think we should be finding and calling hidden by C++ lookup rules decls. Why should we find decls that the language hid intentionally?

vgvassilev: To be clear I meant:
```
else if (DeclContext* DC = dyn_cast<DeclContext(DI))
  DC->lookup()...
```

marsupial: Searching every DeclContext would take much longer and I don't think you want to resolve the name against classes, C-style structs, namespaces.

Completing the search for a -strict- match once does seem perhaps the more prudent thing.

karies: These two things have different semantics. lookup looks through all occurrences (and extern "C" is transparent for lookup so that will be a lot), while this function is meant to find Decls that are part of that specific transaction. I.e. we need the current code (incl @marsupial's improvement); lookup is not what we need.

vgvassilev: DeclContext::lookup is not a complete C++ lookup. We don't need to search every DeclContext but only the current one.

marsupial: What is the current DeclContext for a Transaction?

I'm happy to submit a patch that does two separate passes, as you are right this can lead to a substantial penalty in it's current form (inclusion of any C headers for example).

vgvassilev: I am thinking how to make this interface work in all possible cases. One of the problems is that we can find only first level declarations, which makes the interface really awkward. We will need some sort of visitor but this makes the implementation complicated. IIRC, the only use is in MetaSema for diagnostic purposes which can be replaced with a normal C++ lookup. Then we can ask the Transaction whether it contains the found Decl* (which is a far more simple job).

Another approach would be to simplify and speed up its implementation of Transaction::containsNamedDecl by performing a SemaLookup and check if the found Decl* were in the range of the transaction. We know the range of the Decl* in a transaction because of the bump allocator used for allocation. Eg. if the pointer is between the pointer of Transaction's first declaration and last declaration that would mean that this transaction contains the Decl (something like clang::DeclContext::containsDecl).

karies: O now I understand. But that's not what this interface is used for, so maybe it should be renamed. For its use it's useless / wrong to do a lookup.

vgvassilev: There is only one (mis?)use and I was wondering whether we should fix the use or keep this interface. Right now it is a misnomer, because it checks if the transaction contains a decl, but also returns it (which is a sort-of lookup).

It seems reasonable to me to rename this into `bool cling::Transaction::containsDecl(Decl*);` and add a Sema lookup in our `MetaSema`.

karies:
That single use is a use not a misuse: misuse can only happen if everyone else it using it differently.  So I don't understand that point if it was meant to be one. Instead, back to fixing the name: `findTopLevelNamedDecl`? I don't want two searches (`containsDecl()` and Sema lookup) where one is enough.

lib/Interpreter/Transaction.cpp
karies: That single use is a use not a misuse: misuse can only happen if everyone else it using it differently.  So I don't understand that point if it was meant to be one. Instead, back to fixing the name: `findTopLevelNamedDecl`? I don't want two searches (`containsDecl()` and Sema lookup) where one is enough.

karies: That's a good improvement for a future PR. Shall we add a FIXME?

vgvassilev: I'd prefer to get it fixed as part of the post-merge review, instead adding a FIXME.

vgvassilev: Yes please open a new PR.

marsupial: https://github.com/root-mirror/cling/pull/54

PR #53, opened 2016-09-17
Rip out platform stuff into separate modules.

marsupial:
There are some fixes along the way.

### DIFF
--- a/include/cling/Utils/Paths.h
+++ b/include/cling/Utils/Paths.h
@@ -52,7 +52,7 @@ namespace cling {
     bool SplitPaths(llvm::StringRef PathStr,
                     llvm::SmallVectorImpl<llvm::StringRef>& Paths,
                     SplitMode Mode = kPruneNonExistant,
-                    llvm::StringRef Delim = llvm::StringRef(platform::kEnvDelim),
+                    llvm::StringRef Delim = platform::kEnvDelim,
                     bool Verbose = false);
 
     ///\brief Adds multiple include paths separated by a delimter into the
--- a/include/cling/Utils/Paths.h
+++ b/include/cling/Utils/Paths.h
@@ -25,6 +25,12 @@ namespace clang {
 namespace cling {
   namespace utils {
 
+    namespace platform {
+      ///\brief Platform specific delimiter for splitting environment variables.
+      /// ':' on Unix, and ';' on Windows
+      extern const char* const kEnvDelim;
+    }
+    
     enum SplitMode {
       kPruneNonExistant,  ///< Don't add non-existant paths into output
       kFailNonExistant,   ///< Fail on any non-existant paths
@@ -46,7 +52,7 @@ namespace cling {
     bool SplitPaths(llvm::StringRef PathStr,
                     llvm::SmallVectorImpl<llvm::StringRef>& Paths,
                     SplitMode Mode = kPruneNonExistant,
-                    llvm::StringRef Delim = llvm::StringRef(":"),
+                    llvm::StringRef Delim = llvm::StringRef(platform::kEnvDelim),
                     bool Verbose = false);
 
     ///\brief Adds multiple include paths separated by a delimter into the
@@ -60,7 +66,7 @@ namespace cling {
     ///
     void AddIncludePaths(llvm::StringRef PathStr,
                          clang::HeaderSearchOptions& Opts,
-                         const char* Delim = ":");
+                         const char* Delim = platform::kEnvDelim);
 
     ///\brief Write to llvm::errs that directory does not exist in a format
     /// matching what 'clang -v' would do
--- a/lib/Interpreter/CIFactory.cpp
+++ b/lib/Interpreter/CIFactory.cpp
@@ -7,13 +7,14 @@
 // LICENSE.TXT for details.
 //------------------------------------------------------------------------------
 
-#include "cling/Interpreter/CIFactory.h"
-#include "cling/Utils/Paths.h"
-#include "cling/Interpreter/InvocationOptions.h"
 #include "ClingUtils.h"
-
 #include "DeclCollector.h"
 #include "cling-compiledata.h"
+
+#include "cling/Interpreter/CIFactory.h"
+#include "cling/Interpreter/InvocationOptions.h"
+#include "cling/Utils/Paths.h"
+#include "cling/Utils/Platform.h"
 
 #include "clang/AST/ASTContext.h"
 #include "clang/Basic/TargetInfo.h"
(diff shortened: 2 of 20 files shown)


lib/Interpreter/IncrementalJIT.cpp
```diff
@@ -217,13 +218,19 @@ IncrementalJIT::getSymbolAddressWithoutMangling(llvm::StringRef Name,
   if (auto Sym = getInjectedSymbols(Name))
     return Sym;
 
+  const std::string SymName = Name;
```
vgvassilev: Why did you need an extra variable here?

lib/Utils/PlatformPosix.cpp
```diff
@@ -167,6 +111,73 @@ bool GetSystemLibraryPaths(llvm::SmallVectorImpl<std::string>& Paths) {
   return true;
 }
 
+/*
+### TODO Use OS X IsMemoryValid on all BSD variants:
```
vgvassilev: Could you use FIXME:, we are trying to use only FIXME: in the codebase.

lib/Interpreter/IncrementalJIT.cpp
marsupial: Avoid converting string ref to std::string 3x.

marsupial:
Also looked into this but two big problems: 
 To construct an MSVCToolChain you need a Driver, but we're running this before we have a Driver.
 Even if we constructed a MSVCToolChain, it's going to pull what it thinks is the best from the environment. But cling should favor the environment it was built with over the one it's being run in.

lib/Interpreter/IncrementalJIT.cpp
vgvassilev: What about making `DLSym` taking a `llvm::StringRef`, this would factor out the need of `CStr` variable, IIUC. In general we are very much in favor of using `llvm::StringRef`-s where possible.

vgvassilev:
I think we can steer what the MSVCToolChain pulls in. I am very much in favor of reducing the code duplication, even at the cost of a few round trips and tweaks of the existing ToolChain to do what we want.

marsupial:
I agree it would be good, and I'm also investigating ways of doing this better as its still failing in a lot of ways, but it's beyond the scope of this PR.

vgvassilev:
Indeed and I'd be interested to review future PRs going in this direction!

lib/Interpreter/IncrementalJIT.cpp
marsupial: Everything past that line takes a std::string so it'd still be good to cache the conversion.

CStr variable?
I agree it's a nicer object and thought to use a StringRef, but currently all code using DLOpen/DLClose is a std::string.
It also seems more succiint as the optional err parameter will need to be a std::string and then you've got to eat the include anyway.

marsupial: And the problem was dlopen needs a c-string, and if you use StringRef you're basically constructing an llvm::StringRef which is cheap enough, but then you have to re-construct a new std::string to get the c-string.

vgvassilev: Hm, there is `llvm::StringRef::data()` which gives you the c-string.

marsupial: So you're the one!
It's a length and a pointer no null termination garauntee.
llvm::StringRef("chopmeup").substr(0,4).data() -> "chopmeup"
llvm::StringRef("chopmeup").substr(0,4).str().c_str() -> "chop"

pcanal: `llvm::StringRef::data()` does not guarantee that the content is null terminated and if the StringRef is referring to a sub-part of the original input, data() will not have an null character at the 'end' of the StringRef.   So data() can not be used when a c-string is expected.

vgvassilev: Ah, ok, indeed.

marsupial: Not a bad idea.
Averaging the time to run each test in the suite 6 times before and after showed a 7% drop in execution time. Could be noise but it's definitely more resource friendly this way.

lib/Interpreter/IncrementalJIT.cpp
```diff
@@ -221,6 +222,11 @@ IncrementalJIT::getSymbolAddressWithoutMangling(llvm::StringRef Name,
     if (RuntimeDyld::SymbolInfo SymInfo = m_ExeMM->findSymbol(Name))
       return llvm::orc::JITSymbol(SymInfo.getAddress(),
                                   llvm::JITSymbolFlags::Exported);
+#ifdef LLVM_ON_WIN32
+    if (const void* Sym = platform::DLSym(Name))
```
karies: Why doesn't the process symbol resolution (in `RTDyldMemoryManager::getSymbolAddressInProcess()`) work, why do we need a special Windows code path here?

marsupial: From what I gather it's because on Windows there isn't a similar thing to RTLD_GLOBAL.

karies: Thanks. I'd prefer to wait for the feedback of Bertrand, our local Windows expert. We expect him back on Thursday. In principle I'd prefer to have the dynamic library handling fixed, rather than sneaking this dlsym call in here. The llvm::DynLibraries are so unhelpful that we have considered getting rid of them. Maybe your suggestion is the right way.

Do you have timing info? How long does it take to enumerate all libs? How long does it take to find a symbol per library? We have users with hundreds of shared libs; I'm a bit scared of that loop...

lib/Utils/PlatformMac.cpp
```diff
@@ -32,6 +34,19 @@
 namespace cling {
 namespace utils {
 namespace platform {
+
+bool IsMemoryValid(const void *P) {
+  char Buf;
+  vm_size_t sizeRead = sizeof(Buf);
+  if (::vm_read_overwrite(mach_task_self(), vm_address_t(P), sizeRead,
+                          vm_address_t(&Buf), &sizeRead) != KERN_SUCCESS)
```
karies: Awesome! Way to go, @marsupial !

lib/Interpreter/CIFactory.cpp
```diff
@@ -1191,12 +715,7 @@ namespace {
     // name), clang will call $PWD "." which is terrible if we ever change
     // directories (see ROOT-7114). By asking for $PWD (and not ".") it will
     // be registered as $PWD instead, which is stable even after chdirs.
-    char cwdbuf[2048];
-    if (!getcwd_func(cwdbuf, sizeof(cwdbuf))) {
-      // getcwd can fail, but that shouldn't mean we have to.
-      ::perror("Could not get current working directory");
-    } else
-      FM.getDirectory(cwdbuf);
+    FM.getDirectory(platform::GetCwd());
```
karies: What if `GetCwd()` fails / returns an empty string?

lib/Interpreter/DynamicLibraryManager.cpp
```diff
@@ -10,179 +10,55 @@
...
+      "DYLD_FRAMEWORK_PATH",
+      "DYLD_FALLBACK_FRAMEWORK_PATH",
+      "DYLD_VERSIONED_FRAMEWORK_PATH",
+      */
+  #elif defined(LLVM_ON_WIN32)
+      "PATH",
+  #endif
+    };
+
+    // Behaviour is to not add path's that don't exists...In an interpreted
```
karies: spelling `path's` -> `paths`; `exists` -> `exist`

lib/Interpreter/IncrementalJIT.cpp
vgvassilev: @karies what about a FIXME here and a note to Bertrand? I'd like to see this PR in soon.

lib/Interpreter/DynamicLibraryManager.cpp
```diff
@@ -10,179 +10,55 @@
...
-#endif
-
-}
-
-namespace cling {
-  DynamicLibraryManager::DynamicLibraryManager(const InvocationOptions& Opts)
-    : m_Opts(Opts), m_Callbacks(0) {
-    GetSystemLibraryPaths(m_SystemSearchPaths);
+    // This will currently be the last path searched, should it be pushed to
+    // the front of the line, or even to the front of user paths?
```
karies: I agree - this should end up very early, before $LD_LIBRARY_PATH and system oaths. I don't care about the order wrt user provided ones.

lib/Interpreter/IncrementalJIT.cpp
vgvassilev: Cool, thanks!

lib/Utils/Paths.cpp
```diff
@@ -16,6 +16,16 @@
 namespace cling {
 namespace utils {
 
+namespace platform {
+#if defined(LLVM_ON_UNIX)
+  const char* const kEnvDelim = ":";
+#elif defined(LLVM_ON_WIN32)
+  const char* const kEnvDelim = ";";
+#else
+  #error "Unknown platform (environmental delimter)"
```
karies: spelling `delimter` -> `delimiter`.

lib/Utils/Paths.cpp
```diff
@@ -121,26 +131,34 @@ void LogNonExistantDirectory(llvm::StringRef Path) {
 bool SplitPaths(llvm::StringRef PathStr,
                 llvm::SmallVectorImpl<llvm::StringRef>& Paths,
                 SplitMode Mode, llvm::StringRef Delim, bool Verbose) {
-  bool AllExisted = true;
-#ifdef _MSC_VER
+  assert(Delim.size() && "Spliting without a delimiter");
```
karies: split*t*ing

lib/Utils/Paths.cpp
```diff
@@ -121,26 +131,34 @@ void LogNonExistantDirectory(llvm::StringRef Path) {
...
        !Split.second.empty(); Split = PathStr.split(Delim)) {
 
     if (!Split.first.empty()) {
       bool Exists = llvm::sys::fs::is_directory(Split.first);
 
-  #ifdef _MSC_VER
-    // TODO: Should this all go away and have user handle platform differences?
-    // Right now there are issues with CMake updating cling-compiledata.h
-    // and it was previoulsy generating 'C:\User\Path:G:\Another\Path'
+  #if defined(LLVM_ON_WIN32)
```
karies: Can the `#` move to column 1?

lib/Utils/PlatformMac.cpp
```diff
@@ -0,0 +1,214 @@
+//--------------------------------------------------------------------*- C++ -*-
+// CLING - the C++ LLVM-based InterpreterG :)
+// author:
```
karies: Author

lib/Utils/PlatformMac.cpp
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+
+#define GET_ISYSROOT_VER(maj, min) \
+  if (getISysRootVersion(SDKs, maj, min, sysRoot, Verbose ? \
+                 "match what cling was compiled with" : nullptr)) \
+    return true;
+
+  // Try to get the SDK for whatever cling was compiled with
+  #if defined(MAC_OS_X_VERSION_10_11)
```
karies: `MAC_OS_X_VERSION_MAX_ALLOWED` seems to have be set to the latest `MAC_OS_X_VERSION_...` at compilation time. They have a specific value, e.g. `101100` for `10.11.00` and `1060` for `MAC_OS_X_VERSION_10_6` (to destroy consistency, I guess ;-).

That should allow you to write a `getISysRootVersion` loop that is future safe, i.e. one that will work for 10.12, too! (and likely with less lines of code)

lib/Utils/PlatformMac.cpp
```diff
@@ -0,0 +1,214 @@
...
+    GET_ISYSROOT_VER(10, 0);
+  #endif
+
+#undef GET_ISYSROOT_VER
+
+  // Nothing left to do but iterate the SDKs directory
+  // copy the paths and then sort for the latest
+  std::error_code ec;
+  std::vector<std::string> srtd;
+  for (fs::directory_iterator it(SDKs, ec), e; it != e; it.increment(ec))
```
karies: Shouldn't `e` be initialized?

karies: Shouldn't we check for `ec` in the loop condition?

lib/Utils/PlatformMac.cpp
```diff
@@ -0,0 +1,214 @@
...
+    srtd.push_back(it->path());
+  if (!srtd.empty()) {
+    std::sort(srtd.begin(), srtd.end(), std::greater<std::string>());
```
karies: I don't think you need the last arg here?

lib/Utils/PlatformPosix.cpp
```diff
@@ -0,0 +1,201 @@
...
+namespace utils {
+namespace platform {
+
+std::string GetCwd() {
+  char Buffer[PATH_MAXC];
+  if (::getcwd(Buffer, sizeof(Buffer)))
+    return Buffer;
+
+  ::perror("Could not get current working directory");
+  return std::string();
```
karies: Generally, up to you, and for all other occurrences: you can also spell this as `return {}` which makes it a bit more obvious that this is not about converting a string to the return type, but that you just want to return a default cted string.

lib/Utils/PlatformPosix.cpp
```diff
@@ -0,0 +1,201 @@
...
+}
+
+static void DLErr(std::string* Err) {
+  if (!Err)
+    return;
+  if (const char* DyLibError = ::dlerror())
```
karies: This does not pick up `dlerror()` if no string is passed. Maybe:
```
if (const char* DyLibError = ::dlerror())
  if (Err)
    *Err = DyLibError;
```

karies: continuing review from here on next week :-( Sorry!

lib/Interpreter/IncrementalJIT.cpp
marsupial: Sorry no info on timing.  I'm not sure llvm::DynLibraries are being used anymore, and would like to keep it that way.  Both because it can't support the 'permanent' flag or unloading.  I have a subsequent commit that implements the 'permanent' flag as well as the following behavior:

```
Interpreter parent;
{
 Interpreter child(parent);
 // Load dynamic libraries
}
// ~child will unload all libraries it loaded
```
This is important to me to reclaim memory that won't be accessed anymore.

include/cling/Utils/Paths.h
```diff
@@ -46,7 +52,7 @@ namespace cling {
     bool SplitPaths(llvm::StringRef PathStr,
                     llvm::SmallVectorImpl<llvm::StringRef>& Paths,
                     SplitMode Mode = kPruneNonExistant,
-                    llvm::StringRef Delim = llvm::StringRef(":"),
+                    llvm::StringRef Delim = llvm::StringRef(platform::kEnvDelim),
```
marsupial: confused

lib/Interpreter/CIFactory.cpp
marsupial: FileManager::getDirectory can handle that, there's already a test doing it.

lib/Utils/PlatformMac.cpp
marsupial: I believe the std::error_code  constructor defaults to success and that it.increment returns false when ec is a failure code....but I'll have another look.

lib/Utils/PlatformPosix.cpp
marsupial: What is that initializer list construction of a string?

lib/Utils/PlatformMac.cpp
marsupial: The only thing is these macros are user configurable, and it's valid to do:
```
-DMAC_OS_X_VERSION_MIN_REQUIRED=1070
-DMAC_OS_X_VERSION_MAX_ALLOWED=1080
```
And then compile with SDK 10.9. Since we're trying to match the SDK compiled it will fail.
If you're ok with this (admittedly bizarre and contrived) case, then simplification would be good.

lib/Utils/PlatformMac.cpp
marsupial: Hmmm, seems a larger problem of lexically sorting 10.9 vs 10.10 needs to be addressed.

lib/Interpreter/IncrementalJIT.cpp
bellenot: According to [this post](http://www.drdobbs.com/dynamic-plug-ins-for-c/184401819), on Windows, exported symbols are like Global symbols on UNIX. And according to the [LoadLibrary function documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684175(v=vs.85).aspx): 
> LoadLibrary can be used to load a library module into the address space of the process and return a handle that can be used in GetProcAddress to get the address of a DLL function. 
But I'm not sure how this could answer the original question...

karies: I now :-) agree that this is the right spot to add that windows specific code. I'd like to request two things:
- Could you add a FIXME that we need to get rid of the call to m_ExeMM->findSymbol (which in the end calls llvm::sys::DynamicLibrary::SearchForAddressOfSymbol), and replace it with our own
- In that "our own", we should revisit whether there is a way to create a hash map of all symbols, (or use / query the one from the Windows kernel for that process) instead of iterating through all opened libraries

One more note: this windows specific part should basically (or really?) replace the call to m_ExeMM->findSymbol(). But we have another call to that a few lines down. Should this get the same Windows-specific treatment?

lib/Utils/PlatformMac.cpp
karies: But we can just `#undef` those before including the apple version magic. Would that work? Can you do it?

lib/Utils/PlatformMac.cpp
karies: llvm/src/lib/Support/CachePruning.cpp has
```
  for (sys::fs::directory_iterator File(CachePathNative, EC), FileEnd;
       File != FileEnd && !EC; File.increment(EC)) {
```
I.e. we're only missing the `&& !ec` part in your code.

lib/Utils/PlatformMac.cpp
karies: Then you'll indeed need your own predicate, sorting first by length then lexically.

lib/Utils/PlatformPosix.cpp
karies: Huh? It compiles down to the same bytes (I hope!) - I was just mentioning that I generally prefer `return {};` to signal that this is innocuous :-)

lib/Utils/PlatformPosix.cpp
karies: I disagree - calling code (way up the callstack) will *sometimes* have it, *sometimes* not, depending on the call path in between. So I'd prefer a simple "DLErr will reset it and hand it over if you want".

karies:
Okay - progress: I think we're now all set except for the small-ish mods in the comments. Thanks for bearing with us! Do you think you'll have a chance to address these?

lib/Utils/PlatformPosix.cpp
marsupial: Largely I don't really care, but my point was this is now user facing code, so it can be used in the Interpreter:
[cling]: void* lib = DLOpen();
[cling]: more commands, and lib is discovered to be null.
[cling]: dlerror() // show the error that was generated.

lib/Utils/PlatformMac.cpp
marsupial: https://github.com/root-mirror/cling/pull/53/commits/24540d84ac17b570bd6a4895662df4f0d93884a3 addressed the problem.

lib/Utils/PlatformMac.cpp
marsupial: Yes, https://github.com/root-mirror/cling/pull/53/commits/24540d84ac17b570bd6a4895662df4f0d93884a3 dressed this too.

lib/Utils/PlatformMac.cpp
marsupial: Sure.

lib/Interpreter/IncrementalJIT.cpp
marsupial: When I talked about RTLD_GLOBAL it was in the context of symbol resolution, and that dlopen(0, RTLD_GLOBAL) provides a -special- handle that will search all loaded images for symbols. As far as I know Windows doesn't have such a mechanism.

m_ExeMM->findSymbol calls RTDyldMemoryManager::getSymbolAddressInProcess which does a lot of other lookup. I'm not sure erasing that call on Windows is the best thing to do (at least right now)

lib/Utils/PlatformPosix.cpp
marsupial: Yes, but you're example is trying to compensate for a possible **misuse** of the dlopen API, where mine is demonstrating a _reasonable expectation_ that cling hasn't changed the behavior of an existing API.

karies: Fair enough, you convinced me. Thanks!

marsupial: And there is also the issue that doing this means any error should be eaten on Windows as well which is a lot more complicated.

marsupial:
I think I've addressed everything other than what was in reference to the comment in DynamicLibraryManager.cpp.  There are some other issues with DynamicLibraryManager.cpp and that change is behavioral which this PR was _mostly_ trying to avoid.

lib/Utils/PlatformWin.cpp
```diff
@@ -0,0 +1,687 @@
...
+      ::RegCloseKey(hKey);
+    } else
+      ReportError(lResult, "RegOpenKeyEx");
+  }
+  return returnValue;
+}
+
+int GetVisualStudioVersionCompiledWith() {
+#if (_MSC_VER >= 1900)
+  return 14;
```
bellenot: Why not:
`   return (_MSC_VER / 100) - 5;`
(instead of `   return 14`)?

marsupial: In either case it will still be broken for the forthcoming Visual Studio "15". And I don't want to guess what will happen. (https://en.wikipedia.org/wiki/Microsoft_Visual_Studio)

Originally there was no discrimination and it was _**(_MSC_VER / 100) - 6**_ in some places.

lib/Interpreter/IncrementalParser.cpp
```diff
@@ -48,116 +48,61 @@
...
-    // For MSVC we do not use CLING_CXXABI*
-#else
-# define CLING_CXXABIV -1 // intentionally invalid macro name
-# define CLING_CXXABIS "-1" // intentionally invalid macro name
-    llvm::errs()
-      << "Warning in cling::IncrementalParser::CheckABICompatibility():\n  "
-      "C++ ABI check not implemented for this standard library\n";
-    return;
-#endif
-#ifdef _MSC_VER
```
bellenot: P.S. I was referring to this original version check...

lib/Utils/PlatformWin.cpp
marsupial: That doesn't work with patches to Visual Studio 2015.

vgvassilev: Maybe that helps: https://blogs.msdn.microsoft.com/vcblog/2016/10/05/visual-c-compiler-version/

karies:
👍

lib/Utils/PlatformWin.cpp
marsupial: But according to the MSDN link, LLVM is doing it wrong now. According to the link it should be:
```
#if (_MSC_VER >= 1910)
  return 15;
#elif (_MSC_VER >= 1900)
  return 14;
#endif
  return (_MSC_VER / 100) - 6;
```

Is this what is wanted?

marsupial: Keeping in mind that this code is predominately used for registry keys, of which I have no idea the success with what VisualStudio "15" does.

karies: Can we please just error out instead of guessing possibly the wrong install? Debugging a failure here (i.e. a case where we pick up the wrong reg keys of a side-by-side install of different VC versions) will be a nightmare. I'd prefer to error out for new versions and then add them when they arrive. New VC happens one a year, we can cope with that.

karies: Sorry @marsupial, we are not at the same physical location at the moment, so we use github to discuss this publicly.

I am questioning that request to adapt for the unknown future. I prefer to have an `#error` that tells us "go fix" instead of guessing, possibly guessing wrong, and then not being able to detect that and having to spend a long time debugging registry path issues. I.e.
```
#if _MSC_VER >= 1910
# error "MSVC version not yet supported."
#endif
```

pcanal: The MSDN suite claims:

> Each major release of the Visual C++ compiler increments the “hundreds” of _MSC_VER. Each update within a major release increments the “units” by 1. For example, in VS “15” Preview 5, the macro _MSC_VER evaluates to 1910. The next update will have _MSC_VER set to 1911.
> 
> Note that VS “15” and VS2015 are both major releases of Visual Studio with different major version numbers. The included compiler toolset, however, will have the same major version number – with the change described here, the minor version number can be used to distinguish compiler toolsets.

From which I am concluding that we should *not* care about the VisualStudio number at all but only about the 'compiler toolset' version number ...  In part because the link between the two version is going to be completely arbitrary ...

But then again, I have to ask :) how are we using this function GetVisualStudioVersionCompiledWith?

lib/Interpreter/IncrementalJIT.cpp
```diff
@@ -212,6 +212,22 @@ IncrementalJIT::getInjectedSymbols(llvm::StringRef Name) const {
   return JITSymbol(nullptr);
 }
 
+llvm::PointerIntPair<void*, 1>
+IncrementalJIT::searchLibraries(llvm::StringRef Name, void *InAddr) {
+  // FIXME See comments on DLSym below
```
vgvassilev: Sorry for being pedantic: missing colon and a full stop in the end.

lib/Interpreter/IncrementalJIT.cpp
```diff
@@ -212,6 +212,22 @@ IncrementalJIT::getInjectedSymbols(llvm::StringRef Name) const {
   return JITSymbol(nullptr);
 }
 
+llvm::PointerIntPair<void*, 1>
```
vgvassilev: Can't we use llvm::Optional here?

lib/Interpreter/IncrementalParser.cpp
```diff
@@ -72,50 +71,37 @@ namespace {
...
     llvm::StringRef CurABI;
     const clang::Preprocessor& PP = CI->getPreprocessor();
-    const clang::Token* Tok = getMacroToken(PP, CLING_CXXABIS);
+    const clang::Token* Tok = getMacroToken(PP, CLING_CXXABI_NAME);
     if (Tok && Tok->isLiteral()) {
-      CurABI = llvm::StringRef(Tok->getLiteralData(), Tok->getLength());
-      std::ostringstream AbiVers;
-      AbiVers << CLING_CXXABIV;
-      if (CurABI.equals(AbiVers.str()))
+      // Strip any quotation marks
```
vgvassilev: Same here.

lib/Interpreter/IncrementalJIT.cpp
marsupial: It's not optional, Address can be non null, and still not cached. 
A _slightly_ more efficient std::pair\<void*,bool\>.

vgvassilev:
Merged, thanks for your contribution!

PR #54, opened 2016-09-19
Peek inside extern "C" declarations only after name hasn't been matched.

marsupial:


vgvassilev:
Thanks for the PR. In the last PR I was going after making this work, too:
```cpp
extern "C" int printf(const char*, ...);

template<typename T>
struct number {
  number(int) {}
  friend int f(number x) { printf("I got called!!!\n"); return 0; };
};

number<double> a(3);
//int error = f(1); // ADL doesn't kick in.
//int ok = f(a); // ADL adds this to the global lookup scope.
```

```bash
$ cling  ".x f.C(a)" # should work and print "I got called!!!".
warning: cannot find function 'f()'; falling back to .L
```

This requires a proper sema lookup.

karies:
Let me understand the goal here. This function is only used to find the function to run for a ".L foo.C". As such it should only find `::foo()`.

I think this function should not be in Transaction; it should move to MetaProcessor/ and be renamed to `findFunctionToExecute(name)` or similar. Then we can tweak its implementation to do exactly what we need.

karies:
@vgvassilev - I disagree that this should work. The mental model is that for a ".x foo.C" there must be a function on global scope that's called `foo` and that takes no arguments. Generic lookup is not the mental model that fits here.

pcanal:
@vgvassilev You example is misformed: "A name first declared in a friend declaration within class or class template X becomes a member of the innermost enclosing namespace of X, but is not accessible for lookup (except argument-dependent lookup that considers X) unless a matching declaration at the namespace scope is provided" http://en.cppreference.com/w/cpp/language/friend.

pcanal:
@karies @vgvassilev More genuine is, should we support:
```
// foo.C
namespace Inner { void f() {}; }
using namespace Inner;
```
```
.x foo.C
```
Cheers,
Philippe.
In CINT, 
```
class Foo {};
```
```
.x Foo.C
```
would call the Foo constructor .... if I remember correctly.

vgvassilev:
@karies ok, if you we decided to not support the given case, how would we 'ban' this one:
```cpp
extern "C" int printf(const char*, ...);

template<typename T>
struct number {
  number(int) {}
  friend int f(number x = 3) { printf("I got called!!!\n"); return 0; };
};

number<double> a(3);
//int error = f(1); // ADL doesn't kick in.
int ok = f(a); // ADL adds this to the global lookup scope.
```
```bash
cling  ".x f.C(a)"
I got called!!!
I got called!!!
(int) 0
```

@pcanal: I didn't invent this code I took it from here http://en.cppreference.com/w/cpp/language/adl

marsupial:
While the discussion is interesting, the PR was related to reverting to previous behavior before delving into contexts. I've pushed another commit onto this, as the only caller of the code was failing for non-posix paths.

As for the discussion:
Why not just build the expression/call and do it? Clang is already handling the resolution.
Seems to me supporting/parsing for ADL would really complicate things.
The current behavior to restrict it to the last Transaction can be beaten in many ways.

[cling] int DotX() { return 10; }
[cling] .x DotX() // file contents: int DotX();

[cling] int DotX(int a, int b) { return 20; }
[cling] .x DotX(20,10) // file contents: int DotX() { return 10; }

lib/Interpreter/Transaction.cpp
```diff
@@ -68,7 +68,12 @@ namespace cling {
           if (name.equals(ND->getNameAsString()))
             return ND;
         }
-        else if (LinkageSpecDecl* LSD = dyn_cast<LinkageSpecDecl>(DI)) {
+      }
+    }
+    // Not found yet, peek inside extern "C" decalrations
```
karies: Spelling `decalrations`

lib/Interpreter/Transaction.cpp
```diff
@@ -68,7 +68,12 @@ namespace cling {
...
+    for (auto I = decls_begin(), E = decls_end(); I < E; ++I) {
```
karies: `I != E`, don't rely on iterator ordering

pcanal: @vgvassilev Even after the first ADL aided find, the function is *still* not really in the global namespace:
```
root [1] f(a)
I got called!!!
(int) 0
root [2] f(3)
input_line_15:2:3: error: use of undeclared identifier 'f'
 (f(3))
  ^
```

marsupial: Yes, and no....
```
[cling]$ f(4)
input_line_6:2:2: error: use of undeclared identifier 'f'
 f(4)
 ^
[cling]$ namespace f {}
input_line_7:1:11: error: redefinition of 'f' as different kind of symbol
namespace f {}
          ^
```
If you're specifically trying to limit per-transaction, you'll have to give up some corner cases that won't work or a lot of convolution matching the spec.
And you'll have to match clang's interpretation of the spec as clang will be the ultimate decider, so why not just let it do its thing?

karies:
Hi everyone,

This PR is fine with me. We can do what we discussed outside this PR. You can then raise your concerns about what that actual implementation does (instead of an imaginary one) on that separate PR.

Deal? Does anyone have any objections against this PR?

Cheers, Axel.

marsupial:
Just to be clear I changed the original loop to != as well.
In the comment, but don't want it missed in the noise.

karies:
Thanks for fixing our bugs :-) LGTM!

marsupial:
You fixed it, I'm just a typing monkey.

karies:
LOL :-) Can you provide a test for this?

marsupial:
The original commit had a test for extern "C".

marsupial:
Still don't know what test you're going after.

karies:
Merged, thanks! (Sorry Vassil, I again overlooked that you self-assigned.)

PR #55, opened 2016-09-20
Does not ignore IPKernelApp.init_io

pedroteixeira:
In latest jupyter versions this fixes AttributeError 'ClingKernelApp object has no attribute displayhook'


vgvassilev:
Merged, thanks for your contribution.

PR #57, opened 2016-10-04
3 Bugs 2 Optimizations

marsupial:


lib/UserInterface/textinput/StreamReaderWin.cpp
```diff
@@ -176,9 +176,20 @@ namespace textinput {
         return false;
       }
     } else {
+      //https://msdn.microsoft.com/en-us/library/windows/desktop/aa365690(v=vs.85).aspx
```
vgvassilev: Could you explain briefly what is the intent here, instead of pasting only a link.

test/SourceCall/unresolved.C
```diff
@@ -6,7 +6,6 @@
 // LICENSE.TXT for details.
 //------------------------------------------------------------------------------
 
-// RUN: cat %s | %cling
```
vgvassilev: In the past we used to have issues when cling crashed, FileCheck could still return success. This is a check against cling crashing.

lib/UserInterface/textinput/StreamReaderWin.cpp
marsupial: The intent is to stop when EOF reached.

vgvassilev: Is this a change, relevant only for windows? Please add two lines explaining in plain English why we need to do that, instead of a link, which might change in a few years from now.

vgvassilev:
Merged, thanks for your contribution!

PR #58, opened 2016-10-06
Windows: Fix EOF not begin handled when reading from input.

marsupial:
Windows isn't handling file input correctly, particularly for piped input.


vgvassilev:
I checked that in in the ROOT repository. The update of the cling version of textinput is manual and @karies  kindly agreed to do it.

karies:
Synced with cling.

PR #59, opened 2016-10-06
Value Printer cleanup / fixes.

marsupial:
There's a fair amount of incongruity with how values are printed in Windows and Unix.
These commits cleanup ValuePrinter.cpp and begin to address those differences.

### DIFF
--- a/include/cling/Interpreter/RuntimePrintValue.h
+++ b/include/cling/Interpreter/RuntimePrintValue.h
@@ -103,10 +103,10 @@ namespace cling {
         return valuePrinterInternal::kEmptyCollection;
 
       std::string str("{ ");
-      str += printPair(*iter);
+      str += printPair(iter);
       while (++iter != iterEnd) {
         str += ", ";
-        str += printPair(*iter);
+        str += printPair(iter);
       }
       return str + " }";
     }
--- a/test/Prompt/ValuePrinter/Collections.C
+++ b/test/Prompt/ValuePrinter/Collections.C
@@ -12,6 +12,7 @@
 #include <string>
 #include <tuple>
 #include <vector>
+#include <map>
 
 std::vector<bool> Bv(5,5)
 // FIXME: Printing std::vector<bool> is still broken.
@@ -33,6 +34,15 @@ cling::printValue(&A) == cling::printValue(&B)
 
 std::tuple<> tA
 // CHECK: (std::tuple<> &) {}
+
+std::map<int, int> m
+// CHECK: (std::map<int, int> &) {}
+
+for (int i = 0; i < 5; ++i) m[i] = i+1;
+
+m
+// CHECK: (std::map<int, int> &) { 0 => 1, 1 => 2, 2 => 3, 3 => 4, 4 => 5 }
+
 
 // expected-no-diagnostics
 .q


lib/Interpreter/ValuePrinter.cpp
```diff
@@ -151,52 +146,59 @@ bool canParseTypeName(cling::Interpreter& Interp, llvm::StringRef typenam) {
...
-  printValueSS << "cling::printValue(";
-  printValueSS << getTypeString(V);
-  printValueSS << (const void *) &val;
-  printValueSS << ");";
-
   Interpreter *Interp = V.getInterpreter();
   Value printValueV;
 
   {
+    // don't use std::stringstream, since it doesn't prepend '0x'
```
vgvassilev: Could you start with comments capital letters and add full stop in the end?

lib/Interpreter/ValuePrinter.cpp
```diff
@@ -151,52 +146,59 @@ bool canParseTypeName(cling::Interpreter& Interp, llvm::StringRef typenam) {
...
+    // in front of hexadecimal values when streaming pointer values
+
+    llvm::SmallString<512> Buf;
+    llvm::raw_svector_ostream Strm(Buf);
+    Strm << "cling::printValue(";
+    Strm << getTypeString(V);
+    Strm << (const void*) &val;
+    Strm << ");";
```
vgvassilev: Wouldn't a call to `enclose` reduce the line count here?

lib/Interpreter/ValuePrinter.cpp
```diff
@@ -398,71 +385,59 @@ static std::string printFunctionValue(const Value &V, const void *ptr, clang::Qu
...
 
-static std::string printUnpackedClingValue(const Value &V) {
-  std::stringstream strm;
+static std::string printAddress(const void* Ptr, const char Prfx = 0) {
+  if (!Ptr)
+    return kNullPtrStr;
+
+  llvm::SmallString<256> Buf;
+  llvm::raw_svector_ostream Strm(Buf);
+  if (Prfx) Strm << Prfx;
```
vgvassilev: I'd prefer to break the if statements. Could you more the then statement on a new line. Makes it easier for debugging with gdb.

test/Prompt/DontWrap.C
```diff
@@ -172,7 +172,7 @@ const Test2::Nested::Value::type & Test2::Nested::reference() const {
 Test2::Nested Nest2(80);
 // CHECK: Nested::Nested(160)
 Nest2.simpleAdd(3)
-// CHECK: (Test2::Nested::Value::type) 83
+// CHECK: (int) 83
```
vgvassilev: Hm... we shouldn't have desugared this typedef here.

test/Prompt/ValuePrinter/Collections.C
```diff
@@ -0,0 +1,37 @@
+//------------------------------------------------------------------------------
+// CLING - the C++ LLVM-based InterpreterG :)
+// author:  Danilo Piparo <danilo.piparo@cern.ch>
```
vgvassilev: Hm. Please either add your name or leave this empty.

lib/Interpreter/ValuePrinter.cpp
```diff
@@ -238,31 +238,25 @@ static std::string invokePrintValueOverload(const Value &V) {
...
+  if (Ty->isFunctionType())
+    return executePrintValue<const void*>(V, &V);
+
+  if (Ty->isPointerType() || Ty->isReferenceType() || Ty->isArrayType())
+    return executePrintValue<void*>(V, V.getPtr());
+
+  if (Ty->isObjCObjectPointerType())
+    return executePrintValue<void*>(V, V.getPtr());
+
+  // struct/fallback case
```
vgvassilev: I'd prefer comments in the form of clear sentences, starting with capital letter and finishing with a punctuation mark.

test/Prompt/DontWrap.C
marsupial: Well this is a problem on Windows.
[cling] std::string s("T")[0]
(ValueWrapper<char>::value_type&) 'T'

'ValueWrapper' may not be the name but it really obfuscates a lot of types that way.
Is there a reason to not use the canonical type?
To me it's a lot more legible when printing the canon type, and can easily imagine templated structs that grow the line phenomenally.

lib/Interpreter/ValuePrinter.cpp
marsupial: That is not my comment.

lib/Interpreter/ValuePrinter.cpp
marsupial: It could drop the line count by 1, but as there is already a local stream would be less efficient.

test/Prompt/DontWrap.C
karies: I prefer `std::vector<int>::iterator` over the implementation detail.

But I prefer `int` over `std::vector<int>::value_type`.

So we cannot prefer one over the other. My suggestion: use the shortest name.

include/cling/Interpreter/RuntimePrintValue.h
```diff
@@ -15,6 +15,7 @@
 #endif
 
 #include <string>
+#include <vector>
```
karies: Can we do the vector<bool> case by SFINAEing on whether we can do &v[0] or not? Then we don't need any `#include <vector>`.

marsupial: It seemed the previous code was attempting to do this, but there were two problems:
1. It wouldn't compile on Windows due to ambiguity.
2. On Unix it would compile, but the wrong specialization would be used, leading to a crash.

I don't know if MS of clang was correct in respect to the standard.

So at this point the include is a necessity to print std::vector\<bool\>. As vector can be hidden inside an inline namespace forward declaration will be ugly.

The best I can offer in regards to speed/memory is that I have a commit that doesn't even bother RuntimePrintValue.h for builtin types.

karies: Right, indeed, `printValue_impl` was supposed to do that and it fails: the second "overload" is meant to be the case where `&v[0]` doesn't work. But indeed, that code is invalid because both "overloads" have the same signature, and the second doesn't get disabled if the first is valid.

We should "just" fix that. Do you want to try?

test/Prompt/DontWrap.C
marsupial: De-sugaring only when a typedef reduces to a builtin type, prior behavior otherwise?

include/cling/Interpreter/RuntimePrintValue.h
vgvassilev: @marsupial what ambiguity on Windows do you mean? This code is only used at cling's runtime, i.e. this will be only compiled by cling.

marsupial: Compilation at runtime yes.
There is ambiguity in the attempt to differentiate containers who's elements cannot have their address taken. See the other part of the commit for what was removed/replaced.
As far as I know STL only allows this behavior for std::vector<bool>.

I'll add a FIXME, but I don't have the time to delve into the arcane semantics of **decltype** to fix it in a more efficient way (especially not knowing if that is even possible).

While I appreciate the inclusion of vector is sub-optimal in a performance sense, not including it and crashing is way worse.

test/Prompt/DontWrap.C
marsupial: [cling] std::vector<int> a(5,5);
[cling] a.begin()
[cling] a\[0\]
A. (std::__1::vector\<int, std::__1::allocator\<int\> \>::iterator) @0x7fd76077e480
a. (std::__1::__vector_base\<int, std::__1::allocator\<int\> \>::value_type) 5
B. (std::__1::__wrap_iter\<int *\>) @0x7f8340653d30
b. (int) 5
C. (std::vector\<int, std::allocator\<int\> \>::iterator) @0x7fad4cf7bbf0
c. (int) 5

So what should happen?
A is behavior before the PR,
B is behavior at PR
C is updating the PR to reflect changes discussed

include/cling/Interpreter/RuntimePrintValue.h
karies: You have suggested one fix, I'd hope for a different one :-) Can you tell me what the error looks like? I can try to provide a separate fix for that.

marsupial: The error on Windows is along the lines of 'disambiguation between x and y not possible'
On Unix there is no error, just a crash trying to get the address when print a std::vector\<bool\>

I believe Windows will also crash because the include failure (a separate issue).

test/Prompt/DontWrap.C
pcanal: I agree that of those choice C. is the better option (even-though it does lose some semantic information compare to A ... which is too verbose.

For the purpose of the PR, I would use C.

For completeness, we could also consider (in the future):

D. (std::vector<int>::iterator) @0x7fad4cf7bbf0
d. (std::vector<int>::value_type) 5
E. (std::vector<int>::iterator) @0x7fad4cf7bbf0
e. (int) 5

where we either somehow find the 'type as spelt' or use name normalization similar to the one use in ROOT (dropping inline namespace and defaulted template parameter).

include/cling/Interpreter/RuntimePrintValue.h
karies: I get
```
[cling]$ #include <vector>
[cling]$ std::vector<bool> vb{true, false, true}
(std::vector<bool> &) { true, false, true }
```
on Linux. Do you have the actual error message from Windows?

karies: Never mind - I was able to reproduce this error on Windows; it should be fixed in the master. (cling repo sync will arrive in a minute.) I.e. I *believe* this part (and the `#include <vector>`) can be removed from the PR.

marsupial: The -Unix- crash was OS X with libc++. I'll try your fix when it lands.

test/Prompt/DontWrap.C
marsupial: Side-effect of behavior C.
class T{};
typedef T T2;
T2 t
(T &) @0x10233d000

pcanal: I don't understand why :)

lib/Interpreter/ValuePrinter.cpp
vgvassilev: Ok, could you nevertheless fix them when you are moving them?

lib/Interpreter/ValuePrinter.cpp
vgvassilev: Let's opt for readability here.

test/Prompt/DontWrap.C
marsupial: Its the way the type is deduced when printing.
template <class T> class vector {
  typedef T value_type;
};
typedef std::vector<int> ivec;

The only difference between the two types (semantically?) is one has a parent in a class, the other not. I've tried to get around it if you check that parent of a typedef is TranslationUnit, Namespace, LinkageSpec: print the typedef type; otherwise print the underlying type.
But then (on libc++ at least) since std::vector<int>::iterator  is std::_1::__wrap_iter<int*> it's parent is a namespace, and prints less good.

pcanal: In the ROOT code we have a function ROOT::TMetaUtils::GetUnderlyingType which giving a type returns the end node (aka the 'raw' type, aka the underlying type).   For example for "size_t**" it return "long" (on platform where size_t is a long).

I wonder if we can import this routine and then test whether the underlying type is a clang::BuiltinType.

marsupial: Testing for a builtin type is easy and is already being done (in the PR).
Whats complicated is delineating between typedefs that are better desugared and those better left alone.

From this thread I have another commit that does this:
\[cling\]$ std::vector\<int\> b(1)
(std::vector\<int\> &) { 0 }
\[cling\]$ b\[0\]
(int) 0
\[cling\]$ b.begin()
(std::vector\<int, std::allocator\<int\> \>::iterator) @0x7fad5d062910
\[cling\]$ 
\[cling\]$ class Test {};
\[cling\]$ typedef Test Test2;
\[cling\]$ Test2 t
(Test2 &) @0x105f42ffc
\[cling\]$ 
\[cling\]$ std::vector\<Test\> tb(1)
(std::vector\<Test\> &) { @0x7fad5d1031b0 }
\[cling\]$ tb\[0\]
(Test &) @0x7fad5d1031b0
\[cling\]$ tb.begin()
(std::vector\<Test, std::allocator\<Test\> \>::iterator) @0x7fad5a8e7440
\[cling\]$ 
\[cling\]$ std::vector\<Test2\> tb2(1)
(std::vector\<Test2\> &) { @0x7fad5a912860 }
\[cling\]$ tb2\[0\]
(Test &) @0x7fad5a912860
\[cling\]$ tb2.begin()
(std::vector\<Test, std::allocator\<Test\> \>::iterator) @0x7fad5aa0ce80
\[cling\]$ 
\[cling\]$ typedef std::vector\<Test2\> TVec;
\[cling\]$ TVec tv
(TVec &) {}

Where current behavior is this:
\[cling\]$ std::vector\<int\> b(1)
(std::vector\<int\> &) @0x106eb3488
\[cling\]$ b\[0\]
(std::__1::__vector_base\<int, std::__1::allocator\<int\> \>::value_type) 0
\[cling\]$ b.begin()
(std::__1::vector\<int, std::__1::allocator\<int\> \>::iterator) @0x7ff61d777ac0
\[cling\]$ 
\[cling\]$ class Test {};
\[cling\]$ typedef Test Test2;
\[cling\]$ Test2 t
(Test2 &) @0x106eb3ffc
\[cling\]$ 
\[cling\]$ std::vector\<Test\> tb(1)
(std::vector\<Test\> &) @0x106eb5688
\[cling\]$ tb\[0\]
(std::__1::__vector_base\<Test, std::__1::allocator\<Test\> \>::value_type &) @0x7ff61f044c70
\[cling\]$ tb.begin()
(std::__1::vector\<Test, std::__1::allocator\<Test\> \>::iterator) @0x7ff61f045e90
\[cling\]$ 
\[cling\]$ std::vector\<Test2\> tb2(1)
(std::vector\<Test2\> &) @0x106eb5ba0
\[cling\]$ tb2\[0\]
(std::__1::__vector_base\<Test, std::__1::allocator\<Test\> \>::value_type &) @0x7ff61ad88b10
\[cling\]$ tb2.begin()
(std::__1::vector\<Test, std::__1::allocator\<Test\> \>::iterator) @0x7ff61f153680
\[cling\]$ 
\[cling\]$ typedef std::vector\<Test2\> TVec;
\[cling\]$ TVec tv
(TVec &) @0x108732fc8

include/cling/Interpreter/RuntimePrintValue.h
marsupial: And I already tried it with **_long_** as well...no go.

marsupial: Are the 2 commits related to this going to be backed out?

test/Prompt/DontWrap.C
pcanal: > Whats complicated is delineating between typedefs that are better desugared and those better left alone.

Yes and the 'underlying type' I was mentioning is a gateway to this.
In particular the output for

typedef int* intPtr_t;
intPtr_t val

should probably be 

(int *) @)0x...

rather than

(intPtr_t) 0x....

For the case, you mentioned the first list (just below 'From this thread I have another commit that does this:') long close to the better we can expect/do reasonably, I think ...

include/cling/Interpreter/RuntimePrintValue.h
karies: Should be fixed.

marsupial: There is a typo.
Needs to have an additional #include <type_traits>
Still doesn't work.

karies: How does it not work for you? (It won't be able to print the content of a vector<bool>, but that's a missing feature I'm willing to live with...)

marsupial: Because the PR explicitly enabled the ability to print a std::vector\<bool\> ?

test/Prompt/DontWrap.C
karies: That said - the situation after this PR is better than before, so I'd vote for merging (after removal of the vector part). Objections?

vgvassilev: I don't like treating only a limited set of decls in a special way. What about a more consistent printout. Eg. `(sugared_name) aka [desugared_name]`.

include/cling/Interpreter/RuntimePrintValue.h
vgvassilev: I'd be very much in favor of removing the #include <string> by allowing the caller to provide the storage. Including string brought significant % of STL, so including vector might be cheap because of that.

@marsupial did you check what is the runtime cost in terms of memory and performance of adding these #include.

vgvassilev:
@marsupial, did you run roottest with these changes?

marsupial:
No i haven't run roottest.
CMake fails at configuration and I don't have the time.

test/Prompt/DontWrap.C
marsupial: But the world isn't perfect and a lot of this is subjective...to me reduction of builtin types is better, though I do see that it is not the best 100% of the time. Without it many tests for Windows will have to be re-written, and what you are suggesting increases that to many more.

I think this thread got it to a point that is reasonably nicer than it was before, and the PR as a whole simplifies some spaghetti to make moving towards the _best_ solution easier.

include/cling/Interpreter/RuntimePrintValue.h
vgvassilev: What about memory, we are much more sensitive to memory footprint nowadays.

test/Prompt/ValuePrinter/Collections.C
```diff
@@ -0,0 +1,38 @@
...
+// RUN: cat %s | %cling -Xclang -verify 2>&1 | FileCheck %s
+
+#include <string>
+#include <tuple>
+#include <vector>
+
+std::vector<bool> Bv(5,5)
+// FIXME: Printing std::vector<bool> is still broken.
+// But the line above at least tests cling doesn't crash because of it.
+// BROKENCHECK: (std::vector<bool> &) { true, true, true, true, true }
```
vgvassilev: That should be fixed too. It was fixed in the meanwhile.

karies: No, it currently prints no `true` or `false` but "unrepresentable value". But indeed, at least it doesn't crash anymore. If we want proper value printing of `vector<bool>` (`vector` of everything else works) then we'll need to `#include <vector>` in RuntimePrintValue.h.

marsupial: No std::vector\<bool\> does not print.
The PR added that ability.
That was overruled and asked to be taken out.

vgvassilev: Well, then at least the fixme should be updated. Let's move the `vector<bool>` part in a separate PR.

test/Prompt/DontWrap.C
pcanal: > Eg., if we have:
> [cling]$ tb2.begin()
> ((std::__1::vector<Test, std::__1::allocator<Test> >::iterator)) aka [std::vector<Test, . std::allocator<Test> >::iterator] @0x7fad5aa0ce80
>
> would be fine to write the tests on Windows because FileCheck is smart and it will scan > only for the aka part.

I strongly disagree with @vgvassilev.  Having nice name whenever possible on a prompt for human consumption is *much* better.  I think the behavior from @marsupial displayed above following "From this thread I have another commit that does this:" is good/desirable.

*If* the user need to know the gorry detail, let's just have a '.desugar typedefname' command.  thanks.

pcanal:
I think at this point the PR should be merged as is and any improvement (if any) should be made in a new PR.

marsupial:
Here roottest doesn't pass for me without the changes, but the changes add no new failures.

vgvassilev:
Sorry for the delay. Integrated. Thanks for your contribution.

vgvassilev:
@marsupial it seems we lost the ability to print things like:
``` cpp
  #include <map>
  std::map<int,int> m
```

marsupial:
Fixed..

vgvassilev:
Merged! Thanks!

PR #60, opened 2016-10-18
Use std::pair instead of llvm::PointerIntPair.

marsupial:
Fixes issues where symbols are not 16 bit aligned.

### DIFF
--- a/lib/Interpreter/IncrementalExecutor.cpp
+++ b/lib/Interpreter/IncrementalExecutor.cpp
@@ -329,13 +329,13 @@ IncrementalExecutor::installLazyFunctionCreator(LazyFunctionCreatorFunc_t fp)
 
 bool
 IncrementalExecutor::addSymbol(const char* symbolName,  void* symbolAddress) {
-  return IncrementalJIT::searchLibraries(symbolName, symbolAddress).getInt();
+  return IncrementalJIT::searchLibraries(symbolName, symbolAddress).second;
 }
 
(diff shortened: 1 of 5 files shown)


vgvassilev:
I'd appreciate a test case here.

marsupial:
As I said in the email, it will be very difficult to force a symbol to be exported not aligned to 16 bit...Not only for one platform, but reliably for 3, that would take days to make.

The problem is llvm::PointerInPair relies on alignment, that problem does not exists with std::pair.

vgvassilev:
I think adjusting this should be sufficient.
```c++
#define CHECK_SIZE(name, size)                          \
  extern int name##_1[sizeof(name) == size ? 1 : -1];

struct  __attribute__((packed)) {
  int a;
  int b : 4;
  int c : 32;
} s0;
CHECK_SIZE(s0,9)

#pragma pack (1)
struct {
  int a;
  int b : 4;
  int c : 32;
} s1;
CHECK_SIZE(s1,9)
```

marsupial:
I don't know what exactly that code is trying to accomplish, but it is interesting that it crashes cling outright....

As I see it to do the test properly, you can't use a heap allocated symbol, as sometimes the symbol would be aligned properly, sometimes not...so you'd have to construct a binary to do this.

If you know an easier way or am seeing something I'm not that why not just write the test?

vgvassilev:
Merged, thanks for your quick followup and contribution!

PR #64, opened 2016-12-01
Fixes for Windows

marsupial:


vgvassilev:
@marsupial Could you fix the build of this PR?

karies:
Thanks a *lot* for compiling these patches!

362a4c8 can probably be removed; de31255 has a different patch in the master. I forgot to sync yesterday's changes to textinput back into cling. Could you check, please?

lib/Interpreter/DeclUnloader.cpp
```diff
@@ -649,7 +649,12 @@ bool DeclUnloader::VisitRedeclarable(clang::Redeclarable<T>* R, DeclContext* DC)
     GD = GlobalDecl(CXXDtor, Dtor_Base);
     MaybeRemoveDeclFromModule(GD);
     GD = GlobalDecl(CXXDtor, Dtor_Comdat);
+#if defined(LLVM_ON_WIN32)
+    // MicrosoftMangle.cpp:954 calls llvm_unreachable when mangling Dtor_Comdat
```
vgvassilev: Is that a current limitation of the MS mangler?

If this was something we shouldn't try to mangle, can't we skip such declarations as `Dtor_Comdat` in `MaybeRemoveDeclFromModule` instead?

vgvassilev:
Commit cc1760b8cc12c3804568be4f6a13c73c5bd28474 looks good to me but I'd like @karies to review it, too.

lib/Interpreter/IncrementalExecutor.cpp
```diff
@@ -44,45 +44,17 @@ using namespace llvm;
...
-
-// Keep in source: ~unique_ptr<ClingJIT> needs ClingJIT
-IncrementalExecutor::~IncrementalExecutor() {}
-
-std::unique_ptr<TargetMachine>
-  IncrementalExecutor::CreateHostTargetMachine(const
-                                           clang::CodeGenOptions& CGOpt) const {
-  // TODO: make this configurable.
+static std::unique_ptr<TargetMachine>
+CreateHostTargetMachine(const clang::CodeGenOptions& CGOpt, unsigned Fmt = 0) {
```
vgvassilev: Could you document the use of `Fmt` for future reference. Please add some documentation explaining that we default to Triple unless Fmt is specified.

marsupial:
Without the unicode flags I get Chinese character in the console.
The recent removal of 'extern' in 43c8a82 should be backed out, it's a requirement for mangling on Windows and proper exporting on OS X.

I can't see what changes relate to to the comments you are giving.

bellenot:
This was fixed in the master. See
https://root.cern.ch/gitweb/?p=root.git;a=commitdiff;h=a6cc9ef3ab89c0392fbf3546f9354ac215aaa2b3
We removed all the “add_definitions(-UUNICODE -U_UNICODE)” in CMakeLists.txt and called the ASCII versions of Windows functions (https://root.cern.ch/gitweb/?p=root.git;a=commitdiff;h=5cdda276c1b8bf5c3380cbf0de1dd8e8d1533d39;hp=a6cc9ef3ab89c0392fbf3546f9354ac215aaa2b3)
Cheers, Bertrand.

marsupial:
I don't know what 'fixed' means.
The removal of
`add_definitions(-UUNICODE -U_UNICODE)` in UserInterface causes Chinese character in the console.

bellenot:
Please read carefully my previous reply. To fix the issue, we had to remove the add_definitions(-UUNICODE -U_UNICODE)everywhere AND call the appropriate functions. As you can see in this diff:
https://root.cern.ch/gitweb/?p=root.git;a=commitdiff;h=5cdda276c1b8bf5c3380cbf0de1dd8e8d1533d39;hp=a6cc9ef3ab89c0392fbf3546f9354ac215aaa2b3

include/cling/Utils/Platform.h
```diff
@@ -22,6 +22,16 @@ namespace platform {
   ///
   std::string GetCwd();
 
+  ///\brief Replace all $TOKENS in a string with environent variable values.
+  ///
+  /// \param [in/out] Str - String with tokens to replace (in place)
+  /// \param [in/out] Path - Check if the result is a valid filesystem path.
```
marsupial: Does not in/out specify that the parameter can mutate?

lib/Interpreter/CIFactory.cpp
```diff
@@ -203,6 +203,13 @@ namespace {
       }
 #endif
 
+      // Ignore warnings about redeclaration with/without dllexport & dllimport
+      sArguments.addArgument("-Wno-dll-attribute-on-redeclaration");
+      sArguments.addArgument("-Wno-inconsistent-dllimport");
```
marsupial: The lack of many tests failing is the test...
`extern "C" int printf(const char*, ...);` used in many tests.
`__declspec(dllexport) __cdecl int printf(const char*, ...);` according to the headers.

lib/Interpreter/IncrementalExecutor.cpp
```diff
@@ -121,6 +87,34 @@ std::unique_ptr<TargetMachine>
                                           OptLevel));
   return TM;
 }
+} // anonymous namespace
+
+IncrementalExecutor::IncrementalExecutor(clang::DiagnosticsEngine& diags,
+                                         const clang::CompilerInstance& CI):
+  m_externalIncrementalExecutor(nullptr)
+#if 0
```
marsupial: Because currently m_Diags is non existent in the structure.

lib/MetaProcessor/MetaParser.cpp
```diff
@@ -193,50 +194,66 @@ namespace cling {
...
-      } else {
-        // check whether we have >>
-        if (getCurTok().is(tok::greater)) {
-          append = true;
-          consumeToken();
+
+        const Token& Tok = lookAhead(2);
+        if (Tok.is(tok::constant)) {
+          switch (Tok.getConstant()) {
+            case 1: file = llvm::StringRef("&1"); break;
```
marsupial: ?

lib/Interpreter/DeclUnloader.cpp
marsupial: Yes limitation of LLVM MS name mangling.

karies:
I started reviewing but I'm only back at work next week; will send it out on Monday.

marsupial:
The first 3 could be cherry-picked, not totally related to Windows.

karies:
General comment: PRs with > 5 commits are *really* difficult to review, especially because it's a bag of everything so everyone has comments and so everything is always in flux / changes while we review it.

Is there a chance you can split things into smaller portions, e.g. the lit stuff, the COFF stuff, test fixes, symbols etc. I'd much rather have 100 PRs than a PR with 200 commits...

lib/Interpreter/IncrementalExecutor.cpp
vgvassilev: Please remove it then.

marsupial:
I was asked to submit all the Windows changes as on PR.

karies:
One PR for Windows: whoever asked that underestimated the amount of work you put it :-) Can you split it?

lib/Interpreter/CIFactory.cpp
vgvassilev: Ok, could you turn this into a comment?

lib/MetaProcessor/MetaParser.cpp
vgvassilev: Who is providing the storage for `&1`?

marsupial:
I'll update and rebase after the first reviews today, and can resubmit a smaller portion, but please keep this in mind:

We've been down this road a few times now, one person asks one thing, another something else.
Bit hesitant to split it up. What ends up happening is everyone chimes in with their 'better' way of doing it and I'm left doing a ton more work both by splitting it and resolving conflicts.

It's already happening regarding the unicode fixes.

lib/Interpreter/DeclUnloader.cpp
vgvassilev: Can you check for this in the `MaybeRemoveDeclFromModule`, checking for such kinds on declarations and skipping them?

include/cling/Utils/Platform.h
marsupial: True, only saw the Str param.

lib/Interpreter/CIFactory.cpp
marsupial: Tell me what you'd like it to say and I'll paste it in.

lib/Interpreter/IncrementalExecutor.cpp
marsupial: I didn't add it.
I've enabled its use in a subsequent commit to actually report errors properly.

lib/MetaProcessor/MetaParser.cpp
marsupial: It's a const char*, those are stored in the text section. The pointer to it has the scope.

lib/Interpreter/CIFactory.cpp
vgvassilev: "// Windows headers add "__declspec(dllexport) __cdecl" to routines such as printf,
 // causing a lot of warnings for different redeclarations (eg. coming from the test suite).
 // Do not warn about such cases."

lib/Interpreter/IncrementalExecutor.cpp
vgvassilev: Ok.

lib/MetaProcessor/MetaParser.cpp
vgvassilev: Uh, ok, thanks!

lib/Interpreter/DeclUnloader.cpp
vgvassilev: We should teach `utils::Analyze::maybeMangleDeclName` about this current (?) limitation of Windows mangler and return the non-mangled name if the dtor is Comdat and we are using the Windows mangler. This way other users of `utils::Analyze::maybeMangleDeclName` would benefit from this.

karies:
Bertrand and I know *exactly* what the unicode thing is about. We have been improving cling on Windows, there is overlap with what you did, but some parts are only in your PR, while other parts are not in your PR (but instead you tried to fix the issue at hand, instead of the more fundamental solution).

Sorry, but I see no way out: we want to progress, we will need to discuss the fundamental parts of this, Bertrand and I have experience with the Windows port, too, and we need to converge, together.

As I said, some parts are just fine and can be merged immediately, but I cannot green light them because all I have is one large PR.

marsupial:
I already said I would try to break it down later today. I'm not saying everything should be taken as is either...clearly there is room for discussion.

All I've asked is to please keep in mind that I've noticed a pattern to how this goes where I'm often [t]asked to do work that could have been avoided.

lib/Interpreter/Interpreter.cpp
```diff
@@ -349,9 +349,11 @@ namespace cling {
       Strm << Linkage << " int __cxa_atexit(void (*f) (void*), void*, void*);\n";
 
       // C atexit, std::atexit
+#if !defined(LLVM_ON_WIN32) || !defined(_M_CEE_PURE)
```
marsupial: RuntimeUniverse.h should remain [somewhat] optional in my opinion. Putting it there would be ugly and require the file always be included. If RuntimeUniverse.h is missing/screwy that would mean destructors would never be called, as well as any usage of atexit crashing on shutdown.

vgvassilev:
On 01/12/16 22:16, marsupial wrote:
>
> *@marsupial* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In lib/Interpreter/Interpreter.cpp 
> <https://github.com/root-mirror/cling/pull/64>:
>
> > @@ -349,9 +349,11 @@ namespace cling {
>         Strm << Linkage << " int __cxa_atexit(void (*f) (void*), void*, void*);\n";
>   
>         // C atexit, std::atexit
> +#if !defined(LLVM_ON_WIN32) || !defined(_M_CEE_PURE)
>
> RuntimeUniverse.h should remain [somewhat] optional in my opinion. 
> Putting it there would be ugly and require the file always be 
> included. If RuntimeUniverse.h is missing/screwy that would mean 
> destructors would never be called, as well as any usage of atexit 
> crashing on shutdown.
>
RuntimeUniverse.h is (must be) always included by the interpreter. This 
should not be a concern.

karies:
Merged, thanks a lot! Next one ;-)

marsupial:
While RuntimeUniverse.h is **_currently_** always included, it needn't be if you are not using _gCling_ or _value-printing_.  There are certain circumstances where it is better no to (trying to debug Decl unloading). Additionally if the file gets mucked up somehow or missing that would mean destructors would stop working entirely.

PR #65, opened 2016-12-02
TTY fixes

marsupial:
Windows currently dumps input so `cat F.C | cling.exe` will show F.C contents on the console.
Make `cling > Out.log` show `[cling$]` prompt.

### DIFF
--- a/lib/Interpreter/CIFactory.cpp
+++ b/lib/Interpreter/CIFactory.cpp
@@ -352,7 +352,9 @@ namespace {
     Opts.RTTIData = 0;
     Opts.setDefaultCallingConv(clang::LangOptions::DCC_CDecl);
 #ifdef _DEBUG
-    Opts.setStackProtector(clang::LangOptions::SSPStrong);
+    // FIXME: This requires bufferoverflowu.lib, but adding:
+    // #pragma comment(lib, "bufferoverflowu.lib") still gives errors!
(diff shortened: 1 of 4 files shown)


karies:
Could you motivate why you want to see the prompt when piping? If I do
```
$ echo 'print 1
print 2
print 3' | python -
```
I get
```
1
2
3
```
I think I'd prefer this, because it allows a C++ script to generate output that's useful by itself, e.g. to pipe into the next command. What do you think? I.e. I think I'd go the other way, and suppress 'cling$' from Windows non-tty.

marsupial:
The point is to always add the prompt when using a TTY. Could you explain why
`cling > Log.txt` should not show the prompt on the TTY?

You're example functions similar in cling with or without these changes, as it's piped not a tty.
```
echo '1                                                                                                                                                                                                                        2
3' | ./bin/cling --nologo
```
->
```
(int) 1
(int) 2
(int) 3
```

marsupial:
And this actually matches python as well:
```
python > OUT.log  
Python 2.7.10 (default, Jul 14 2015, 19:46:27) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>>
```

marsupial:
The requested changes were added.

karies:
Thanks, merged - with the appropriate fix on ROOT's side. Thank you!

PR #67, opened 2016-12-06
InputValidator fixes

marsupial:


karies:
Great! Alas, this causes a problem in ROOT with this file:
```
$ cat roottest/cling/specialobj/write.C
{
   gROOT->ProcessLine(".L AthIndex.h+");
   gROOT->ProcessLine(".L T0Result.h+");
   gROOT->ProcessLine(".L calib.h+");

#ifdef ClingWorkAroundMissingDynamicScope
   TObject *calib = 0;
   calib = (TObject*)gROOT->ProcessLine("new RTCalib;");
#else
   RTCalib *calib = new RTCalib;
#endif
   TFile* f = TFile::Open("tmp.root", "recreate");
   calib->Write();
   f->Close();

}
```
namely
```
Processing roottest/cling/specialobj/write.C...
Error in cling::MetaProcessor: file write.C is incomplete (missing parenthesis or similar)!
```

marsupial:
I cannot as roottest never worked here, and ROOT is now failing at CMake configuration.

pcanal:
>  ROOT is now failing at CMake configuration.
How is it failing?

marsupial:
I'll email you so I can attach the logs.

marsupial:
That error seems to be from MetaProcessor::readInputFromFile.
That function seems to be transforming the file contents in a variety of ways.

marsupial:
Fixed.
There were also a few problems in MetaProcessor::readInputFromFile, including an API misuse which was the trigger.

karies:
Thanks! I see some more failures from this PR, namely tutorial-graphics-tmathtext, roottest-cling-return-constRetrunTest_WILL_FAIL, roottest-cling-typedef_global-scopeTest2, roottest-root-collection-execTClonesArrayAbsorb-auto - and roottest is still running.

The first issue seems to be due to a change in the resulting Value for readInputFromFile. I will debug this tomorrow and try to fix this myself, as well as cross check the other failures.

marsupial:
Again, I don't care if the last commit is discarded. It would be more efficient to call:
`m_Interp.process(content, result);
`
As meta-commands weren't called with the prior behavior (past the first line) anyway.

karies:
roottest-root-treeformula-array-make fails with
1334: Error in cling::MetaProcessor: file runmultidim.C is incomplete (missing parenthesis or similar)!

even after removing the last commit. I'll debug tomorrow unless you get to it still today.

marsupial:
My guess is the the lines:
`if (_file0)
{`

Could you try collapsing it to:
`if (_file0) {`

marsupial:
Fixed. Though the file `runmultidim.C` proves the last commit should be ignored or the new argument defaulted to false.

karies:
Merged with new argument defaulted to false, as you suggested. Thanks a lot!

PR #69, opened 2016-12-08
Output redirection.

marsupial:


karies:
We have one ROOT test failure from this PR: roottest-root-rint-TabCom. That test pipes "user input" (including tabs) into the process, redirects the completion output and compares it to the expected output. That piping fails withe PR.

I currently don't see how we can simulate user input (even with custom pty-s) with this PR. Do you have a proposal?

marsupial:
Running **driveTabComp.py** via **cling** is working as expected, both the pipe and tab completion work...
Is there anything more specific about the error?

marsupial:
Ok I see it now when incomplete lines are piped in and then added back after the prompt.

marsupial:
Could it be the failure is correct behavior?
The test is sending cling:
`TObj\t\n` which should fail as it emulates:
```
root [0] TObj\t
TObjArray
TObjArrayIter
TObjString
TObject
TObjectRefSpy
TObjectSpy
TObjectTable
root [0] TObj\n
input_line_9:2:3: error: use of undeclared identifier 'TObj'
 (TObj)
  ^
root [1] 
```

marsupial:
Sorry for all the noise...begining to understand how roottest test stdout/err.
`cat TabCom_input.txt | python driveTabCom.py 2>stderr.log 1>stdout.log `
**stdout.log** matches **TabCom.oref**
**stderr.log** matches **TabCom.eref** except the first line:
`input_line_9:2:3: error: use of undeclared identifier 'TObj'` TabCom.eref
`input_line_10:2:3: error: use of undeclared identifier 'TObj'` stderr.log

karies:
That test is complex - because we cannot trigger tab completion from piped text, so we have to use a python driver to allocate a pty for the subprocess. Anyway, I now understand where this is coming from: ROOT's tab completion, on seeing `gROOT->[TAB]` does "gROOT", pipes the result into a temporary file and uses the output to determine the type of the expression (here, `gROOT`) to figure out completion options. That redirect to file doesn't work anymore.

In ROOT's core/rint/srcTTabCom.cxx:812 we call the equivalent of
```
interp.process(".> tempFile");
interp.process("interp.process(\"gROOT\")");
interp.process(".>");
```
and that doesn't redirect the gROOT value printing to the temp file anymore.

marsupial:
Have you tried to remove the usage of **MaybeRedirectOutputRAII** in ROOT?

Prior behavior was to always dup _stdout_ & _stderr_ and use **MaybeRedirectOutputRAII** to _redirect_ and _un-redirect_ to those duped FDs. That behavior was a little complicated, and unnecessary when redirection was not used, and particularly bad in the case where an owning process actually uses _stdout_ for its own output.

What **MaybeRedirectOutputRAII** now does is suspend any redirection that is taking place in order to write to the original stdout which should only be used when writing to a prompt/tty.

karies:
Thanks for your contribution, merged!

And keep them coming, please! :-)

PR #71, opened 2016-12-14
Add cling output streams.

marsupial:
Fixes output synch and formatting issues on Windows.

### DIFF
--- a/include/cling/Interpreter/Interpreter.h
+++ b/include/cling/Interpreter/Interpreter.h
@@ -402,7 +402,7 @@ namespace cling {
 
     ///\brief Prints the current include paths that are used.
     ///
-    ///\param[in] S - stream to dump to or nullptr for default (llvm::outs)
+    ///\param[in] S - stream to dump to or nullptr for default (cling::outs)
     ///
     void DumpIncludePath(llvm::raw_ostream* S = nullptr);
 
--- a/include/cling/Utils/Output.h
+++ b/include/cling/Utils/Output.h
@@ -1,0 +1,37 @@
+//--------------------------------------------------------------------*- C++ -*-
+// CLING - the C++ LLVM-based InterpreterG :)
+// author:
+//
+// This file is dual-licensed: you can choose to license it under the University
+// of Illinois Open Source License or the GNU Lesser General Public License. See
+// LICENSE.TXT for details.
+//------------------------------------------------------------------------------
+
+#ifndef CLING_OUTPUT_H
+#define CLING_OUTPUT_H
+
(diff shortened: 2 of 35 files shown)


marsupial:
Curious if this has been overlooked?

karies:
No, but it creates a test failure in roottest-root-meta-cmsUnload-make that I'll have to debug. I can reproduce the failure like this:

```
$ cat in.txt 
struct S { ~S() {fprintf(stderr,"FOO!\n");}} s;
$ root.exe -e '#define FOO' -l -b < in.txt 

FOO!
$ root.exe -e '#define FOO' -l -b < in.txt | head
FOO!

$
```
I.e. once I pipe, the order of newline vs stdout changes. I need to understand why to know whether that is a problem or simply irrelevant, in which case I will adjhust the reference output.

marsupial:
On OS X both match. One newline before seemingly from `-e '#define FOO'`, and then the one from printf.

```
root.exe -e '#define FOO' -l -b < in.txt                                                                                              

FOO!

```
```
$ root.exe -e '#define FOO' -l -b < in.txt | head                                                                                       

FOO!

```

karies:
Right - so this PR introduces a platform inconsistency. I realize that this by itself is not very helpful - I will try to figure out tomorrow where this empty line is coming from.

marsupial:
By default C++ output streams are synchronized with their C counterparts. No such contract exists with `llvm::outs` or `llvm::errs`, and things can get even more out of synch when is a user uses `std` streams as `llvm::outs` does it's own buffering.

`std::cout << "1"; llvm::outs() << "2"; printf("3\n");`  -> **`213`**
`std::cout << "1"; cling::outs() << "2"; printf("3\n");` -> **`123`**

I would think moving ROOT to use the added cling output streams would be the best solution (hopefully avoiding any need for explicit flushing), or they could simple return the llvm variants when compiling root.

Additionally line _1130_ of **TCling.cxx** might benefit from using `cling::outs`

karies:
The problem here is that ROOT calls `printf("\n")` (in TRint.cxx:394) and that's of course not synchronized with the test's `fprintf(stderr,...)`. I.e. it's the test's fault; I have fixed it.

Thus: merged, thanks for your contribution!

marsupial:
The last commit seems to have been split in two, wit the second half not making it here?

karies:
Right - I need to handle textinput separately (it sits at different places in ROOT and cling) and thus need to split commits that change both cling and textinput into two. And then I need to upload the textinput commits to both repos. That's done now. Thanks for checking!

PR #72, opened 2016-12-14
Platform specific fixes.

marsupial:
LLVM cannot mangle Dtor_Comdat on Windows.
Some constants are not aliased.
GCC on OS X doesn't like setting declaration.


karies:
Merged - thanks!

PR #77, opened 2016-12-20
Add platform::Demangle function.

marsupial:


lib/Utils/PlatformWin.cpp
```diff
@@ -685,6 +689,16 @@ bool Popen(const std::string& Cmd, llvm::SmallVectorImpl<char>& Buf, bool RdE) {
   return !Buf.empty();
 }
 
+std::string Demangle(const std::string& Symbol) {
+  struct AutoFree {
+    const char* Str;
+    AutoFree(const char* Ptr) : Str(Ptr) {}
+    ~AutoFree() { if (Str) ::free((void*)Str); };
```
karies: As this deletes the string I'd prefer it to take and keep a `char*`, not a `const char`.

lib/Utils/PlatformPosix.cpp
```diff
@@ -199,6 +200,19 @@ bool GetSystemLibraryPaths(llvm::SmallVectorImpl<std::string>& Paths) {
   return true;
 }
 
+std::string Demangle(const std::string& Symbol) {
+  struct AutoFree {
+    const char* Str;
```
karies: Same as for Windows variant.

lib/Utils/PlatformPosix.cpp
```diff
@@ -199,6 +200,19 @@ bool GetSystemLibraryPaths(llvm::SmallVectorImpl<std::string>& Paths) {
...
+    AutoFree(const char* Ptr) : Str(Ptr) {}
+    ~AutoFree() { if (Str) ::free((void*)Str); };
```
karies: Again here, same as for Windows variant.

lib/Utils/PlatformWin.cpp
karies: `free(nullptr)` is just fine, no need to check.

karies:
Thanks, merged!

PR #80, opened 2017-01-20
Fix API usage error in LLVM unicode routines.

marsupial:
Use reinterpret_cast to force compile time errors.


karies:
Great, thanks for the speedy patch! Merged!

PR #81, opened 2017-01-24
Comment redirection test and restore its previous glory.

marsupial:
Testing lack of space in redirection commands already exists and the test that was changed actually tests something else.
Minor quibble, but rather it should fail _only_ for what it was designed to test.


karies:
Thanks and apologies for not realizing that. Merged!

PR #82, opened 2017-01-24
Add 'decl' and 'undo' as predicates to .stats command.

marsupial:
Useful for debugging & required for tests in subsequent PRs.

### DIFF
--- a/include/cling/Interpreter/Interpreter.h
+++ b/include/cling/Interpreter/Interpreter.h
@@ -409,6 +409,13 @@ namespace cling {
     ///
     void DumpIncludePath(llvm::raw_ostream* S = nullptr);
 
+    ///\brief Dump various internal data.
+    ///
+    ///\param[in] what - which data to dump. 'undo', 'ast', 'asttree'
+    ///\param[in] filter - optional argument to filter data with.
+    ///
+    void dump(llvm::StringRef what, llvm::StringRef filter);
+
     ///\brief Store the interpreter state in files
     /// Store the AST, the included files and the lookup tables
     ///
--- a/lib/MetaProcessor/MetaSema.cpp
+++ b/lib/MetaProcessor/MetaSema.cpp
@@ -360,6 +360,7 @@ namespace cling {
       "\n"
       "   " << metaString << "stats [name]\t\t- Show stats for internal data structures\n"
                              "\t\t\t\t  'ast'  abstract syntax tree stats\n"
+                             "\t\t\t\t  'asttree [filter]'  abstract syntax tree layout\n"
                              "\t\t\t\t  'decl' dump ast declarations\n"
                              "\t\t\t\t  'undo' show undo stack\n"
       "\n"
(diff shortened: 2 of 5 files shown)


include/cling/Interpreter/Interpreter.h
```diff
@@ -613,6 +613,7 @@ namespace cling {
 
     clang::CompilerInstance* getCI() const;
     clang::Sema& getSema() const;
+    IncrementalParser& getIncrParser() const { return *m_IncrParser; }
```
karies: We don't want to publish the `IncrementalParser`. Can you implement this differently, e.g. by moving the implementation out of the `IncrementalParser` or by creating a facade interface in `Interpreter`?

lib/MetaProcessor/MetaSema.cpp
```diff
@@ -288,9 +289,14 @@ namespace cling {
   }
 
   void MetaSema::actOnstatsCommand(llvm::StringRef name) const {
-    if (name.equals("ast")) {
-      m_Interpreter.getCI()->getSema().getASTContext().PrintStats();
+    if (name.equals("decl")) {
+      ClangInternalState::printLookupTables(m_MetaProcessor.getOuts(),
+       m_Interpreter.getCI()->getSema().getASTContext());
```
karies: indentation?

include/cling/Interpreter/Interpreter.h
marsupial: Is this really necessary for internal code? The header still isn't public/published.
I've had other uses for getting the `IncrementalParser` and writing facades seems to double the complexity/dependencies rather than reduce it.

lib/MetaProcessor/MetaSema.cpp
marsupial: Fixed

include/cling/Interpreter/Interpreter.h
karies: Vassil feels fairly strongly about this (I don't), so for now I'd appreciate if you could remove this interface.

The goal behind this is not to reduce the interface size. It's to encourage separation of parsing and codegen and jitting, by making cross-talk more difficult.

marsupial: Ok, I'll cave...but I think my point still stands that the separation is actually less as three classes are now involved for one feature.

In hopes to of saving me time later are there any complaints to roll in my suggestions for PR #85 while I'm at it?

karies: Great, thanks!

Regarding PR #85 I'd really prefer to have a separate dot verb, as this is only the first step of the tracing feature.

marsupial:
Done.

include/cling/Interpreter/Interpreter.h
marsupial: Lost in all the noise.
Is this the final word, and if so can the commit just be ignored on you side or must I chop it off?

karies:
Merged. I don't mind the tiny duplication wrt to `.trace`. Thanks!

PR #84, opened 2017-01-25
Define NDEBUG on all platforms.

marsupial:
LLVM headers use NDEBUG to define additional data members, so this should match at runtime.


karies:
Thanks, merged!

PR #85, opened 2017-01-25
adding AST dump functionality

vkhristenko:
- updated productions list for MetaParser
- clang already has ASTDumper/ASTPrinter tree walkers.
- the additional optional identifier is the string to filter on

VK


marsupial:
**clangFrontend** needs to be added to the libs in the project files.
Why do this as a subcommand of the **stats** command (similar to #82)?

vkhristenko:
@marsupial , I assume that is what you meant by project files - CMakeLists

regarding the stats - I don't have the preference, to me stats is ambiguous a bit.
on top stats is expected to return false when there is no identifier, which in case of dumpAST dumps the whole AST, e.g. filter string is empty

VK

marsupial:
Yes **lib/MetaProcessor/CMakeLists.txt**

It's not my call, but the point would be to behave like this:
```
.stats ast -- current behavior
.stats asttree [filter] -- your behavior
```

Having actOnstatsCommand take another optional argument seems a lot less code that others could benefit from too.
I know these changes will conflict downstream for me so I'm also trying to minimize that.

Thats my vote at least.

marsupial:
I'm going to have to rework my patch.
Would you mind if I included this with the behavior I mentioned?

vkhristenko:
@marsupial, that's fine.

VK

karies:
My only quibble with this is `.stats` being a weird verb for dumping AST nodes. In the future I hope to also see dump of a AST and / or IR for each transaction, i.e. a trace mode - and here, too, `.stats` isn't a good fit. I.e. I'd vote for `.trace ast foobar` or `.trace on ir` etc. @vkhristenko - could you do that? And then rename your new `actOn...` and friends accordingly?

marsupial:
I've integrated this into #82 already. And it requires a lot less code using the a method Interpreter::dump.

If you want the credit for the commit or it's decided a new command is really necessary let me know now and I'll chop that commit off.

If it does go in a new command, it seems a new method should be in **ClangInternalState** to be reachable anywhere and avoid the additional dependency on **clangFrontend** in **libMetaProcessor**

vkhristenko:
updated as requested

@marsupial @karies I think you guys decide what is the best way to move forward (new cmd or not).
Indeed, this is my first contribution, but I hope not the last, therefore I won't die if I don't get credits for writting 50 lines of code.

I do agree on the linkage point, it's better to avoid unnecessary deps, therefore the call is sent to interpreter

VK

marsupial:
I think @karies has already decide it should be a separate command.
Also think there't too many cooks so I'll let this be my final thoughts and he can decide how to proceed:

#82 adds a nice interface so that additional methods like **Interpreter::dumpTrace** won't have to be added for every additional command. As it stands now these two PRs are conflicting and someone will have to do additional changes.

vgvassilev:
Perhaps you could use `#pragma clang __debug dump <ident>` to achieve what you want.

karies:
@vgvassilev : `__debug dump` doesn't work that well for our purposes:

```
root [3] #pragma clang __debug dump std::string
lookup results for std:
NamespaceDecl 0x7f96bb922308 prev 0x7f96b909ab38 <input_line_6:1:1, col:43> col:11 std
|-original Namespace 0x7f96b909abc0 'std'
`-NamespaceDecl 0x7f96bb922370 prev 0x7f96b90bc150 <col:17, col:41> col:34 __1 inline
  `-original Namespace 0x7f96b909ae90 '__1'
```
I.e. no mention of "string". It's also a fairly low-level interface. I prefer .dump!

vgvassilev:
Hm..., you don't need the qualification. `#pragma clang __debug dump string`. This gives you the list of 'visible' at this point unqualified names. The advantage of this approach is that you can put the `#pragma ...` wherever you want, even at places where there is no `MetaProcessor` support.

marsupial:
FWIW I have a patch on my branch that makes all commands accessible via `#pragma cling`.
There's currently a backlog so I'll hold off, but don't think it would be too difficult to push it to the front.

karies:
Summary: Vassil and I would like to have this PR with  ".trace", for now with the functionality provided by the PR. @vkhristenko could you update this PR accordingly, please?

vkhristenko:
@karies ,

May be I misunderstood something, but I updated this upon the first request back then (renaming dump -> trace and renaming actions etc...):
```
[cling]$ int aaaa = 5
int bbbb = (int) 5
[cling]$ int bbbb = 10
(int) 10
[cling]$ int cccc = aaaa+bbbb*aaaa
(int) 55
[cling]$ .trace ast cccc
Dumping cccc:
VarDecl 0x7fb2d13cedc8 <input_line_9:2:2, col:23> col:6 used cccc 'int' cinit
`-BinaryOperator 0x7fb2d13cef10 <col:13, col:23> 'int' '+'
  |-ImplicitCastExpr 0x7fb2d13ceef8 <col:13> 'int' <LValueToRValue>
  | `-DeclRefExpr 0x7fb2d13cee28 <col:13> 'int' lvalue Var 0x7fb2d08cee48 'aaaa' 'int'
  `-BinaryOperator 0x7fb2d13ceed0 <col:18, col:23> 'int' '*'
    |-ImplicitCastExpr 0x7fb2d13ceea0 <col:18> 'int' <LValueToRValue>
    | `-DeclRefExpr 0x7fb2d13cee50 <col:18> 'int' lvalue Var 0x7fb2d0d93ba8 'bbbb' 'int'
    `-ImplicitCastExpr 0x7fb2d13ceeb8 <col:23> 'int' <LValueToRValue>
      `-DeclRefExpr 0x7fb2d13cee78 <col:23> 'int' lvalue Var 0x7fb2d08cee48 'aaaa' 'int'
```

Let me know if I misunderstood something
VK

karies:
Silly me! Could you fix the white-space errors? `$ git diff --check`

karies:
And could you add a new test to cling/test/Prompt/MetaProcessor/ that tests this?

vkhristenko:
@karies 

So, I removed all the whitespaces and added a test. If something is not correct about the test - let me know, but I think I did it properly, at least according to llvm-lit specification.

1 question is about this conflict - so I can resolve it of course, but whatever was merged enabled this AST printing/filtering with .stats basically, which made my additions meaningless basically. That's ok - the question is do I just add a test for that then (using this new .stats asttree syntax) or I should resolve the conflict and have this duplication.

VK

marsupial:
I think he still wanted it as a "trace" command.
Maybe just resolve the conflict with what in the tree and adjust MetaSema to use Interpreter::dump ?

karies:
Indeed: I want your .trace (filtering AST printing is just a first step of that) and I'm sure you'll find a reasonable way to reduce code duplication, e.g. as marsupial suggested. Thanks a lot!

vkhristenko:
apology that it took so long.

Now, it's working with .trace and uses actOnstatsCommand - no addition to the MetaSema...

VK

lib/Interpreter/Interpreter.cpp
```diff
@@ -38,8 +38,8 @@
 #include "clang/Basic/TargetInfo.h"
 #include "clang/Basic/SourceManager.h"
 #include "clang/CodeGen/ModuleBuilder.h"
-#include "clang/Frontend/ASTConsumers.h"
 #include "clang/Frontend/CompilerInstance.h"
+#include "clang/Frontend/ASTConsumers.h"
```
karies: We prefer alphabetical order... Can you remove the change to this file?

lib/MetaProcessor/MetaParser.cpp
```diff
@@ -474,6 +474,34 @@ namespace cling {
...
+      consumeToken();
+      skipWhitespace();
+      if (!getCurTok().is(tok::ident))
+          return false;
+      std::string ident = getCurTok().getIdent();
+      consumeToken();
+      skipWhitespace();
+      if (getCurTok().is(tok::eof)) {
+        m_Actions->actOnstatsCommand(ident=="ast" 
+            ? llvm::StringRef("asttree") : llvm::StringRef(ident), llvm::StringRef());
```
karies: isn't the "t" in "ast" already the "tree"? :-) I'd suggest "<complete>" (no identifier can be called `<complete>`) if that is what it means?

lib/MetaProcessor/MetaParser.cpp
```diff
@@ -474,6 +474,34 @@ namespace cling {
...
+        consumeToken();
+      }
+      else if (getCurTok().is(tok::ident)) {
+        m_Actions->actOnstatsCommand(ident=="ast"
+            ? llvm::StringRef("asttree") : llvm::StringRef(ident),
```
karies: Ditto. And this seems to be a replication, can you combine these two if branches with an `||`?

lib/MetaProcessor/MetaParser.cpp
```diff
@@ -474,6 +474,34 @@ namespace cling {
...
+            ? llvm::StringRef("asttree") : llvm::StringRef(ident), llvm::StringRef());
+        consumeToken();
+      }
+      else if (getCurTok().is(tok::ident)) {
+        m_Actions->actOnstatsCommand(ident=="ast"
+            ? llvm::StringRef("asttree") : llvm::StringRef(ident), 
+            getCurTok().getIdent());
+        consumeToken();
+      }
+      else return false;
```
karies: For cling we try to keep the `else` action on a separate line.

lib/MetaProcessor/MetaParser.h
```diff
@@ -47,6 +47,7 @@ namespace cling {
   //                 StoreStateCommand := 'storeState' "Ident"
   //                 CompareStateCommand := 'compareState' "Ident"
   //                 StatsCommand := 'stats' ['ast']
+  //                 trace := 'trace' "Ident" ["Ident"]
```
karies: Isn't the first Ident significant, i.e. one out of `ast` (and in the future `ir`)? Then maybe put `'ast'` there, instead of `"Ident"`? Or did I misread the parser?

test/Prompt/MetaProcessor/Trace.C
```diff
@@ -0,0 +1,15 @@
...
+// LICENSE.TXT for details.
+//------------------------------------------------------------------------------
+
+// RUN: cat %s | %cling | FileCheck %s
+int abcdefghxyz = 10
+//CHECK: (int) 10
+.trace ast abcdefghxyz
+//CHECK: Dumping abcdefghxyz:
+//CHECK: VarDecl {{0x[0-9a-f]+}} <input_line_[[@LINE-3]]:2:2, col:20> col:6 used abcdefghxyz 'int' cinit
+//CHECK: `-IntegerLiteral {{0x[0-9a-f]+}} <col:20> 'int' 10
```
karies: Very nice!

Axel-Naumann:
@vkhristenko ping...

vkhristenko:
@Axel-Naumann , please give me time until monday... if possible... apology...

Axel-Naumann:
@vkhristenko o no hurry - I just wanted to clarify that I am really looking forward to merging this PR!

lib/Interpreter/Interpreter.cpp
vkhristenko: done

lib/MetaProcessor/MetaParser.cpp
vkhristenko: but here, since I didn't create an action for my "trace" cmd, I reuse the "stats" cmd, which requires asttree as second identifier...  so I have to remap........ I didn't touch the stats cmd and we wanted to avoid duplication
the syntax for trace is
.trace ast <filterstr> and for stats is .stats asttree ....

VK

lib/MetaProcessor/MetaParser.cpp
vkhristenko: done

lib/MetaProcessor/MetaParser.cpp
vkhristenko: no action - removed completely.

lib/MetaProcessor/MetaParser.h
vkhristenko: indeed - fixed

lib/MetaProcessor/MetaParser.cpp
marsupial: You can remove the redundant conversion of **ident** to a _std::string_ and then back to a _llvm::StringRef_ starting in MetaParser.cpp at line 485.

I'd also get rid of the name re-mapping, possibly:
```
.stats ast -> Interpreter::dump("ast")
.trace ast -> Interpreter::dump("ast", "*")
.trace ast filter -> Interpreter::dump("ast", "filter")
```

vkhristenko: @marsupial ,

if you mean 
```
Interpreter::dump("asttree", "filter")
```
then I understand, however MetaParser doesn't have a handle to Interpreter!

test/Prompt/MetaProcessor/Trace.C
marsupial: Interpreter::dump understands **asttree** to match the syntax you originally proposed.
If that has changed because **asttree** is an improper noun then Interpreter::dump should not be using it. So  rather than remapping the name make Interpreter::dump understand:
```
dump("ast", "") -> equivalent to .stats ast
dump("ast", "*") -> equivalent to .trace ast
dump("ast", "name") -> equivalent to .trace ast name
```

Should **_istraceCommand_** be **_isTraceCommand_**?

lib/MetaProcessor/MetaParser.cpp
marsupial: The useless conversion from llvm::StringRef to std::string and back to llvm::StringRef is more important to me.

The other thought was that Interpreter::dump understands **asttree** to match the syntax you originally proposed. If that has changed because **asttree** is an improper noun then Interpreter::dump should not be using it either. The two solutions I see are to change **asttree** to **astree** or:

```
m_Interp.dump("ast", "") -> equivalent to .stats ast
m_Interp.dump("ast", "*") -> equivalent to .trace ast
m_Interp.dump("ast", "name") -> equivalent to .trace ast name
```

But that can also be handled in a subsequent PR...

lib/MetaProcessor/MetaParser.cpp
```diff
@@ -474,6 +474,26 @@ namespace cling {
...
 
+  // dumps/creates a trace of the requested representation.
+  bool MetaParser::istraceCommand() {
+    if (getCurTok().is(tok::ident) &&
+        getCurTok().getIdent().equals("trace")) {
+      consumeToken();
+      skipWhitespace();
+      if (!getCurTok().is(tok::ident))
+          return false;
+      std::string ident = getCurTok().getIdent();
```
Axel-Naumann: Can you make this a `StringRef`, as per @marsupial's suggestion?

vkhristenko: done, with the commit 141c5c4

Axel-Naumann:
Thanks a lot, merged!

And next time please simply rebase the branch instead of creating merge commits - then you can keep *your* commit identity instead of me saying "by Viktor Khristenko" :-)

PR #87, opened 2017-01-25
Move IncrementalExecutor::CreateHostTargetMachine into module scope.

marsupial:
No need for this function to be a member.


karies:
Merged, thanks!

PR #88, opened 2017-01-25
Cast function pointers legally.

marsupial:


include/cling/Utils/Casting.h
```diff
@@ -0,0 +1,56 @@
...
+
+    ///\brief Legally cast a void* to a function.
+    ///
+    /// \param [in] The void* to cast
+    ///
+    /// \return The function's address.
+    ///
+    template <class T>
+    T VoidToFunction(void *ptr) {
+      return UintToFunction<T>(static_cast<void*>(ptr));
```
karies: Why do you need to `static_cast` a `void*` to a `void*`? Did you mean `static_cast<uintptr_t>`?

include/cling/Utils/Casting.h
```diff
@@ -0,0 +1,56 @@
...
+  namespace utils {
+
+    ///\brief Legally cast a pointer to a function to void*.
+    ///
+    /// \param [in] The function to cast
+    ///
+    /// \return A void* of the functions address.
+    ///
+    template <class T>
+    void* FunctionToVoid(T *fPtr) {
```
karies: Can we throw in `Ptr`? "Void" Without "Ptr" sounds weird, but then we also need it for Function, so I guess `FunctionPtrToVoidPtr`? Or `FunPtrToVoidPtr` is fine, too.

karies: Could you rename `fPtr` to e.g. `funptr`? Else it's an unfortunate clash with ROOT's coding conventions - yes, not relevant for cling, but that change would still be much appreciated :-)

marsupial:
Sure, was over confident and had many a mistake on the way.
Could you go into Travis and cancel all of this PRs builds prior to the last?

karies:
My tests pass - good enough for me. Thanks for the really nice PR, merged!

PR #89, opened 2017-01-25
Windows: Add compiler flags to ignore linkage spec errors.

marsupial:
So things like `extern "C" int printf(const char*, ...)` work on Windows.
For the tests and so users won't have to declare imported functions with `declspec(dllimport)`


karies:
Good idea, merged!

PR #91, opened 2017-01-25
Fix return type from read and allow a debugger to be attached on OS X.

marsupial:
Currently whenever a debugger is attached, paused, or new breakpoint is set cling will exit.


karies:
Thanks! Merged.

PR #92, opened 2017-01-27
Additional tweaks for recent commits.

marsupial:
Use utils::FunctionToVoidPtr.
Use uintptr_t rather than size_t when casting pointer to integer.
Add test for ROOT-8283.
Fix errors from interrupting Value::print's string output in terminals.


karies:
dea2cd5 is already in the repo; skipped that one. The rest is merged - thanks a lot!

PR #94, opened 2017-01-28
Don’t unload past first transaction.

marsupial:
It is currently possible to unload all Transactions in the Interpreter.
1. Doing this will pretty definitively result in a crash later.
2. Even if it didn't, remaining codegen would be non conforming as static destructors wouldn't be registered.

This makes it impossible to unload  the first Transaction.


karies:
Thanks, merged!

PR #95, opened 2017-02-02
Fixes for some DeclUnloading.

marsupial:
Most of the tests would either crash or fail to fully unload the decls making later inclusion impossible.
Hopefully they aren't dependent on previous commits.

### DIFF
--- a/lib/Interpreter/DeclUnloader.cpp
+++ b/lib/Interpreter/DeclUnloader.cpp
@@ -771,6 +771,38 @@ bool DeclUnloader::VisitRedeclarable(clang::Redeclarable<T>* R, DeclContext* DC)
     else if (NamedDecl* ND = FD->getFriendDecl())
       Successful &= Visit(ND);
 
+    // Is it possible to unload a friend but not the parent class?
+    // This requires adding DeclUnloader as a friend to FriendDecl
+#if 0
+    // Unlink the FriendDecl from linked list in CXXRecordDecl
+    TagDecl* Parent = dyn_cast_or_null<TagDecl>(FD->getDeclContext());
+    if (CXXRecordDecl* CD = dyn_cast_or_null<CXXRecordDecl>(Parent)) {
+      Successful = false;
+      if (FriendDecl *First = CD->getFirstFriend()) {
+        if (First != FD) {
+          FriendDecl *Prev = First;
+          for (FriendDecl* Cur : CD->friends()) {
+            if (Cur == FD) {
+              Prev->NextFriend = Cur->NextFriend;
+              Successful = true;
+              break;
+            }
+            Prev = Cur;
+          }
+        } else {
+          Successful = true;
+          CD->data().FirstFriend = nullptr;
+        }
+      }
+
+      // Arrived back here through recursion?
+      if (!Successful)
+        return true;
+
+      Successful = true;
+    }
+#endif
+
     Successful &= VisitDecl(FD);
     return Successful;
   }
--- a/lib/Interpreter/DeclUnloader.cpp
+++ b/lib/Interpreter/DeclUnloader.cpp
@@ -557,8 +557,8 @@ bool DeclUnloader::VisitRedeclarable(clang::Redeclarable<T>* R, DeclContext* DC)
     // We start with the decl context first, because parameters are part of the
     // DeclContext and when trying to remove them we need the full redecl chain
     // still in place.
-    bool Successful = VisitDeclContext(FD);
-    Successful &= VisitRedeclarable(FD, FD->getDeclContext());
+    bool Successful = VisitRedeclarable(FD, FD->getDeclContext());
+    Successful &= VisitDeclContext(FD);
     Successful &= VisitDeclaratorDecl(FD);
 
     // Template instantiation of templated function first creates a canonical
@@ -711,8 +711,8 @@ bool DeclUnloader::VisitRedeclarable(clang::Redeclarable<T>* R, DeclContext* DC)
 
   bool DeclUnloader::VisitNamespaceDecl(NamespaceDecl* NSD) {
     // NamespaceDecl: NamedDecl, DeclContext, Redeclarable
-    bool Successful = VisitDeclContext(NSD);
-    Successful &= VisitRedeclarable(NSD, NSD->getDeclContext());
+    bool Successful = VisitRedeclarable(NSD, NSD->getDeclContext());
+    Successful &= VisitDeclContext(NSD);
     Successful &= VisitNamedDecl(NSD);
 
     return Successful;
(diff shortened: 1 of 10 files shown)


lib/Interpreter/DeclUnloader.cpp
```diff
@@ -711,10 +736,49 @@ bool DeclUnloader::VisitRedeclarable(clang::Redeclarable<T>* R, DeclContext* DC)
 
   bool DeclUnloader::VisitNamespaceDecl(NamespaceDecl* NSD) {
     // NamespaceDecl: NamedDecl, DeclContext, Redeclarable
-    bool Successful = VisitDeclContext(NSD);
-    Successful &= VisitRedeclarable(NSD, NSD->getDeclContext());
+
+    // When unloading a NamedDecl within an inline namespace within std the
+    // NamedDecl needs to be removed from the parent std as well.
+    // Not sure why, but test/CodeUnloading/NameSpaces.C demonstrates
```
vgvassilev: The lookup tables of the inline namespaces are in their enclosing parent namespaces. So we should be asking if the NSD is inline and we should get the Enclosing lookup namespace. This should be reproducible outside of the namespace `std` too.

marsupial: As far as I remember it only affects **std** 
The test is already testing other namespaces as well.

vgvassilev: You probably need a friend declaration in the inlined namespace.

karies: @marsupial - let us know when you can reproduce this outside `std` or whether Vassil should provide more details.

marsupial: If he know otherwise..but I cannot reproduce it outside of std. The test explicitly makes a case that it only affects std.
I'm pretty convinced it is specific to std as the comment says.

vgvassilev: Could you reduce the std-related issue to a few lines. Creduce/delta can help.

marsupial: I think it's on you to clarify what you are saying or provide info that my conclusion is inaccurate. I've already investigated, fixed, and wrote a test.
The tests were as small as I could get it:
```
test/CodeUnloading/NameSpaces.h
test/CodeUnloading/NameSpaces.C
```

#105 can help as well.

vgvassilev: Ok, I can reformulate: when testing unloading I think it doesn't make sense to `#include <new>` or other stl headers. This makes the tests very library implementation dependent. I'd appreciate instead of the header <new>, to have a header which only expose the issues you fix and not more.

marsupial: The tests don't rely on any std headers, just use lines 12-24 that tests the issue.
Later in that file there is an unloading of an included `std except` but that just tests the larger issue that exposed the problem.

My point is that it seems you are speculating what I've said is not right. Its up to you to proove that this is in fact more general of an issue as I've done everything I could and came to the conclusion **IT ONLY AFFECTS STD!**

vgvassilev: I am confused.

```
// T.h 
 namespace A {
   inline namespace __1 { int f; }
 }
// B.cpp
namespace A{}

#include "T.h"

```

Clang seems to put the lookup information in the parent non-inline enclosing namespace:

```
clang -std=c++11 -fsyntax-only -Xclang -ast-dump-lookups -Xclang -ast-dump-filter -Xclang A B.cpp 
Dumping A:
StoredDeclsMap Namespace 0x7f8253837e50 'A'
|-DeclarationName 'f'
| `-Var 0x7f8253837fa0 'f' 'int'
`-DeclarationName '__1'
  `-Namespace 0x7f8253837f20 '__1'

Dumping A:
Lookup map is in primary DeclContext 0x7f8253837e78
```
It seems that unloading T.h would leave the lookup table of A broken.

marsupial: But it would be impossible to unload T.h without first unloading B.cpp!

marsupial: Have you run the tests in cling before and after the patches?

pcanal: @marsupial 

> As far as I remember it only affects std

Out of curiosity, do you understand what makes 'std' special in this regard?

Thanks,
Philippe.

marsupial: I do not, but behind the scenes clang hold's onto **many** _special_ declarations in _special_ ways. Seems to me someone may have found an optimization that made inline namespaces in std behave differently as in C++11 all implementations I've seen are using this.

vgvassilev: Current master:
```
[cling]$ .rawInput
Using raw input
[cling]! namespace A{}
[cling]! namespace A { inline namespace __BBB { int f; } }
[cling]! auto var = A::f;
[cling]! .undo 2
[cling]! auto var1 = A::f;
IncrementalExecutor::executeFunction: symbol '_ZN1A5__BBB1fE' unresolved while linking function '_GLOBAL__sub_I_cling_module_4'!
You are probably missing the definition of A::__BBB::f
Maybe you need to load the corresponding shared library?
[cling]!
```
Your PR:
```
[cling]$ .rawInput
Using raw input
[cling]! namespace A{}
[cling]! namespace A { inline namespace __BBB { int f; } }
[cling]! auto var = A::f;
[cling]! .undo 2
[cling]! auto var1 = A::f;
IncrementalExecutor::executeFunction: symbol '_ZN1A5__BBB1fE' unresolved while linking function '_GLOBAL__sub_I_cling_module_4'!
You are probably missing the definition of A::__BBB::f
Maybe you need to load the corresponding shared library?
```

marsupial: Ok, am I missing something or is the error output the same?
Exactly how is this related to this PR?

marsupial: And doesn't the `.undo 2` mean there is no longer an `f`, so the error is correct?

pcanal: I suspect Vassil is pointing the behavior is not changed with the PR (as expected since the change is specific to std).

> And doesn't the .undo 2 mean there is no longer an f, so the error is correct?

Wouldn't the .undo remove the decl for A::[__BBB]::f and thus, shouldn't it be claiming that A::f does not exist rather than failing to find the symbol?

Cheers,
Philippe.

vgvassilev: @marsupial: the error shows that the lookup succeeded but the linker cannot resolve the address because it is gone.

Here is how the error should look like after unloading:
```
[cling]$ .rawInput
Using raw input
[cling]! namespace A{}
[cling]! auto var1 = A::f;
input_line_4:1:16: error: no member named 'f' in namespace 'A'
auto var1 = A::f;
            ~~~^
[cling]!
```

vgvassilev: I suspect if you remove the if-stmt on line 747 the example I gave you would be unloaded correctly.

marsupial:
Ok, inserted a commit that fixes the issue you demonstrated.
I'd really like to keep it separate; looks to cause no new issues on my HEAD, but the prior behavior has been working for months.

vgvassilev:
Ok, thanks!

Having smaller test cases is always better in many situations, esp in error recovery and unloading.

karies:
Thanks a lot, Roman. These unloading patches are really crucial for us.

Merged!

PR #98, opened 2017-02-02
Fix issue with stored state potentially referencing freed memory.

marsupial:


lib/Interpreter/Interpreter.cpp
```diff
@@ -1177,6 +1177,27 @@ namespace cling {
   }
 
   void Interpreter::unload(Transaction& T) {
+    // Clear any stored states that reference the llvm::Module.
+    // Do it first in case
```
karies: "...in case" what? (Merging nonetheless, but I'd appreciate an email or a follow-up PR. I can guess but you know!)

karies:
Merged - thanks!

lib/Interpreter/Interpreter.cpp
marsupial: I guess it's a euphemism, generally bad things, like "just in case"
Specifically, do it first while we know Transaction& T to be valid.

PR #99, opened 2017-02-05
added a command line flag and associated code to support dumb terminals

aperotte:
Emacs and other subshells that manage their own input result in noisy output at the cling interpreter.  More specifically, entering "int x = 5;" at the prompt will result in the interpreter recursively printing the input adding one character each time - "iinintint int xint x int x =int x = int x = 5int x = 5;int x = 5;".   This seems to be because cling cannot control the cursor location in dumb terminals and therefore cannot update the text at the prompt with each new character.  It instead prints each update.

Subshells such as those in emacs, manage user input until a newline and then send the input to the underlying process.  The output of the underlying process is echoed in the subshell.  Because this shell is managed and cling does not have as much control, this results in noisy output.

This pull request adds a command line flag to indicate whether this is a dumb terminal.  If so, it does not update the content of the terminal on each input character and does not add a newline (since the dumb terminal would add its own).  This is implemented as a member variable of the TextInput class.

### DIFF
--- a/lib/Interpreter/InvocationOptions.cpp
+++ b/lib/Interpreter/InvocationOptions.cpp
@@ -63,6 +63,7 @@ namespace {
                                InputArgList& Args) {
     Opts.ErrorOut = Args.hasArg(OPT__errorout);
     Opts.NoLogo = Args.hasArg(OPT__nologo);
+    Opts.DumbTerm = Args.hasArg(OPT__dumbterm);
     Opts.ShowVersion = Args.hasArg(OPT_version);
     Opts.Help = Args.hasArg(OPT_help);
     if (Arg* MetaStringArg = Args.getLastArg(OPT__metastr, OPT__metastr_EQ)) {
@@ -136,7 +137,7 @@ void CompilerOptions::Parse(int argc, const char* const argv[],
 }
 
 InvocationOptions::InvocationOptions(int argc, const char* const* argv) :
-  MetaString("."), ErrorOut(false), NoLogo(false), ShowVersion(false),
+  MetaString("."), ErrorOut(false), NoLogo(false), DumbTerm(false), ShowVersion(false),
   Help(false) {
 
   ArrayRef<const char *> ArgStrings(argv, argv + argc);
--- a/lib/UserInterface/UserInterface.cpp
+++ b/lib/UserInterface/UserInterface.cpp
@@ -98,6 +98,9 @@ namespace cling {
     UITabCompletion* Completion =
                       new UITabCompletion(m_MetaProcessor->getInterpreter());
     TI.SetCompletion(Completion);
+    if (m_MetaProcessor->getInterpreter().getOptions().DumbTerm) {
+        TI.SetIsDumbTerm(true);
+    }
 
     std::string Line;
     std::string Prompt("[cling]$ ");
--- a/lib/UserInterface/textinput/TextInput.cpp
+++ b/lib/UserInterface/textinput/TextInput.cpp
@@ -31,6 +31,7 @@
 namespace textinput {
   TextInput::TextInput(Reader& reader, Display& display,
                        const char* HistFile /* = 0 */):
+  fIsDumbTerm(false),
   fMasked(false),
   fAutoHistAdd(true),
   fLastKey(0),
@@ -61,9 +62,11 @@ namespace textinput {
     }
     fContext->GetEditor()->ResetText();
 
-    // Signal displays that the input got taken.
-    std::for_each(fContext->GetDisplays().begin(), fContext->GetDisplays().end(),
-             std::mem_fun(&Display::NotifyResetInput));
+    if (!fIsDumbTerm) {
+        // Signal displays that the input got taken.
+        std::for_each(fContext->GetDisplays().begin(), fContext->GetDisplays().end(),
+                      std::mem_fun(&Display::NotifyResetInput));
+    }
 
     ReleaseInputOutput();
 
@@ -124,7 +127,9 @@ namespace textinput {
              || (*iR)->HaveBufferedInput()) {
         if ((*iR)->ReadInput(nRead, in)) {
           ProcessNewInput(in, R);
-          DisplayNewInput(R, OldCursorPos);
+          if (!IsDumbTerm()) {
+              DisplayNewInput(R, OldCursorPos);
+          }
           if (fLastReadResult == kRREOF
               || fLastReadResult == kRRReadEOLDelimiter)
             break;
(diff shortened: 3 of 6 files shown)


karies:
Thanks, Adler!

> therefore cannot update the text at the prompt with each new character. It instead prints each update.

I was hoping for textinput to only print that new character. That's what I expected it to do - instead of re-writing the whole line. Can you find out / remind us why it prints everything again and again when adding a single character?

If we manage to fix that (i.e. only add that new character) then we don't need a flag at all - everything will work, magically :-)

aperotte:
I'll take a look!  However, I do think it's still a better user experience if the input is not echoed at all on these dumb terminals because the input will already be there and would be duplicated if there is any printing on user input.

karies:
Let's step back for a second: when I run `M-x term` in emacs and then start cling, the prompt without your patch seems to work fine. I have this as TERM:

```
$ echo $TERM
eterm-color
```

What do you have?

aperotte:
Yes, unfortunately, term is not the subshell that the other emacs tools are built on (comint, for example).  I believe that the tools are built on eshell which is why the interactive cling mode that I am working on for emacs has the noisy echo.

```
Welcome to the Emacs shell

~ $ echo $TERM
dumb
~ $ 
```

karies:
Ah that's great! textinput already checks for `$TERM` (in TerminalDisplayUnix.cpp) - can we switch to dumb mode for a dumb TERM instead of adding an extra flag?

aperotte:
Yes, that makes sense!  I'll look into that.

karies:
Any news? I really like your patch, and changing this to be `$TERM` sensitive sounds like a small-ish change?

aperotte:
I've been swamped, but I will work on it very soon

aperotte:
Sorry for the delay.  I pushed an update to my fork that removes the flag and detects when terminal type is "dumb".  However, since I had to make the change to TerminalDisplay, this may affect Windows as well (which I didn't test).  Since TerminalDisplay has a fTerm field that holds the terminal type as a string, I set it to "win" in the subclass TerminalDisplayWin

lib/Interpreter/InvocationOptions.cpp
```diff
@@ -145,9 +145,11 @@ void CompilerOptions::Parse(int argc, const char* const argv[],
 }
 
 InvocationOptions::InvocationOptions(int argc, const char* const* argv) :
+
```
karies: Can you remove the changes in this file from this PR?

lib/UserInterface/UserInterface.cpp
```diff
@@ -98,6 +98,9 @@ namespace cling {
     UITabCompletion* Completion =
                       new UITabCompletion(m_MetaProcessor->getInterpreter());
     TI.SetCompletion(Completion);
+    if (D->GetTERM() &&  strstr(D->GetTERM(), "dumb")) {
+        TI.SetIsDumbTerm(true);
```
karies: Can't this be done inside textinput itself? See below...

lib/UserInterface/textinput/TerminalDisplay.h
```diff
@@ -40,6 +40,8 @@ namespace textinput {
     void Detach();
     void DisplayInfo(const std::vector<std::string>& Options);
     bool IsTTY() const { return fIsTTY; }
+    void SetTERM(const char* TERM) { fTERM = TERM; };
+    const char* GetTERM() { return fTERM; };
```
karies: These interfaces (nor the data member) should not be needed, see below. Instead we'd need an `IsDumb()`.

lib/UserInterface/textinput/TextInput.cpp
```diff
@@ -31,6 +31,7 @@
 namespace textinput {
   TextInput::TextInput(Reader& reader, Display& display,
                        const char* HistFile /* = 0 */):
+  fIsDumbTerm(false),
```
karies: Isn't that a property of the `TerminalDisplay`? I'd prefer to have the interface there.

lib/UserInterface/textinput/TerminalDisplayUnix.cpp
```diff
@@ -119,6 +119,7 @@ namespace textinput {
     TerminalConfigUnix::Get().TIOS()->c_lflag |= ECHOCTL|ECHOKE|ECHOE;
 #endif
     const char* TERM = getenv("TERM");
+    SetTERM(TERM);
```
karies: If `TERM` is `"dumb"`, the dumb flag on the `TerminalDisplay` base should be set.

lib/UserInterface/textinput/TextInput.cpp
```diff
@@ -61,9 +62,11 @@ namespace textinput {
     }
     fContext->GetEditor()->ResetText();
 
-    // Signal displays that the input got taken.
-    std::for_each(fContext->GetDisplays().begin(), fContext->GetDisplays().end(),
-             std::mem_fun(&Display::NotifyResetInput));
+    if (!fIsDumbTerm) {
+        // Signal displays that the input got taken.
+        std::for_each(fContext->GetDisplays().begin(), fContext->GetDisplays().end(),
+                      std::mem_fun(&Display::NotifyResetInput));
```
karies: The `if` should test the display of the current iteration: if it's dumb, don't call NotifyResetInput..

lib/UserInterface/textinput/TextInput.cpp
```diff
@@ -124,7 +127,9 @@ namespace textinput {
              || (*iR)->HaveBufferedInput()) {
         if ((*iR)->ReadInput(nRead, in)) {
           ProcessNewInput(in, R);
-          DisplayNewInput(R, OldCursorPos);
+          if (!IsDumbTerm()) {
```
karies: I'd move that `if` into `DisplayNewInput`, testing the dumbness of the display.

lib/UserInterface/textinput/TextInput.h
```diff
@@ -88,6 +88,13 @@ namespace textinput {
     void EnableAutoHistAdd(bool enable = true) { fAutoHistAdd = enable; }
     void AddHistoryLine(const char* line);
 
+    // Dumb terminal getter and setter
+    bool IsDumbTerm() const {return fIsDumbTerm;}
+    void SetIsDumbTerm(bool isDumbTerm) { fIsDumbTerm = isDumbTerm; }
+
+  protected:
+    bool fIsDumbTerm; // whether this is a dumb terminal or not
```
karies: These should all be moved to `TerminalDisplay`.

aperotte:
I will take a look, thanks for the comments.

lazywithclass:
@Axel-Naumann Hello, I'd love this to get merged. C++ is definitely not my thing but I'm learning it, would you be open to accept the requested changes and possibly give me some advice on the way?

aperotte:
Hi, apologies, but I had to stop working on this pull request because of other responsibilities.  I would also love to see this completed if @rilerez or @lazywithclass has some time to work on the refactoring and new conflicts.

rilerez:
Has there been any progress? If not, I'm happy to give it a go, but I might need some pointers to get started.

ferdymercury:

PR #100, opened 2017-02-07
Add ability to control cling output streams colorization.

marsupial:
Don't know if this never was or was lost during the move to cling::errs() etc, but it's very nice.
Default to colorize when running in a terminal.

### DIFF
--- a/lib/Utils/Output.cpp
+++ b/lib/Utils/Output.cpp
@@ -8,47 +8,82 @@
 //------------------------------------------------------------------------------
 
 #include "cling/Utils/Output.h"
+
+#include "llvm/Support/Format.h"
+#include "llvm/Support/Process.h"
 #include "llvm/Support/raw_os_ostream.h"
-#include <atomic>
+
 #include <iostream>
 
 namespace cling {
   namespace utils {
 
-    ///\brief Unbuffer the llvm::raw_os_ostream wrapper
-    /// Shouldn't affect the underlying iostream buffering
-    template <unsigned I>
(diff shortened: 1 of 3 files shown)


lib/Utils/Output.cpp
```diff
@@ -8,47 +8,82 @@
...
+      return sOut;
     }
 
     llvm::raw_ostream& errs() {
-      static llvm::raw_os_ostream S(std::cerr);
-      // For command line applications, it make sense to run unbuffered
-      // so that we will synch with llvm::errs()
-      Unbuffer<2>(S);
-      return S;
+      static ColoredOutput sErr(std::cerr);
```
karies: Can we not re-use the `raw_fd_ostream` here instead of implementing the color actions ourselves?

marsupial: Large point of having cling::errs / cling::outs was that `llvm::raw_fd_ostream` wasn't synching very well on Windows. By using `llvm::raw_os_ostream` that problem goes away and you get synch with `std::out/cerr` for free. Plus a guarantee of synch with `stdout/stderr` via the std, something else `llvm::outs/errs` could not provide.

karies:
Okay - that's really unfortunate :-( I'd rather have a patch on clang (the we upstream) than having to maintain our own stream that now even needs color support... But I like colors, so this *is* an improvement ;-) Thanks - merged!

marsupial:
I agree its a bit wonky. But colors are great, and it lets the standard handle the threading issues more gracefully, and can easily be removed if `llvm::raw_os_ostream` ever supports coloring.

PR #101, opened 2017-02-07
Add some accessors to Interpreter.

marsupial:
Makes it easier to get a DiagnosticsEngine and current source location (ignoring wrapping)
without knowing about CompilerInstance or other internals.

Really helps moving to using DiagnosticsEngine to report errors upstream.


lib/Interpreter/Interpreter.cpp
```diff
@@ -1034,6 +1037,21 @@ namespace cling {
...
+  clang::SourceLocation Interpreter::getSourceLocation(bool skipWrapper) const {
+    const Transaction* T = getLatestTransaction();
+    if (!T)
+      return SourceLocation();
+
+    const SourceManager &SM = getCI()->getSourceManager();
+    if (skipWrapper) {
+      cling::ostrstream Strm;
+      Strm << "void " << utils::Synthesize::UniquePrefix << m_UniqueCounter
+           << "(void* vpClingValue) {\n ";
```
karies: This looks really fragile - if we ever change `WrapInput()` then we'll get a wrong offset here.

Could we move this code into a separate function and then have it called from here and from `WrapInput()`?

marsupial: The only way I saw to do this was refactor some stuff.
In particular removing the createUniqueWrapper (the optimization of inserting the name into the AST tree seems un-needed and pretty dependent on clang).
Removed some stuff that was duplicate code and not used in either cling or ROOT.

karies:
Perfect, thanks a lot for refactoring that! Merged!

PR #102, opened 2017-02-07
Force __cxa_demangle to null terminate its output.

marsupial:
Turns out some older gcc implementation either do not honor or were written before the docs for `__cxa_demangle`, and they return the size of the buffer allocated not the string length.


karies:
Yup, we saw that, too - merged, thanks!

PR #103, opened 2017-02-08
Fix returning to stdout after redirection.

marsupial:
Calling dup2 to get back to STDOUT_FILENO will reopen stdout without line buffering if no output had been written before the redirect occurred.


karies:
Wonderful, that does it! And I'll still look into reducing the amount of Attach/Detach calls - it's probably related to PR #99 Thanks for this fix and pointing out the Attach issue!

marsupial:
Don't forget the other issue where ROOT should be creating `MetaProcessor::MaybeRedirectOutputRAII` before it writes its prompt.

PR #104, opened 2017-02-08
Fixes for Windows test.

marsupial:
Whether these functions aren't being exported because of CMake, the build configuration, or MSVC thinking they shouldn't be is unknown to me.

### DIFF
--- a/test/Driver/E.C
+++ b/test/Driver/E.C
@@ -6,10 +6,10 @@
 // LICENSE.TXT for details.
 //------------------------------------------------------------------------------
 
-// RUN: %cling -dM -E -xc++ /dev/null | grep CLING | FileCheck %s
+// RUN: echo "" | %cling -dM -E -xc++ -std=c++11 | grep CLING | FileCheck %s
 
 // CHECK: #define __CLING__ 1
 // CHECK: #define __CLING__CXX11 1
-// CHECK: #define __CLING__{{GNUC|clang}}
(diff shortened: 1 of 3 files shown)


test/Driver/E.C
```diff
@@ -6,10 +6,9 @@
 // LICENSE.TXT for details.
 //------------------------------------------------------------------------------
 
-// RUN: %cling -dM -E -xc++ /dev/null | grep CLING | FileCheck %s
+// RUN: echo "" | %cling -dM -E -xc++ -std=c++11 | grep CLING | FileCheck %s
```
karies: Why do we need to pass `-std=c++11`?

test/Driver/E.C
```diff
@@ -6,10 +6,9 @@
...
 
 // CHECK: #define __CLING__ 1
 // CHECK: #define __CLING__CXX11 1
-// CHECK: #define __CLING__{{GNUC|clang}}
```
karies: Can we instead add a `__CLING__MSVC`?

test/Driver/E.C
marsupial: I want to force c++11 for the __CLING__CXX11 check.
At some point (MSVC) c++14/17 may be the default?

test/Driver/E.C
marsupial: Updated.
Seems a wiser solution.

karies:
Yippee, thanks! :-) Merged.

PR #105, opened 2017-02-08
Add -noruntime flag to force cling to load minimum decls at startup

marsupial:
This makes it tons easier to debug DeclUnload and discover what isn't being unloaded.

May be helpful for discussions relating to DeclUnloading as it allows on to run cling in a way where startup only...for example proving the test case in #95 is an accurate


vgvassilev:
This makes sense to me. I think we should make it even a default mode when running cling in the test suite. In the cases where we need extra runtime (eg. ValuePrinting) we should explicitly enable it. Maybe this way we can encourage writing more platform independent tests...

marsupial:
Running it as the default in tests makes no sense as it disables runtime printing.
And as most users will never touch it makes the test suit less reliable.

Any platform dependent behavior need in a test can now be written using cling to preprocess the output and pipe that back into itself.

pcanal:
I agree it seems best to thoroughly test the mode most used by users :)

karies:
Nice, thanks, merged!

PR #106, opened 2017-02-08
Travis: Try to get in the OS X queue faster

marsupial:
CI is always waiting for OS X machines at the end, by interleaving them in the build config try and get into the queue faster while Linux builds can carry on.


karies:
Merged, thanks!

PR #107, opened 2017-02-09
Add additional loading methods to Interpreter

marsupial:
Add additional `Interpreter::loadLibrary` and `Interpreter::loadHeader` calls and have `Interpreter::loadFile` call though to those.

This is mostly in preparation for 1 (maybe 2) more PR than needs to occur before print unloading can work.
They are pretty trivial and the other(s) less so.

### DIFF
--- a/include/cling/Interpreter/Interpreter.h
+++ b/include/cling/Interpreter/Interpreter.h
@@ -569,9 +569,29 @@ namespace cling {
     ///
     std::string lookupFileOrLibrary(llvm::StringRef file);
 
-    ///\brief Loads header file or shared library.
+    ///\brief Loads a shared library.
     ///
     ///\param [in] filename - The file to loaded.
+    ///\param [in] lookup - Whether to try to resolve the filepath
+    ///
+    ///\returns kMoreInputExpected is returned when file could not be found
+    /// otherwise kSuccess or kFailure
+    ///
+    CompilationResult loadLibrary(const std::string& filename,
+                                  bool lookup = true);
+
+    ///\brief Loads header file
+    ///
+    ///\param [in] filename - The file to loaded.
+    ///\param [out] T -  Transaction containing the loaded file.
+    ///\returns result of the compilation.
+    ///
+    CompilationResult loadHeader(const std::string& filename,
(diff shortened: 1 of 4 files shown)


lib/Interpreter/AutoloadCallback.cpp
```diff
@@ -105,14 +105,15 @@ namespace cling {
...
       assert(m_PP);
 
-      auto addFile = [this,decl](llvm::StringRef FileName, bool warn) {
-        if (FileName.empty()) return (const FileEntry*)nullptr;
+      auto addFile = [this,decl](llvm::StringRef FileName, bool warn)
+                                                    -> const clang::FileEntry* {
+        if (FileName.empty()) return nullptr;
 
-        const FileEntry* FE = 0;
+        const clang::FileEntry* FE = 0;
```
karies: Why do you qualify `FileEntry` but none of the other names from `clang`?

marsupial: I chose the same name for a structure in the next PR, and wanted to separate actual behavioral changes with syntax only.

Trying to get to a PR that allows unloading of the printing mechanism and I split it into two parts that are necessary: the original subset of DeclUnloading changes, and the aforementioned one adding **cling::FileEntry**

I guess if the DeclUnloading is delayed, 4e78f9f can just as well be chopped off and I resubmit it with the next PR.

karies:
Merged, thanks!

include/cling/Interpreter/Interpreter.h
```diff
@@ -569,9 +569,29 @@ namespace cling {
...
+                                  bool lookup = true);
+
+    ///\brief Loads header file
+    ///
+    ///\param [in] filename - The file to loaded.
+    ///\param [out] T -  Transaction containing the loaded file.
+    ///\returns result of the compilation.
+    ///
+    CompilationResult loadHeader(const std::string& filename,
+                               Transaction** T = 0);
```
vgvassilev: @marsupial, could you fix the indentation?

marsupial:
Sorry.
https://github.com/root-project/cling/pull/120

PR #109, opened 2017-02-14
Fix ParseTopLevelDecl not always propagated error to Transaction.

marsupial:


karies:
Great, thanks a lot! Merged!

PR #111, opened 2017-02-15
Windows cleanup

marsupial:
Cleanup CMake exports.
Remove horrible string hack, one of the new exports seems to fix the issue.
Remove exception specification from one test.


marsupial:
Dropped the edits to Strings.C until a real fix is in place.

karies:
Thanks, merged!

PR #112, opened 2017-02-16
Fix EscapeSequence::encode returning a StringRef pointing to stack memory

marsupial:
Pretty bad one...surprised no compiler or test has caught it yet.


karies:
Thanks, merged!

PR #116, opened 2017-02-21
Fix CMake for Ninja on Windows.

marsupial:
Seemed as it exposed the CMake error on Windows it might be nice to also build with it on Appveyor.
I have no real preference if Appveyor builds with Ninja in addition to, replacing VStudio generator, or not at all.


karies:
Merged, thanks!

PR #117, opened 2017-02-22
fix typo in jupyer kernel readme.

sglyon:


karies:
Merged, thanks! (It will show up in cling's git in a couple of minutes...)

PR #122, opened 2017-03-08
Add methods for inspecting macros via Interpreter.

marsupial:


include/cling/Interpreter/Interpreter.h
```diff
@@ -712,6 +713,24 @@ namespace cling {
     ///
     void* getAddressOfGlobal(llvm::StringRef SymName, bool* fromJIT = 0) const;
 
+    ///\brief Get a given macro definition by name.
+    ///
+    ///\param[in]  Name - the name of the macro to look for
+    ///
+    ///\returns the MacroInfo if the macro was defined, otherwise null
+    ///
+    const clang::MacroInfo* getMacro(llvm::StringRef Name) const;
```
pcanal: Would it make sense to rename this getCppMacro ? (In some context macro is also sometimes synonym with script)

marsupial: Dunno.
If it were a global function I might agree, but couldn't any future introduction of the functionality you're describing be done using **_command_** instead of **_macro_** and just leave macro to mean what it does in LLVM/clang?

pcanal: > any future introduction 
It was a description of the current usage of the word in ROOT (see for example https://root.cern.ch/doc/v608/TMacro_8h_source.html)

Now on the other hand, you are right, it is indeed right it already scoped by being inside cling::Interpreter.  So fair enough, let's leave it.

thanks,
Philippe

marsupial:
ping

Axel-Naumann:
Nice! Thanks! Merged!

PR #124, opened 2017-03-22
Fix dereferencing NULL and writing to freed memory.

marsupial:
In certain circumstances on failure cling will try to dereference null.
In the same and additional circumstances it attempts to write to freed memory.

### DIFF
--- a/include/cling/Interpreter/Interpreter.h
+++ b/include/cling/Interpreter/Interpreter.h
@@ -126,6 +126,8 @@ namespace cling {
       kExeCompilationError,
       ///\brief The function is not known.
       kExeUnkownFunction,
+      ///\brief The Transaction had no module (probably an error in CodeGen).
+      kExeNoModule,
 
       ///\brief Number of possible results.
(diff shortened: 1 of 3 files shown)


include/cling/Interpreter/Interpreter.h
```diff
@@ -126,6 +126,8 @@ namespace cling {
       kExeCompilationError,
       ///\brief The function is not known.
       kExeUnkownFunction,
+      ///\brief The Transaction had no module (probably an error in CodeGen).
+      kExeNoModule,
```
Axel-Naumann: Do we really need this state? Isn't this `kExeCompilationError` with the `Transaction` having a `nullptr` module?

marsupial: I have no opinion, you can change it to whatever you like.

Axel-Naumann:
Changed my mind, I think adding a new failure state makes sense. Merged, thanks for the PR!

PR #125, opened 2017-03-23
Fix collection printing for pairs, recursion, and std::vector<bool>.

marsupial:


Axel-Naumann:
Great, thanks a lot, Roman! Merged!

PR #127, opened 2017-03-23
Change how Interpreter intercepts library functions at runtime.

marsupial:
Emit llvm::Functions with the (possibly mangled) name as a stub that will forward the first two arguments and hardcode two more: a command ID and user data.

This is done with LLVM IR, so the Interpreter must only know the number of arguments and return type of the real function, so the ASTContext needn’t be polluted with redundant function declarations.

Most importantly this can be done extremely early in initialization, before any files are included so that every call will be intercepted.

### DIFF
--- a/include/cling/Interpreter/Interpreter.h
+++ b/include/cling/Interpreter/Interpreter.h
@@ -19,6 +19,10 @@
 #include <string>
 #include <unordered_map>
 
+#ifdef LLVM_ON_WIN32
+#include <unordered_set>
+#endif
+
 namespace llvm {
   class raw_ostream;
   struct GenericValue;
@@ -190,6 +194,14 @@ namespace cling {
     ///\brief Information about the last stored states through .storeState
     ///
     mutable std::vector<ClangInternalState*> m_StoredStates;
+
+#ifdef LLVM_ON_WIN32
+    // Windows specific _Facet_base pointers registered at runtime.
+    std::unordered_set<void*> m_RuntimeFacets;
+
+    // Windows specific _Facet_base registration function.
+    static void RegisterFacet(void *Facet, void *Interp);
+#endif
 
     ///\brief Worker function, building block for interpreter's public
     /// interfaces.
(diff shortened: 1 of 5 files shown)


Axel-Naumann:
With this PR we now get in ROOT:
```
$ root -l -b
input_line_2:2:71: error: redefinition of 'gCling'
namespace cling { class Interpreter; namespace runtime { Interpreter* gCling=(Interpreter*)0xc55490; }}
                                                                      ^
input_line_2:2:71: note: previous definition is here
namespace cling { class Interpreter; namespace runtime { Interpreter* gCling=(Interpreter*)0x890de0; }}
                                                                      ^
input_line_2:3:9: warning: 'thisCling' macro redefined [-Wmacro-redefined]
#define thisCling ((cling::Interpreter*)0xc55490)
        ^
input_line_2:3:9: note: previous definition is here
#define thisCling ((cling::Interpreter*)0x890de0)
        ^
root [0] .q
```

Could you have a look?

marsupial:
Is there anyway of building ROOT so the slightest change doesn't cause 50% of it to be rebuilt?

Axel-Naumann:
I wish. You're free to create a Jira ticket on it at https://root.cern/bugs - maybe that'll help...

marsupial:
This PR allows cling to be run with -fsyntax-only that will work when using gCling (test/CodeUnloading/AtExit.C) and runtime printing constructs.

_rootcling_ seems to use **fsyntax-only** to mean what **noruntime** does now.
I think the prior behavior adding additional meaning to  **-fsyntax-only** is incorrect, and having that flag work for all sources that cling can read is better.

Adding the following to rootcling_impl.cxx@4116 fixes the error:
`    clingArgs.push_back("-noruntime");
`

The change sadly make the build system spew _**error: unsupported option**_
Where that is occurring is beyond me (not cling).

marsupial:
Making this insertion at line 4146 instead fixes the error without subsequent build errors:
`clingArgsC.push_back("-noruntime");`

marsupial:
Should not be a problem anymore.

PR #130, opened 2017-04-11
Add pkg-config support

dobragab:
Note:
- This pc file expects that cling is installed to /opt/cling. This is true for AUR package, but might be false for other releases.
- Maybe tools/packaging isn't the best place for this file.


ferdymercury:

PR #135, opened 2017-06-05
Travis CI for OS X.

marsupial:
Fix gcc-6 on OS X because of changes in homebrew.
Re-enable build on Xcode - 6.4 for testing on 10.9 older libc++ versions.
Distribute OS X builds more evenly (for wait/queuing purposes).


Axel-Naumann:
Thanks, merged!

PR #136, opened 2017-06-05
Fix and expand string type caching.

marsupial:
cb2384f fixed a bug using a static to cache the type of a std:string; however, if the Transaction owning the 'std::string' type is undone this local variable will then point to invalid memory.

Also adds caching of other std::strings types (wchar, u16, and u32).

### DIFF
--- a/include/cling/Interpreter/Interpreter.h
+++ b/include/cling/Interpreter/Interpreter.h
@@ -48,6 +48,7 @@ namespace clang {
   class Sema;
   class SourceLocation;
   class SourceManager;
+  class Type;
   class PresumedLoc;
 }
 
@@ -195,6 +196,12 @@ namespace cling {
     ///\brief Information about the last stored states through .storeState
     ///
     mutable std::vector<ClangInternalState*> m_StoredStates;
+
+    enum {
+      kStdStringTransaction = 0, // Transaction known to contain std::string
+      kNumTransactions
+    };
+    mutable const Transaction* m_CachedTrns[kNumTransactions] = {};
 
     ///\brief Worker function, building block for interpreter's public
     /// interfaces.
@@ -689,6 +696,12 @@ namespace cling {
     ///\brief Returns the current or last Transaction.
     ///
     const Transaction* getLatestTransaction() const;
+
+    ///\brief Returns a reference to a Transaction known to contain std::string.
+    ///
+    const Transaction*& getStdStringTransaction() const {
+      return m_CachedTrns[kStdStringTransaction];
+    }
 
     ///\brief Compile extern "C" function and return its address.
     ///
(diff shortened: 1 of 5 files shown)


include/cling/Interpreter/Interpreter.h
```diff
@@ -195,6 +196,12 @@ namespace cling {
     ///
     mutable std::vector<ClangInternalState*> m_StoredStates;
 
+    enum {
+      kStdStringTransaction = 0, // Transaction known to contain std::string
+      kNumTransactions
+    };
+    mutable const Transaction* m_CachedTrns[kNumTransactions];
```
pcanal: Could add a comment to explain the choice an array of one element rather than 'just' a pointer?

marsupial: The next PR will expand this to 2 elements.
There are also problems when rolling back to the state before **cling::printValue** was invoked for the first time that can be solved using the same mechanism.

marsupial:
Part 2/3 to get builtin-types, arrays of them, and character literals, to print without RuntimePrintValue.h.

Part 1/4 to support unloading after an invocation of **cling::printValue**.

Part 1/5 to get full C-language runtime printing.

marsupial:
Ping.

include/cling/Interpreter/Interpreter.h
Axel-Naumann: Can you use an in-class initializer `= {}`? I prefer that over the `memset`...

include/cling/Interpreter/LookupHelper.h
```diff
@@ -222,10 +233,8 @@ namespace cling {
     bool hasFunction(const clang::Decl* scopeDecl, llvm::StringRef funcName,
                      DiagSetting diagOnOff) const;
 
-
-    ///\brief Retrieve the QualType of `std::string`.
-    const clang::Type* getStringType();
-
+    ///\brief Retrieve the StringType of given Type.
+     StringType isStringType(const clang::Type* Type);
```
Axel-Naumann: Can we call this `getStringType`, given that it returns more than a `bool`?

marsupial:
As noted, these commits are part of larger things I'm trying to get in. Being that these changes are purely aesthetic I can:
1. Push all relevant commits and then go from there.
2. Delay aesthetic changes until the end of smaller batches.
3. Become unhappy doing unnecessary conflict resolution.

?

Axel-Naumann:
Look - I don't want to see no `memset` :-) I'm okay with you opening all relevant subsequent PRs before merging this in, where one of the subsequent PRs replaces the `memset` by an initializer. Same for the `is` vs `get`.

Right now all I have is a vague prospect that we might remember to fix this after merging your PRs in... I am terrible at keeping state / remembering things for weeks, so this *will* get dropped.

marsupial:
Done.

Axel-Naumann:
Thanks, merged! (There's still a `memset`, I'll fix that for you.)

PR #137, opened 2017-06-05
Refactor specializations of executePrintValue.

marsupial:
The templated variants of ExecutePrintValue can cause multiple instantiations of code that is not dependent on any template parameter, as well a lot of duplicate/dependent function signatures.

### DIFF
--- a/lib/Interpreter/ValuePrinter.cpp
+++ b/lib/Interpreter/ValuePrinter.cpp
@@ -582,55 +582,42 @@ namespace cling {
 
 namespace {
 
-template<typename T, bool> struct ExecutePrintValue;
+static std::string callPrintValue(const Value& V, const void* Val) {
+  Interpreter *Interp = V.getInterpreter();
+  Value printValueV;
 
-template<typename T>
-struct ExecutePrintValue<T, false> {
-  std::string operator()(const Value &V, const T &val) {
-    Interpreter *Interp = V.getInterpreter();
-    Value printValueV;
+  {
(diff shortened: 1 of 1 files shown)


marsupial:
Part 1/3 to get builtin-types, arrays of them, and character literals, to print without RuntimePrintValue.h.

lib/Interpreter/ValuePrinter.cpp
```diff
@@ -581,56 +580,44 @@ namespace cling {
...
 
-      cling::ostrstream Strm;
-      Strm << "cling::printValue(";
-      Strm << getTypeString(V);
-      Strm << (const void*) &val;
-      Strm << ");";
+    cling::ostrstream Strm;
+    Strm << "cling::printValue(";
+    Strm << getTypeString(V);
+    Strm << &Val;
```
Axel-Naumann: This prints the address of the parameter. I doubt that's what we want here?

marsupial: It's as it was before, just the (const void*) cast removed.
I believe it is what is wanted as it's being dereferenced in the `getTypeString` call.
A lot of tests would fail if this wasn't correct.

Axel-Naumann: Right, and the use of `&Val` stays within its scope, so I was wrong and all is good. Thanks for explaining!

Axel-Naumann:
Thanks, merged!

PR #138, opened 2017-06-05
Fix -fsyntax-only flag which I broke in 0d15357.

marsupial:
Also fixes a generated error for C language on gcc for OS X.

### DIFF
--- a/lib/Interpreter/Interpreter.cpp
+++ b/lib/Interpreter/Interpreter.cpp
@@ -248,22 +248,24 @@ namespace cling {
 
     // Now that the transactions have been commited, force symbol emission
     // and overrides.
-    if (const Transaction* T = getLastTransaction()) {
-      if (llvm::Module* M = T->getModule()) {
-        for (const llvm::StringRef& Sym : Syms) {
-          const llvm::GlobalValue* GV = M->getNamedValue(Sym);
-#if defined(__GLIBCXX__) && !defined(__APPLE__)
-          // libstdc++ mangles at_quick_exit on Linux when headers from g++ < 5
-          if (!GV && Sym.equals("at_quick_exit"))
-            GV = M->getNamedValue("_Z13at_quick_exitPFvvE");
-#endif
-          if (GV) {
-            if (void* Addr = m_Executor->getPointerToGlobalFromJIT(*GV))
-              m_Executor->addSymbol(Sym.str().c_str(), Addr, true);
-            else
-              cling::errs() << Sym << " not defined\n";
-          } else
-            cling::errs() << Sym << " not in Module!\n";
+    if (!isInSyntaxOnlyMode()) {
+      if (const Transaction* T = getLastTransaction()) {
+        if (llvm::Module* M = T->getModule()) {
+          for (const llvm::StringRef& Sym : Syms) {
+            const llvm::GlobalValue* GV = M->getNamedValue(Sym);
+  #if defined(__GLIBCXX__) && !defined(__APPLE__)
+            // libstdc++ mangles at_quick_exit on Linux when g++ < 5
+            if (!GV && Sym.equals("at_quick_exit"))
+              GV = M->getNamedValue("_Z13at_quick_exitPFvvE");
+  #endif
+            if (GV) {
+              if (void* Addr = m_Executor->getPointerToGlobalFromJIT(*GV))
+                m_Executor->addSymbol(Sym.str().c_str(), Addr, true);
+              else
+                cling::errs() << Sym << " not defined\n";
+            } else
+              cling::errs() << Sym << " not in Module!\n";
+          }
         }
       }
     }
@@ -346,7 +348,7 @@ namespace cling {
     llvm::SmallString<1024> Buf;
     llvm::raw_svector_ostream Strm(Buf);
     const clang::LangOptions& LangOpts = getCI()->getLangOpts();
-    const void* thisP = static_cast<void*>(this);
+    const void* ThisP = static_cast<void*>(this);
 
     // FIXME: gCling should be const so assignemnt is a compile time error.
     // Currently the name mangling is coming up wrong for the const version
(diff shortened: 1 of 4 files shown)


marsupial:
Ping.

lib/Interpreter/Interpreter.cpp
```diff
@@ -356,95 +365,97 @@ namespace cling {
...
     // assignemnt and get around the mangling issue.
     const char* Linkage = LangOpts.CPlusPlus ? "extern \"C\"" : "";
-    if (!NoRuntime && !SyntaxOnly) {
+    if (!NoRuntime) {
       if (LangOpts.CPlusPlus) {
         Strm << "#include \"cling/Interpreter/RuntimeUniverse.h\"\n"
                 "namespace cling { class Interpreter; namespace runtime { "
-                "Interpreter* gCling=(Interpreter*)" << thisP << "; }}\n";
+                "Interpreter* gCling";
+        AssignIfValue(Strm, ThisP, "Interpreter*", " }}");
```
Axel-Naumann: Why don't you set it to `nullptr` if it's `nullptr`, even for syntax-only? Also, I find the version without the extra function call (i.e. the original code that spells out the `<<`) much simpler to read.

test/Driver/C.c
```diff
@@ -6,11 +6,15 @@
...
 
+// Fix value printing!
+
 int printf(const char*,...);
-printf("CHECK 123\n"); // CHECK: CHECK 123
+printf("CHECK 123 %p\n", gCling); // CHECK: CHECK 123
 
-// fix value printing!
+// expected-no-diagnostics
+.q
```
Axel-Naumann: Can you throw in a newline at the end of the file?

lib/Interpreter/Interpreter.cpp
marsupial: Going off the original version, which did this:
```
    if (!isInSyntaxOnlyMode()) {
      // Set up the gCling variable if it can be used
      initializer << "namespace cling {namespace runtime { "
        "cling::Interpreter *gCling=(cling::Interpreter*)"
        << "0x" << std::hex << (uintptr_t)this << " ;} }";
    }
```

I don't know why it did not set the value before, overall it was wrong as gCling was unavailable with **-fsyntax only**.

But in https://github.com/root-project/cling/pull/127 there was an issue with ROOT in `rootling_impl.cxx` using **-fsyntax-only**, and when gCling was set to a value it was barking loudly, so I just made it a declaration in syntax only mode.

Axel-Naumann: Fine, I can investigate why setting `gCling` would fail in syntax-only (I don't see it yet), and I can do that after merging this PR.

Still, could you remove `AssingIfValue` and just spell out `<<`?

marsupial:
Both done.

Axel-Naumann:
Thanks. I'm rebasing / resolving conflicts and testing on ROOT... I.e. please keep your hands off this one for the moment :-)

lib/Interpreter/Interpreter.cpp
```diff
@@ -356,84 +358,84 @@ namespace cling {
...
+    //     atexit below isn't linking to the __dso_handle symbol.
+
+    Strm << "#define __dso_handle ((void*)" << ThisP << ")\n";
+
+    // Use __cxa_atexit to intercept all of the following routines
+    Strm << Linkage << " int __cxa_atexit(void (*f)(void*), void*, void*) "
+         << cxa_atexit_is_noexcept << ";\n";
+
+    // C atexit, std::atexit
+    Strm << Linkage << " int atexit(void(*f)()) " << Attr << " { return "
```
Axel-Naumann: Do we need this block for the no-runtime case? I'd like to not declare those; else rootcling will dump them into the PCH and we get duplicate definitions, one from the PCH and one from the cling runtime. Objections against that?

lib/Interpreter/Interpreter.cpp
```diff
@@ -356,84 +358,84 @@ namespace cling {
...
+    //      [cling] __cxa_atexit(f, 0, __dso_handle) // seg-fault
+    //  5. Code that actually uses __dso_handle will fail as a declaration is
+    //     needed which is not possible with the macro.
+    //  6. Assuming 4 is sorted out in user code, calling __cxa_atexit through
+    //     atexit below isn't linking to the __dso_handle symbol.
+
+    Strm << "#define __dso_handle ((void*)" << ThisP << ")\n";
+
+    // Use __cxa_atexit to intercept all of the following routines
+    Strm << Linkage << " int __cxa_atexit(void (*f)(void*), void*, void*) "
```
Axel-Naumann: What's your motivation of declaring these for syntax-only? To increase coverage? In principle atexit is a pure CodeGen thing; we won't need it unless we CodeGen, and syntax-only means exactly "no CodeGen".

lib/Interpreter/Interpreter.cpp
marsupial: You're asking if it's possible, not that this is being done, right?
--noruntime still needs everything else to work (including static destruction), it just skips the transformations for value printing and null ref detection.

lib/Interpreter/Interpreter.cpp
marsupial: Yes, predominantly to have -fsyntax-only behave exactly them same (the part that I broke).
I have successfully been able to remove all of the ugliness of needed declarations by directly creating the functions in IR and sending them to the linker, but that change set is much larger.

lib/Interpreter/Interpreter.cpp
Axel-Naumann: In that case, neither --noruntime nor -fsyntax-only will work for rootcling's PCH generation, so that's not an option. I don't want to throw in a third flag.

Please select whether you want to suppress these definitions (all that can trigger duplicate definition errors, i.e. function and variable definitions as well as preprocessor macro definitions) for --noruntime or -fsyntax-only. For me it makes more sense to suppress them for -fsyntax-only, given that we know we won't have static destruction as we won't have any codegen.

marsupial: I agree, it's not an option for --noruntime and only pedantic for -fsyntax-only.

**But then gCling has to go back to how it was prior and not set the value for -fsyntax-only case too ???**

marsupial: If it's just for PCH generation why can't rootcling invoke cling to generate them with:
`-x c++-header Input.h -o Output.pch`

Axel-Naumann: I'm taking care of it - now that I understand what your plan was :-) I need much more explanation of what your goal is to be able to follow what you're doing...

marsupial: So it stays as is now?
(_He said after changing it again_)

Axel-Naumann: We need to "do things" before we can emit the PCH, so no, `-o Output.pch` isn't an option.

I am adding a commit that prevents definitions from being emitted if -fsyntax-only. I think that's a reasonable compromise between what you want and what we need?

marsupial: Ok, at the risk of being asked to change it back I stored the partial revert here:
https://github.com/marsupial/cling/commits/PR/6

Axel-Naumann:
Thanks, merged!

PR #140, opened 2017-06-05
Minor additions / fixes.

marsupial:


Axel-Naumann:
@vgvassilev asked me to hold off with merging the others until he bumped llvm, which according to him should happen still this week.

marsupial:
FWIW I have far more time this week than I will next.

marsupial:
Might be worth getting #135 so CI works properly again.

Axel-Naumann:
LGTM (testing ROOT), but the first commit is already part of PR #139 so I'll skip that. If there is something missing from it (the patch seems to not be a 100% overlap if I interpret git's reaction correctly) then please let me know!

marsupial:
Yes probably just backed out to the wrong place when building get the PR...should be exactly the same.

marsupial:
Yes probably just backed out to the wrong place when building get the PR...change-set should be exactly the same as the one already committed.

Axel-Naumann:
Merged!

PR #141, opened 2017-06-06
Destroy cling::Value elements in reverse order of construction.

marsupial:
Matches C++ better, and fixes a problem where a constructed array element can reference an earlier element already destructed.


Axel-Naumann:
Wonderful, thanks! Merged.

PR #142, opened 2017-06-06
Have MetaProcessor::process take an llvm::StringRef.

marsupial:
Remove unnecessary std::string -> const char* -> std::string conversions.


Axel-Naumann:
Thanks, merged!

PR #145, opened 2017-06-06
Refactor common code for creating modules in IncrementalParser.

marsupial:


lib/Interpreter/IncrementalParser.cpp
```diff
@@ -377,6 +377,18 @@ namespace cling {
     return ParseResultTransaction(T, ParseResult);
   }
 
+  std::string IncrementalParser::makeModuleName() {
+    stdstrstream ModuleName;
+    ModuleName << "cling-module-" << m_ModuleNo++;
+    return std::move(ModuleName.str());
```
Axel-Naumann: I'd prefer
```
return std::string("cling-module-") + std::to_string(m_ModuleNo++);
```
Makes sense?

marsupial: Yes better.
If they didn't make std::to_string so useless, I wouldn't always forget about it.
Added another commit to remove sstream & iostream includes.

marsupial:
These were done.

Axel-Naumann:
Thanks, merged!

PR #146, opened 2017-06-09
Fixes for LLVM upgrade.

marsupial:
Some small things, but the lack of including **StringRef.h** pretty much makes all users include it explicitly.

### DIFF
--- a/include/cling/Interpreter/Interpreter.h
+++ b/include/cling/Interpreter/Interpreter.h
@@ -12,6 +12,8 @@
 
 #include "cling/Interpreter/InvocationOptions.h"
 
+#include "llvm/ADT/StringRef.h"
+
 #include <cstdlib>
 #include <memory>
 #include <string>
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,14 +52,6 @@ matrix:
     # across all platforms. By interleaving OS X, the hope is to get in the
     # queue faster while not blocking Linux builds from occuring.
     - os: linux
-      env: DIST=pangolin COMPILER=g++-4.7 CCOMPILER=gcc-4.7 NIGHTLY_CRON_SYNC=1
-      addons:
-        apt:
-          sources: *sources
-          packages: g++-4.7
-      compiler: gcc-4.7
-
-    - os: linux
       env: DIST=pangolin COMPILER=g++-4.8 CCOMPILER=gcc-4.8
       addons:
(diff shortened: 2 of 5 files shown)


include/cling/Interpreter/Interpreter.h
```diff
@@ -12,6 +12,8 @@
 
 #include "cling/Interpreter/InvocationOptions.h"
 
+#include "llvm/ADT/StringRef.h"
```
vgvassilev: This is intentional. Cling's DynamicLookup `#includes "Interpreter.h"` at runtime. This means that systems like ROOT, need to provide all headers that are included in the Interpreter.h and the indirect includes thereof.

StringRef has been very fluctuating over the upgrades. IIRC, now it includes stuff outside `llvm/ADT` (llvm/Support/STLExtras.h, IIRC). This means that we need even more scaffolding to move. This doesn't simplify the overly complicated llvm upgrade procedure.

That said, @Axel-Naumann is planning to try to remove the dependency on the Interpreter.h in cling's runtime. This would hopefully allow Interpreter.h to be able to include again things like StringRef.

vgvassilev:
Moved (except for `#include "llvm/ADT/StringRef.h"`) [here](https://github.com/root-project/root/pull/625) to make sure ROOT test suite are also happy.

marsupial:
I understand the reason to a degree. The problem is it breaks tests and it just requires the user to do more typing, as they most likely want to use the interface.

vgvassilev:
If we have a macro that we are in testing mode, lets make this inclusion conditional until the proper fix is in place.

A proper fix would be to outline [this](https://github.com/root-project/cling/blob/master/include/cling/Interpreter/DynamicLookupRuntimeUniverse.h#L49) and [this](https://github.com/root-project/cling/blob/master/include/cling/Interpreter/DynamicLookupRuntimeUniverse.h#L64) call to a static library functions taking `gCling` as an argument.

This would allow us to remove the inclusion [here](https://github.com/root-project/cling/blob/master/include/cling/Interpreter/DynamicLookupRuntimeUniverse.h#L16)

Then we would be totally fine to bring back StringRef.

Axel-Naumann:
This is now fixed by 212ad1b7d1c545e8441e56187799ae6ea7c12d6f..0889fabd036441eee98448abc943c64e362823c9

PR #147, opened 2017-06-09
Fix buffer overflow.

marsupial:
Am I reading this wrong or is a buffer of `InputSize + 1` allocated but the final memcpy is equivalent to:
```
Buf[InputSize] = '\n';
Buf[InputSize+1] = 0;
```


vgvassilev:
Clang requires 0 terminated memory buffers but llvm interface doesn't require that, so we have to add a trailing 0 ourselves.

marsupial:
The real issue is writing the null termination outside the allocated region.

Clang doesn't like null termination, because if you null terminate within the memory bounds it will error. Easily verified by just allocating `InputSize + 2` and keeping everything else as it was.

Axel-Naumann:
Indeed. Thanks, merged!

PR #149, opened 2017-06-12
Faster and more efficient to not write the lines that won’t be compared.

marsupial:
State comparison involves comparing the include file cache. Currently all input_line_X entries are ignored. As such just not writing them to disk should be more efficient than ignoring via diff.


Axel-Naumann:
Indeed - it is more efficient, but for an area where we value code maintainability over efficiency. And here, keeping the pattern used for other state ingredients seems to make the code simpler. I hope that's an understandable explanation! Thanks nonetheless - but your other PRs are dramatically more relevant! :-) (I bet also for you.)

PR #152, opened 2017-06-15
Fix gcc-6 on OS X with C++17

marsupial:
Refactors C++ version extraction.

### DIFF
--- a/lib/Interpreter/CIFactory.cpp
+++ b/lib/Interpreter/CIFactory.cpp
@@ -53,6 +53,26 @@ using namespace clang;
 using namespace cling;
 
 namespace {
+  static constexpr unsigned CxxStdCompiledWith() {
+    // Extracted from Boost/config/compiler.
+    // SunProCC has no C++11.
+    // VisualC's support is not obvious to extract from Boost...
+
+    // The value of __cplusplus in GCC < 5.0 (e.g. 4.9.3) when
+    // either -std=c++1y or -std=c++14 is specified is 201300L, which fails
+    // the test for C++14 or more (201402L) as previously specified.
+    // I would claim that the check should be relaxed to:
+#if __cplusplus > 201402L
+    return 17;
+#elif __cplusplus > 201103L
+    return 14;
+#elif __cplusplus >= 201103L
+    return 11;
+#else
+#error "Unknown __cplusplus version"
+#endif
+  }
+
   // This function isn't referenced outside its translation unit, but it
   // can't use the "static" keyword because its address is used for
   // GetMainExecutable (since some platforms don't support taking the
@@ -297,8 +317,18 @@ namespace {
 
     #if defined(__GLIBCXX__)
       // Avoid '__float128 is not supported on this target' errors
-      if (!opts.Language && !opts.StdVersion)
-        sArguments.addArgument("-std=c++11");
+      if (!opts.Language && !opts.StdVersion) {
+        switch (CxxStdCompiledWith()) {
+          case 17:
+            // Hopefully -D__float128=void can be removed when -std=c++17 works.
+            sArguments.addArgument("-D__float128=void");
+            sArguments.addArgument("-std=c++1z");
+            break;
+          case 14: sArguments.addArgument("-std=c++14"); break;
+          case 11: sArguments.addArgument("-std=c++11"); break;
+          default: break;
+        }
+      }
     #endif //__GLIBCXX__
   #endif // __APPLE__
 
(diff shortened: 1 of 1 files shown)


marsupial:
Ping.

lib/Interpreter/CIFactory.cpp
```diff
@@ -53,6 +53,27 @@ using namespace clang;
...
+#if __cplusplus > 201402L
+    return 17;
+#endif
+#if __cplusplus > 201103L
+    return 14;
+#endif
+#if __cplusplus >= 201103L
+    return 11;
+#endif
+    // Compile error is none are true!
```
Axel-Naumann: Can you use `#elif` to combine them, with a final `#else #error`?

lib/Interpreter/CIFactory.cpp
```diff
@@ -53,6 +53,27 @@ using namespace clang;
 using namespace cling;
 
 namespace {
+  static unsigned CxxStdCompiledWith() {
```
Axel-Naumann: Can you mark this constexpr?

lib/Interpreter/CIFactory.cpp
```diff
@@ -297,8 +318,18 @@ namespace {
 
     #if defined(__GLIBCXX__)
       // Avoid '__float128 is not supported on this target' errors
-      if (!opts.Language && !opts.StdVersion)
-        sArguments.addArgument("-std=c++11");
+      if (!opts.Language && !opts.StdVersion) {
+        switch (CxxStdCompiledWith()) {
+          case 17:
+            // Hopefully -D__float128=void can be removed when -std=c++17 works.
+            sArguments.addArgument("-D__float128=void");
```
Axel-Naumann: What is the problem this is solving? (I see an error message, but how to do get it?)

Do we need this even when compiling with clang? Do you have a bug tracker entry for GCC? As is I don't know what to make of this; e.g. https://godbolt.org/g/y9qoMM seems to work just fine with clang and C++1z on Linux. Does this problem only exist on OSX with libstdc++?

I'd really prefer *not* having this in here; `-D__float128=void` risks havoc that is greater than saying "it's a GCC bug" (if it is).

marsupial: Yes it's when cling is complied on OS X with gcc-6 with -std=c++1z.
I need this configuration (gcc on OS X) to work and would like to start testing it against c++17 and this issue showed up again.  I tried to at least move the c++17 test to  gcc-7 but there were even more errors.

I don't know of any better solution, but I'm pretty sure it's not a common configuration.
If it really bothers you just cut it out. But I think Travis is testing this.

Axel-Naumann: Thanks. Then lets see what Travis says without that line, and if it barfs we see how to fix it. We should at least find or file a clang bug in that case.

lib/Interpreter/CIFactory.cpp
```diff
@@ -316,14 +316,9 @@ namespace {
       }
 
     #if defined(__GLIBCXX__)
-      // Avoid '__float128 is not supported on this target' errors
```
marsupial: Please don't remove this comment, the original `sArguments.addArgument("-std=c++11")` was/is properly fixing the **__float128** problem.  The issue is just that it doesn't work for `-std=c++1z`

Axel-Naumann: What does this comment add? We'd *always* pass `-std=c++11`, right? I find this comment very confusing - it's almost like saying "Avoid '__float128 is not supported on this target' errors: pass `-x c++`". I understand clang might have had at some point this issue, but I don't think there is a close enough causal connection to put it here.

On the contrary: the fact that cling on GCC6 / mac complains
```
/usr/local/Cellar/gcc@6/6.3.0/lib/gcc/6/gcc/x86_64-apple-darwin15.6.0/6.3.0/../../../../../../include/c++/6.3.0/type_traits:344:39: error: __float128 is not supported on this target
```
means likely that we are missing an arch flag - completely unrelated to the standard version!

I.e. I acknowledge that it's broken now; I'll try to figure out why. This here is not the *real* reason.

Axel-Naumann: Uh okay, that was easy. It's not an arch issue. type_traits:344 has
```
    342 #if !defined(__STRICT_ANSI__) && defined(_GLIBCXX_USE_FLOAT128)
    343   template<>
    344     struct __is_floating_point_helper<__float128>
```
I.e. we don't have `__STRICT_ANSI__` defined for 1z. Okay, now I see the relevance of -std=c++11 - this defines `__STRICT_ANSI__`. (So this is about `__STRICT_ANSI__` and `__float128` is just a symptom.)

I need to think... But either way, this comment is out of context.

Axel-Naumann:
I'll merge this now to see how Travis reacts. My suspicion is that gcc sees `-std=gnu++1z` instead of `-std=c++1z`. That might be due to CMake - I don't know yet.

lib/Interpreter/CIFactory.cpp
marsupial: That whole block is specifically dealing with gcc on OS X and how to avoid the __float128 problem, so I don't see how the comment it is out of context. The fact it exists and a simple search of the file shows that `-std=c++11` is not in fact ever passed without it.

Can we step back. What are you trying to accomplish?

I can accept `-D __float128` is ugly and can easily add it on top later.
But you don't seem to actively use gcc on OS X, so why is it so important to change anything else in this block which I know solves actual problems and is commented explicitly stating what it is doing?

marsupial:
I know what Travis will do:
Without -D __float128:
https://travis-ci.org/marsupial/cling/jobs/242964027

With -D __float128:
https://travis-ci.org/marsupial/cling/jobs/243101762

Axel-Naumann:
You were working on a symptom, I'm trying to fix the underlying issue. That float issue is just the only one you stumbled across, strict ANSI (if it's that) has many more consequences. As would the wrong arch. I'm taking care of this.

Axel-Naumann:
The `__float128` issue is fixed and documented by eed1ebca03.

PR #155, opened 2017-06-22
Revert "Avoid memset by using in-class member init."

marsupial:
Add comments as to why resetting to zero is necessary.

This reverts commit 65b4cd2cc3150b4d582338cf8920d20d1416681c.

### DIFF
--- a/include/cling/Interpreter/LookupHelper.h
+++ b/include/cling/Interpreter/LookupHelper.h
@@ -57,7 +57,7 @@ namespace cling {
   private:
     std::unique_ptr<clang::Parser> m_Parser;
     Interpreter* m_Interpreter; // we do not own.
-    const clang::Type* m_StringTy[kNumCachedStrings] = {};
+    const clang::Type* m_StringTy[kNumCachedStrings];
 
   public:
     LookupHelper(clang::Parser* P, Interpreter* interp);
--- a/lib/Interpreter/LookupHelper.cpp
+++ b/lib/Interpreter/LookupHelper.cpp
@@ -39,6 +39,7 @@ namespace cling {
   // the dtor on the OwningPtr
   LookupHelper::LookupHelper(clang::Parser* P, Interpreter* interp)
     : m_Parser(P), m_Interpreter(interp) {
+    // m_StringTy is always properly initialized in getStringType
   }
 
   LookupHelper::~LookupHelper() {}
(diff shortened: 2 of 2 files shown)


include/cling/Interpreter/LookupHelper.h
```diff
@@ -57,7 +57,7 @@ namespace cling {
   private:
     std::unique_ptr<clang::Parser> m_Parser;
     Interpreter* m_Interpreter; // we do not own.
-    const clang::Type* m_StringTy[kNumCachedStrings] = {};
+    const clang::Type* m_StringTy[kNumCachedStrings];
```
Axel-Naumann: I really don't want to leave this uninitialized. Too fragile; this is not C.

lib/Interpreter/LookupHelper.cpp
```diff
@@ -1910,6 +1911,10 @@ namespace cling {
     assert(Type && "Type cannot be null");
     const Transaction*& Cache = m_Interpreter->getStdStringTransaction();
     if (!Cache || !m_StringTy[kStdString]) {
+      // getStringType can be called multiple times with Cache being null, and
+      // the local cache should be discarded when that occurs.
+      if (!Cache) ::memset(m_StringTy, 0, sizeof(m_StringTy));
```
Axel-Naumann: Can we leave this line (and this comment) but rephrase it as
```
m_StringTy = {};
```
That should be fine, even if leaving the in-class initializer, or am I overlooking something?

marsupial:
You can adjust however you like, but I tried m_StringTy = {}, and clang complained it was not a legal statement.  I don't have any particular care as to how the m_StringTy data gets back to zero, just that it does when Cache is null.

I'm a little against the in class init, just because it is a non trial amount of data that doesn't need to be initialized if it is never used, but understand that -appearing- safer is not a bad thing.

Up to you.

Axel-Naumann:
Thanks! And indeed, that'll need to become a `std::array`. I will take care of it and then close this (unless you have already provided an update PR when I get up tomorrow ;-)

marsupial:
Probably not me, I'll be busy dancing with conflicts from deb1fcb !

Axel-Naumann:
Thanks for explaining! Solved differently, see aae2870f62177c7ad82143dabbb97b02c30f80b2

PR #157, opened 2017-06-23
Fix crash and propogate/report errors better.

marsupial:
ValueExtractionSynthesizer is currently doing -error checking- via asserts which is meaningless in Release and will just let a crash happen. These changes make it more defensive and report error's in a nicer way.


Axel-Naumann:
Merged through ROOT (fixing merge conflicts). Thanks for the contribution, @marsupial !

marsupial:
Oldie but goodie !

PR #159, opened 2017-06-23
Fix crash when dumping an un-evaluated cling::Value

marsupial:
This fixes the following code which will currently crash:
```
cling::Value V;
V.dump();
```

The first commit while not totally related was for conflict resolution against 27aea4b and reports the error through clang's diagnostics.


Axel-Naumann:
The value printing has changed too much. While I'd love to incorporate these changes they don't apply as is, and porting them takes more effort than I can invest here. Thus closing - but I'd be happy if anyone can re-implement these changes for the current cling!

PR #160, opened 2017-06-23
Fix crash when undoing past the first print transaction.

marsupial:
The use of a static makes this crash, but also makes it so a second interpreter in a session would never work.  It also merges all Transactions generated by printing, so a statement become atomic for undo rather than a series unrelated Transactions (which abstracts the implementation away from the user a bit better).

Was
```
// Startup transactions
<cling::Transaction* 0x7fae5385a800 isEmpty=0 isCommitted=1> 
`   <cling::Transaction* 0x7fae538ec000 isEmpty=0 isCommitted=1> 
// Trigger {} Tr{}
<cling::Transaction* 0x7fae53858a00 isEmpty=0 isCommitted=1> 
<cling::Transaction* 0x7fae54154200 isEmpty=0 isCommitted=1> 
<cling::Transaction* 0x7fae53939000 isEmpty=0 isCommitted=1> 
<cling::Transaction* 0x7fae550e7400 isEmpty=0 isCommitted=1> 

```
Becomes
```
// Startup transactions
<cling::Transaction* 0x7f9fd104ac00 isEmpty=0 isCommitted=1> 
`   <cling::Transaction* 0x7f9fd10d1400 isEmpty=0 isCommitted=1> 
// Trigger {} Tr{}
<cling::Transaction* 0x7f9fd11c0200 isEmpty=0 isCommitted=1> 
`   <cling::Transaction* 0x7f9fd2028000 isEmpty=0 isCommitted=1> 
`   <cling::Transaction* 0x7f9fd1481200 isEmpty=0 isCommitted=1> 
`   <cling::Transaction* 0x7f9fd219c000 isEmpty=0 isCommitted=1> 
```

This includes and is dependent on https://github.com/root-project/cling/pull/157 & https://github.com/root-project/cling/pull/159.


lib/Interpreter/Value.cpp
```diff
@@ -231,9 +231,15 @@ namespace cling {
   namespace valuePrinterInternal {
     std::string printTypeInternal(const Value& V);
     std::string printValueInternal(const Value& V);
+    extern const char* const kUndefined;
   } // end namespace valuePrinterInternal
 
   void Value::print(llvm::raw_ostream& Out, bool Escape) const {
+    if (!m_Interpreter) {
+      Out << valuePrinterInternal::kUndefined << "\n";
```
Axel-Naumann: I would prefer if the different "failure" cases had different output strings, instead of all mapping to `kUndefined`, e.g. here: `Out << "<ERROR: missing Interpreter!>\n"`.

This will also remove the awkwardness of having to redeclare `kUndefined` in different, seemingly unrelated places - runtime and compiled code, Value.cpp and ValuePrinter.cpp.

lib/Interpreter/ValueExtractionSynthesizer.cpp
```diff
@@ -167,13 +167,19 @@ namespace cling {
...
+          if (RS) {
+            if (ImplicitCastExpr* VoidCast
+                = dyn_cast<ImplicitCastExpr>(RS->getRetValue()))
+              VoidCast->setSubExpr(SVRInit);
+          }
+          else
+            **I = SVRInit;
+        }
+        else {
+          // FIXME: Is this leaking anything (lastExpr?)
```
Axel-Naumann: No, not leaking - any created `Expr` is created in the ASTContext's arena.

lib/Interpreter/ValueExtractionSynthesizer.cpp
```diff
@@ -395,34 +408,54 @@ namespace {
     return Call.get();
   }
 
-  void ValueExtractionSynthesizer::FindAndCacheRuntimeDecls() {
+  static bool VSError(clang::Sema* Sema, clang::Expr* E, llvm::StringRef Err) {
```
Axel-Naumann: Can you teach the `Lookup` routines to not suppress the diagnostics if told? I.e. use the `DiagSetting` technique from `LookupHelper` also for `Lookup` and then pass `WithDiagnostics` here?

lib/Interpreter/ValueExtractionSynthesizer.cpp
```diff
@@ -395,34 +408,54 @@ namespace {
...
     }
     LookupResult R(*m_Sema, &m_Context->Idents.get("setValueNoAlloc"),
                    SourceLocation(), Sema::LookupOrdinaryName,
                    Sema::ForRedeclaration);
 
     m_Sema->LookupQualifiedName(R, NSD);
-    assert(!R.empty()
-           && "Cannot find cling::runtime::internal::setValueNoAlloc");
+    if (R.empty())
+      return VSError(m_Sema, E, "cling::runtime::internal::setValueNoAlloc");
```
Axel-Naumann: Do we really need an extra diagnostic here? Doesn't `LookupQualifiedName()` complain already?

lib/Interpreter/ValueExtractionSynthesizer.cpp
```diff
@@ -395,34 +408,54 @@ namespace {
...
 
     R.clear();
     R.setLookupName(&m_Context->Idents.get("setValueWithAlloc"));
     m_Sema->LookupQualifiedName(R, NSD);
-    assert(!R.empty()
-           && "Cannot find cling::runtime::internal::setValueWithAlloc");
-    m_UnresolvedWithAlloc
-      = m_Sema->BuildDeclarationNameExpr(CSS, R, /*ADL*/ false).get();
+    if (R.empty())
+      return VSError(m_Sema, E, "cling::runtime::internal::setValueWithAlloc");
```
Axel-Naumann: ditto

lib/Interpreter/ValueExtractionSynthesizer.cpp
```diff
@@ -433,10 +466,15 @@ namespace {
     // parent interpreter, but it will fail, because this is a template function.
     // Once the import of template functions becomes supported by clang,
     // this check can be de-activated.
-    if (!m_isChildInterpreter)
-      assert(!R.empty() && "Cannot find cling::runtime::internal::copyArray");
-    m_UnresolvedCopyArray
-      = m_Sema->BuildDeclarationNameExpr(CSS, R, /*ADL*/ false).get();
+    if (!m_isChildInterpreter && R.empty())
+      return VSError(m_Sema, E, "cling::runtime::internal::copyArray");
```
Axel-Naumann: `LookupQualifiedName()` already diagnoses; just `return false` here,

lib/Interpreter/ValueExtractionSynthesizer.cpp
```diff
@@ -433,10 +466,15 @@ namespace {
...
+
+    m_UnresolvedCopyArray = m_Sema->BuildDeclarationNameExpr(CSS, R, ADL).get();
+    if (!m_UnresolvedCopyArray)
+      return VSError(m_Sema, E, "cling::runtime::internal::copyArray");
```
Axel-Naumann: No need for an extra diag; `BuildDeclarationNameExpr()` already diagnoses failures. Please simply `return false` here.

lib/Interpreter/ValueExtractionSynthesizer.cpp
```diff
@@ -395,34 +408,54 @@ namespace {
...
-    assert(!R.empty()
-           && "Cannot find cling::runtime::internal::setValueWithAlloc");
-    m_UnresolvedWithAlloc
-      = m_Sema->BuildDeclarationNameExpr(CSS, R, /*ADL*/ false).get();
+    if (R.empty())
+      return VSError(m_Sema, E, "cling::runtime::internal::setValueWithAlloc");
+
+    m_UnresolvedWithAlloc = m_Sema->BuildDeclarationNameExpr(CSS, R, ADL).get();
+    if (!m_UnresolvedWithAlloc)
+      return VSError(m_Sema, E, "cling::runtime::internal::setValueWithAlloc");
```
Axel-Naumann: No need for an extra diag; `BuildDeclarationNameExpr()` already diagnoses failures. Please simply `return false` here.

lib/Interpreter/ValueExtractionSynthesizer.h
```diff
@@ -90,7 +90,7 @@ namespace cling {
 
     // Find and cache cling::runtime::gCling, setValueNoAlloc,
     // setValueWithAlloc on first request.
-    void FindAndCacheRuntimeDecls();
+    bool FindAndCacheRuntimeDecls(clang::Expr*);
```
Axel-Naumann: Please add doc for the `Expr` parameter, especially as it's not obvious without reading the implementation.

lib/Interpreter/ValueExtractionSynthesizer.cpp
```diff
@@ -12,8 +12,8 @@
 #include "cling/Interpreter/Interpreter.h"
 #include "cling/Interpreter/Transaction.h"
 #include "cling/Interpreter/Value.h"
-
 #include "cling/Utils/AST.h"
+#include "cling/Utils/Output.h"
```
Axel-Naumann: What do we need this header for? (I might have missed it, or it might have been from an earlier revision of your code that didn't use the CustomDiag mechanism?)

lib/Interpreter/ValuePrinter.cpp
```diff
@@ -862,25 +847,55 @@ namespace cling {
         strm << printUnpackedClingValue(*value);
       }
       strm << "]";
-    } else
-      strm << "<<<invalid>>> " << printAddress(value, '@');
-
+    } else {
+      strm << valuePrinterInternal::kUndefined
```
Axel-Naumann: Again - I prefer to keep the distinction of error and invalid `Value`. I.e. I prefer to keep the original code here.

include/cling/Interpreter/Interpreter.h
```diff
@@ -102,6 +102,19 @@ namespace cling {
       ~StateDebuggerRAII();
     };
 
+    ///\brief Merge all transactions between construction and destruction
+    /// Use Prev to merge the Transactions with the last Transaction (true)
+    /// or the first Transaction created after instantiation (false).
+    class TransactionMerge {
```
Axel-Naumann: We traditionally use nouns for classes. Maybe `ScopedTransactionMerger`? (Well, we traditionally would have called that `TransactionMergerRAII`, choose whatever you prefer.)

lib/Interpreter/IncrementalParser.cpp
```diff
@@ -280,6 +280,55 @@ namespace cling {
...
+    }
+    if (merge.size()>1) {
+      llvm::SmallVector<Transaction*, 8>::reverse_iterator itr = merge.rbegin();
+      Transaction *parent = *itr;
+      ++itr;
+      for (llvm::SmallVector<Transaction*, 8>::reverse_iterator end = merge.rend();
+           itr != end; ++itr) {
+        Transaction *cur = *itr;
+        // Try to keep everything current
+        if (m_Consumer->getTransaction()==cur)
```
Axel-Naumann: Please add spaces around relops (several occurrences).

lib/Interpreter/IncrementalParser.cpp
```diff
@@ -280,6 +280,55 @@ namespace cling {
     return m_Consumer->getTransaction();
   }
 
+  const Transaction*
+  IncrementalParser::mergeTransactionsAfter(const Transaction *T, bool Prev) {
```
Axel-Naumann: Isn't `T` meant to be a top-most transaction? If so, could you assert?

lib/Interpreter/IncrementalParser.cpp
```diff
@@ -280,6 +280,55 @@ namespace cling {
...
+    for (std::deque<Transaction*>::reverse_iterator itr = m_Transactions.rbegin(),
+            end  = m_Transactions.rend(); itr != end; ++itr) {
+      Transaction *cur = *itr;
+      merge.push_back(cur);
+      if (cur==T) {
+        if (!Prev)
+          merge.pop_back();
+        break;
+      }
+    }
```
Axel-Naumann: If `T` is not a top-most transaction, or unloaded, or rolled back, then it will not be part of `m_Transactions`. In that case, your code currently merges *all* transactions into the first (if I read your code correctly).

You should at least assert that you've seen `T` in `m_Transactions`. I'd even simply use `find()` instead of building the auxiliary `SmallVector merge`.

lib/Interpreter/IncrementalParser.cpp
```diff
@@ -280,6 +280,55 @@ namespace cling {
...
+      llvm::SmallVector<Transaction*, 8>::reverse_iterator itr = merge.rbegin();
+      Transaction *parent = *itr;
+      ++itr;
+      for (llvm::SmallVector<Transaction*, 8>::reverse_iterator end = merge.rend();
+           itr != end; ++itr) {
+        Transaction *cur = *itr;
+        // Try to keep everything current
+        if (m_Consumer->getTransaction()==cur)
+          m_Consumer->setTransaction(parent);
+        if (cur==parent->getNext()) // Make sure parent->next isn't a child
```
Axel-Naumann: `patent->getNext()` *must* point to the first transaction to be merged. You should probably assert that, and then set it to `nullptr`, because after `mergeTransactionsAfter()` is done the parent aught to not have a subsequent transaction anymore.

lib/Interpreter/IncrementalParser.cpp
```diff
@@ -280,6 +280,55 @@ namespace cling {
...
+          parent->setNext(const_cast<Transaction*>(cur->getNext()));
+        while (cur->getParent())   // Keep parents in place
```
Axel-Naumann: Are you saying `m_Transaction" contains nested transactions? IIRC that would be a bug... I.e. this should probably be an assert.

lib/Interpreter/IncrementalParser.cpp
```diff
@@ -280,6 +280,55 @@ namespace cling {
     return m_Consumer->getTransaction();
   }
 
+  const Transaction*
```
Axel-Naumann: The return type seems unused. Can we make it `void` until we need something?

test/ErrorRecovery/ABI/cling/Interpreter/RuntimePrintValue.h
```diff
@@ -0,0 +1,2 @@
+
+ERROR;
```
Axel-Naumann: :-) Could you maybe add a comment that this is meant to simulate a parsing error of cling's original RuntimePrintValue.h?

lib/Interpreter/Value.cpp
marsupial: That make parsing cling's output and detecting invalid results, way more difficult.  There were previously 3 different result which were all invalid, all marked in different ways.

I think a lot of what you're after is achieved by properly using diagnostics to actually report the problem.

lib/Interpreter/IncrementalParser.h
```diff
@@ -175,6 +175,14 @@ namespace cling {
       return m_Transactions.back();
     }
 
+    ///\brief Merge transactions after the one given.
+    ///
+    ///\param[in] T - transactions after which to merge
+    ///\param[in] prev - merge into transaction prior to T or T itself
+    ///\returns the parent transaction of the merge.
+    ///
+    const Transaction* mergeTransactionsAfter(const Transaction *T, bool prev);
```
marsupial: To me that couples the interfaces in a way that is undesirable. That is: to use the function (or wrapper class) one would also need to know an implementation detail of Transaction. A simple bool avoids this.

lib/Interpreter/Interpreter.cpp
```diff
@@ -141,6 +141,16 @@ namespace cling {
     }
   }
 
+  Interpreter::TransactionMerge::TransactionMerge(Interpreter *interp,
+                                                  bool prev) :
```
marsupial: ditto for `prev`

lib/Interpreter/Value.cpp
marsupial: And the fact it needs to be declared in Value.cpp is not a unique problem i.e. `printTypeInternal` and `printValueInternal`.

lib/Interpreter/ValueExtractionSynthesizer.cpp
marsupial: Is this more descriptive:
```
// FIXME: Do this atomically or something so that AST context will not
// contain Expr(s) that are unused for the rest of it's life.
```

lib/Interpreter/ValueExtractionSynthesizer.cpp
marsupial: I have a separate PR that I think adresses what you want...

lib/Interpreter/ValueExtractionSynthesizer.cpp
marsupial: I don't think that is the case, it's used to lookup things that may not exist heavily in clang.
**test/ErrorRecovery/ABI.C** would seem to verify that it doesn't.

lib/Interpreter/IncrementalParser.h
Axel-Naumann: I do not understand your argument.

lib/Interpreter/ValueExtractionSynthesizer.cpp
marsupial: I know BuildDeclarationNameExpr can suppress diagnostics in some cases.

lib/Interpreter/Value.cpp
Axel-Naumann: > That make parsing cling's output and detecting invalid results, way more difficult.

Correct, but don't care: I'm after end user usability, not test simplicity. And yes, diagnostics help, but I don't see why they are a motivation to shove all differences in error versus invalid `Value` under the same rug?

lib/Interpreter/ValueExtractionSynthesizer.cpp
Axel-Naumann: For me: yes, absolutely! Now I understand :-)

lib/Interpreter/ValueExtractionSynthesizer.cpp
Axel-Naumann: Well - but why introduce VSError here then if a simple "don't suppress diagnostics" suffices? With the change proposed here, we first disable the default diagnostics to then emit our custom ones.

If this is a sequencing / merge conflict etc issue then I am not against merging this first, as long as you promise by the honor of all kangaroos, wallabies, koalas, possums, opossums, wombats and Tasmanian devils to hand that PR in at some later point! :-)

lib/Interpreter/IncrementalParser.cpp
marsupial: No it checks `merge.size()>1` which I'll change to `!merge.empty()`.
The auxiliary vector is used for keeping and inspecting the existing parent structure.

lib/Interpreter/ValueExtractionSynthesizer.cpp
Axel-Naumann: Thanks, ignore my comment.

lib/Interpreter/ValueExtractionSynthesizer.cpp
Axel-Naumann: Okay, thanks, then please ignore my comment.

lib/Interpreter/ValueExtractionSynthesizer.cpp
Axel-Naumann: I'm wrong, please ignore.

lib/Interpreter/ValueExtractionSynthesizer.cpp
Axel-Naumann: Ignore my comment.

lib/Interpreter/ValueExtractionSynthesizer.cpp
Axel-Naumann: Ignore.

lib/Interpreter/ValueExtractionSynthesizer.cpp
marsupial: Yes, cruft from:
> Fix crash in ValueExtractionSynthesizer::FindAndCacheRuntimeDecls() and report errors.

Axel-Naumann: Sorry, I don't understand. Do you need this include, and if so, which declarations are used?

lib/Interpreter/IncrementalParser.cpp
marsupial: I tried to make it more apparent, but considering the case:
```
<cling::Transaction* A isEmpty=0 isCommitted=1> 
<cling::Transaction* B isEmpty=0 isCommitted=1> 
<cling::Transaction* C isEmpty=0 isCommitted=1> 
<cling::Transaction* D isEmpty=0 isCommitted=1> 
<cling::Transaction* E isEmpty=0 isCommitted=1> 

Interp->mergeTransactionsAfter(A)
```
As each lower Transaction is put into Parent(A), Parent should absorb whatever **next** the Child had, until finally reaching E where it's **next** is _nullptr_.

lib/Interpreter/IncrementalParser.cpp
marsupial: To be clearer "keeping and inspecting the existing parent structure," means that the Transactions must be cached as their real pointer values first, as iterators will be invalidated when fiddling with the nesting structure.

lib/Interpreter/IncrementalParser.h
marsupial: 1. To use this functionality I don't think a requirement should be an `#include "Transaction.h"` and the caller knowing how Transactions are implemented.

2. The difference is more complicated than T or T->getNext().  As **Previous** get's the Transaction before what was passed.

lib/Interpreter/Value.cpp
marsupial: Because the error is already reported, and what "invalid" means is really not clear.  How is it invalid? What aspect.
Undefined represents a value in many context's whose content should not be relied upon to consistently be what was shown.

It's not test simplicity, it using and parsing the output (especially in a shell needs to trap two or three distinct cases that could grow even more as people invent new problems to represent when they should be reporting it properly):

VAL=$(cling blah blah)
```
if [[ $VAL != *"<undefined>"* ]]; then
 ...
fi
```
vs.

```
if [[ $VAL != *"<<<invalid>>>"* ]] || [[ $VAL != *"ERROR in cling"* ]]
 ...
fi
```

And the above example doesn't even take into account the bug that was fixed which would then have a totally new string representation, or if it didn't would suffer the same problem of Value.cpp needing a declaration to share.

lib/Interpreter/ValueExtractionSynthesizer.cpp
marsupial: 1. I don't want any errors suppressed. Even if setting LookupResult mostly catches the errors, I know BuildDeclarationNameExpr will in some cases suppress them anyway.
2. LookupResult is being passed an invalid source location which makes any diagnostics generated poor.
3. Yes there is a conflict, which will alleviate 2, but 1 will still stand.

lib/Interpreter/ValueExtractionSynthesizer.cpp
marsupial: You were right to note it should be removed. Just saying it does come into existence in an early commit in the PR, but was not taken out when it should have been later.

lib/Interpreter/Value.cpp
Axel-Naumann: I want an empty `Value` to have a unique string representation that is different from error cases. I do not care whether you rename that to e.g. `"<empty>"` or leave it at `"<<<invalid>>>"`. But the output must be different from the output due to an infrastructure issue.

lib/Interpreter/IncrementalParser.h
Axel-Naumann: Okay, we don't agree. Let's try from a different angle: do you actually need that argument? All calls seem to pass `true`. Can we just remove the argument and rename this to `mergeTransactionsInfoPrevious(const Transaction *T)`?

lib/Interpreter/Interpreter.cpp
```diff
@@ -1193,20 +1203,40 @@ namespace cling {
...
+               && "Value printer failed to load after success?");
+        // As the stack is unwinded from recursive calls to EvaluateInternal
+        // we merge the subsequent transactions into the current one.
+        // Then m_CachedTrns[kPrintValue] is marked as this transaction as it
+        // is the first known Transaction that caused the printer to be loaded.
+        m_IncrParser->mergeTransactionsAfter(lastT, true);
+        m_CachedTrns[kPrintValueTransaction] = lastT;
+      }
+
+      if ( rslt < kExeFirstError) {
```
Axel-Naumann: Not yet done...

lib/Interpreter/ValueExtractionSynthesizer.cpp
Axel-Naumann: I buy that case 1). Thanks. Consider this settled.

lib/Interpreter/IncrementalParser.cpp
Axel-Naumann: I don't think your comment addresses my concern:

If I call `IncrementalParser::mergeTransactionsAfter((Transaction*)0x123, prev)` it will merge all transactions into the first. Or crash. Probably depending on the value of `prev`. I don't see how `!merge.empty()` prevents that, as `merge` will contain *all* transactions.

lib/Interpreter/IncrementalParser.cpp
Axel-Naumann: Let me try again. In the loop you do
```
   Parent->setNext(const_cast<Transaction*>(CurChild->getNext()));
```
In the end you do
```
assert(Parent->getNext() == nullptr && "Parent still has next");
```
Why do you update `Parent->getName()` during the loop?

lib/Interpreter/IncrementalParser.cpp
marsupial: Are you looking at the latest? The "Show/Hide outdated" above indicates not.

lib/Interpreter/ValueExtractionSynthesizer.h
marsupial: This will conflict with something else, if the conflicting PR is denied I can handle it then.

lib/Interpreter/Value.cpp
marsupial: That was changed back in the last commit.
If <<<invalid>>> is the only thing necessary, can the `return "ERROR in cling's callPrintValue(): missing value string.";` be put back to `<undefined>`, trapping 3 cases seems a bit much for no new information...

Axel-Naumann: Thanks! Do you agree that this is awkward:
```
 V // CHECK: (cling::Value &) <<<invalid>>> @0x{{.*}}
  		  
 +V.dump(); // CHECK-NEXT: <undefined>
```

For all but this case, `V.dum()` is the value part of the value printer. Can we make this:
```
 V // CHECK: (cling::Value &) <<<invalid>>> @0x{{.*}}
  		  
 +V.dump(); // CHECK-NEXT: <<<invalid>>>
```

Regarding
> can the `return "ERROR in cling's callPrintValue(): missing value string.";` be put back to `<undefined>`

That's fine with me. With that change, `<undefined>` would be printed whenever there is an infrastructure issue - that's perfect. Thank you!

lib/Interpreter/Interpreter.cpp
```diff
@@ -1193,20 +1203,40 @@ namespace cling {
...
+      // value printer and must be recorded as such.
+      if ( prntT != m_CachedTrns[kPrintValueTransaction] ) {
+        assert(m_CachedTrns[kPrintValueTransaction] != nullptr
+               && "Value printer failed to load after success?");
+        // As the stack is unwinded from recursive calls to EvaluateInternal
+        // we merge the subsequent transactions into the current one.
+        // Then m_CachedTrns[kPrintValue] is marked as this transaction as it
+        // is the first known Transaction that caused the printer to be loaded.
+        m_IncrParser->mergeTransactionsAfter(lastT, true);
+        m_CachedTrns[kPrintValueTransaction] = lastT;
```
Axel-Naumann: Done, thanks.

lib/Interpreter/ValueExtractionSynthesizer.h
Axel-Naumann: Okay.

lib/Interpreter/IncrementalParser.cpp
Axel-Naumann: I am looking at the latest. I expect to see
```
assert(!T->isNestedTransaction() && "Stem of transaction merge must be a topmost transaction.");
```
(or a nicer worded version of the same) at the top of the function.

lib/Interpreter/IncrementalParser.cpp
marsupial: Because as it loops it is stealing next from whatever it is absorbing.
```
<cling::Transaction* A isEmpty=0 isCommitted=1> 
<cling::Transaction* B isEmpty=0 isCommitted=1> 
<cling::Transaction* C isEmpty=0 isCommitted=1> 
<cling::Transaction* D isEmpty=0 isCommitted=1> 
<cling::Transaction* E isEmpty=0 isCommitted=1> 

A.Next = B, B.Next = C, C.Next = D, D.Next = NULL

Interp->mergeTransactionsAfter(A):

Make B a child: A.Next = B.Next, B.Next = 0
Make C a child: A.Next = C.Next, C.Next = 0
Make D a child: A.Next = D.Next, D.Next = 0
Make D a child: A.Next = E.Next, E.Next = 0
```

If an exception is thrown before all children could be nested, then it should hopefully be in  a state that is valid.

lib/Interpreter/IncrementalParser.cpp
marsupial: Done, the assert takes place after all the runtime errors that can be handled properly.

PR #161, opened 2017-06-27
Remove extra semicolon.

marsupial:


Axel-Naumann:
Merged, thanks!

Axel-Naumann:
Btw - shouldn't we have a test for this? As in, shouldn't one of the *existing* tests have complained about this?

PR #162, opened 2017-06-27
Allow GNU extensions when requested.

marsupial:
& remove outdated comments.

### DIFF
--- a/include/cling/Interpreter/InvocationOptions.h
+++ b/include/cling/Interpreter/InvocationOptions.h
@@ -13,6 +13,10 @@
 #include <string>
 #include <vector>
 
+namespace clang {
+  class LangOptions;
+};
+
 namespace cling {
 
   ///\brief Class that stores options that are used by both cling and
@@ -42,7 +46,7 @@ namespace cling {
     /// or './cling -x c') that this shouldn't be done.  This will return false
     /// in those cases.
     ///
-    bool DefaultLanguage() const { return !Language && !StdVersion; }
+    bool DefaultLanguage(const clang::LangOptions&) const;
 
     unsigned Language : 1;
     unsigned ResourceDir : 1;
--- a/include/cling/Interpreter/InvocationOptions.h
+++ b/include/cling/Interpreter/InvocationOptions.h
@@ -36,6 +36,14 @@ namespace cling {
     void Parse(int argc, const char* const argv[],
                std::vector<std::string>* Inputs = nullptr);
 
+    ///\brief By default clang will try to set up an Interpreter with features
+    /// that match how it was compiled.  There are cases; however, where the
+    /// client is asking for something so specific (i.e './cling -std=gnu++11'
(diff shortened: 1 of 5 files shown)


include/cling/Interpreter/InvocationOptions.h
```diff
@@ -36,6 +36,11 @@ namespace cling {
     void Parse(int argc, const char* const argv[],
                std::vector<std::string>* Inputs = nullptr);
 
+    ///\brief Whether the options given will not override anything for the
```
Axel-Naumann: The options given where? Who will override? What's "for the Interpreter's default language options"? What does "false" mean given "whether... will not"? I.e. could you expand yet simplify this doc a bit? :-)

lib/Interpreter/CIFactory.cpp
```diff
@@ -628,20 +632,21 @@ static void stringifyPreprocSetting(PreprocessorOptions& PPOpts,
...
     SetPreprocessorFromBinary(PPOpts);
 
     PPOpts.addMacroDef("__CLING__");
-    if (CI->getLangOpts().CPlusPlus11 == 1) {
-      // http://llvm.org/bugs/show_bug.cgi?id=13530
+    if (CI->getLangOpts().CPlusPlus11 == 1)
       PPOpts.addMacroDef("__CLING__CXX11");
-    }
+    if (CI->getLangOpts().CPlusPlus14 == 1)
+      PPOpts.addMacroDef("__CLING__CXX14");
```
Axel-Naumann: ..and why not 17?

test/Driver/Gnu.C
```diff
@@ -0,0 +1,25 @@
+//------------------------------------------------------------------------------
+// CLING - the C++ LLVM-based InterpreterG :)
+//
+// This file is dual-licensed: you can choose to license it under the University
+// of Illinois Open Source License or the GNU Lesser General Public License. See
+// LICENSE.TXT for details.
+//------------------------------------------------------------------------------
+
+// RUN: cat %s | %cling -D__STRICT_ANSI__ -std=gnu++11 -Xclang -verify 2>&1 | FileCheck %s
+// RUN: cat %s | %cling -std=gnu99 -x c -Xclang -verify 2>&1 | FileCheck %s
```
Axel-Naumann: Could you add a test for -std=c++1z? Reason is that cling cannot be compiled with both 1z and 11, so this will check that we don't override the -std by whatever cling detects at build time. (Yes, one could argue gnu99 could play that role, but that has a language change included; I'd prefer to test these two items separately.)

lib/Interpreter/CIFactory.cpp
marsupial: One day, I didn't think wise to mark __CLING__CXX17 until C++17 was ratified.

marsupial:
Done excluding `__CLING__CXX17`.

Axel-Naumann:
Thanks, merged!

PR #163, opened 2017-06-29
Fix CxxStdCompiledWith for Visual Studio, which requires C++14.

marsupial:
An infinite loop, some discarded input, and a poor decision by Microsoft.

### DIFF
--- a/lib/Interpreter/CIFactory.cpp
+++ b/lib/Interpreter/CIFactory.cpp
@@ -60,9 +60,9 @@ namespace {
     // I would claim that the check should be relaxed to:
 #if __cplusplus > 201402L
     return 17;
-#elif __cplusplus > 201103L || (defined(LLVM_ON_WIN32) && _MSC_VER == 1910)
+#elif __cplusplus > 201103L || (defined(LLVM_ON_WIN32) && _MSC_VER >= 1900)
     return 14;
-#elif __cplusplus >= 201103L || (defined(LLVM_ON_WIN32) && _MSC_VER == 1900)
+#elif __cplusplus
     return 11;
 #else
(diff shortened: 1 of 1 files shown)


marsupial:
Ok, if you want to change to use **_MSVC_LANG** just make sure that the default cases compiling cling will not error out as it was doing.

bellenot:
Don't worry, your patch is good enough for me

marsupial:
Thanks.
Also worth noting that in the travels of this I noticed that if a file is using Windows line endings `\r\n` it is being parsed as two distinct newlines.

marsupial:
Looks like C++14 is a requirement for MSVC headers, so I changed it to:
`#elif __cplusplus > 201103L || (defined(LLVM_ON_WIN32) && _MSC_VER >= 1900)`

Axel-Naumann:
Argh so https://github.com/root-project/root/pull/711 went out of sync... I'm reverting it and then re-applying this.

Axel-Naumann:
No wait now you confused me. I have applied https://github.com/root-project/root/pull/711 but this PR now morphed into something completely unrelated, right? Should I apply this PR, too? Or revert https://github.com/root-project/root/pull/711 and only apply this? Let me know, please - ideally within the next couple of days, I'll be traveling afterwards...

marsupial:
Yes seems it was cloned, but github doesn't update.
I pushed the simplest fix that would apply to master here as everything else in the PR was the same.

bellenot:
I committed this (more or less) in ROOT. So I close this PR

Axel-Naumann:
Thanks for your help, @marsupial @bellenot !

PR #164, opened 2017-06-29
Fix Windows exception handling for newer LLVM interfaces.

marsupial:
Block it’s usage around CLING_WIN_SEH_EXCEPTIONS macro.

### DIFF
--- a/lib/Utils/PlatformWin.cpp
+++ b/lib/Utils/PlatformWin.cpp
@@ -701,9 +701,13 @@ std::string Demangle(const std::string& Symbol) {
   return af.Str ? std::string(af.Str) : std::string();
 }
 
+#ifdef CLING_WIN_SEH_EXCEPTIONS
 namespace windows {
 // Use less memory and store the function ranges to watch as a mapping
 // between of BaseAddr to Ranges watched.
+//
+// FIXME: Now that having sibling Interpreters is becoming possible, this
+// data should be held per Interpeter (possibly only by the top-most parent).
 //
 typedef std::vector<std::pair<DWORD, DWORD>> ImageRanges;
 typedef std::map<uintptr_t, ImageRanges> ImageBaseMap;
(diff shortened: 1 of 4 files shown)


lib/Utils/PlatformWin.cpp
```diff
@@ -735,49 +739,64 @@ static void MergeRanges(ImageRanges &Ranges) {
...
   }
 
-  ::RtlAddFunctionTable(RFunc, N, BaseAddr);
+  if (!Block)
+    MergeRanges(Ranges); // Initial sort and merge
 }
 
-void DeRegisterEHFrames(uint8_t* Addr, size_t Size) {
-  assert(getImageBaseMap().find(DWORD64(Addr)) != getImageBaseMap().end());
+void DeRegisterEHFrames(uintptr_t ImgBase, EHFrameInfos Frames) {
```
bellenot: Should be: `void DeRegisterEHFrames(uintptr_t BaseAddr, const EHFrameInfos& Frames) {`

marsupial: Indeed....How'd that happen?

bellenot:
BTW, why blocking it’s usage around CLING_WIN_SEH_EXCEPTIONS macro?
Without defining CLING_WIN_SEH_EXCEPTIONS, it gives the following compilation error:
```
"C:\Users\bellenot\build\release\ALL_BUILD.vcxproj" (default target) (1) ->
"C:\Users\bellenot\build\release\tools\cling\tools\driver\cling.vcxproj" (default target) (235) ->
"C:\Users\bellenot\build\release\tools\cling\lib\Interpreter\clingInterpreter.vcxproj" (default target) (236) ->
"C:\Users\bellenot\build\release\tools\cling\lib\Interpreter\obj.clingInterpreter.vcxproj" (default target) (237) ->
(ClCompile target) ->
  C:\Users\bellenot\git\original\tools\cling\lib\Interpreter\Interpreter.cpp(467): error C2039: 'ClingRaiseSEHException
': is not a member of 'cling::utils::platform' [C:\Users\bellenot\build\release\tools\cling\lib\Interpreter\obj.clingIn
terpreter.vcxproj]
  C:\Users\bellenot\git\original\tools\cling\lib\Interpreter\Interpreter.cpp(467): error C2065: 'ClingRaiseSEHException
': undeclared identifier [C:\Users\bellenot\build\release\tools\cling\lib\Interpreter\obj.clingInterpreter.vcxproj]
  C:\Users\bellenot\git\original\tools\cling\lib\Interpreter\Interpreter.cpp(467): error C2664: 'bool cling::Incrementa
lExecutor::addSymbol(const char *,void *,bool)': cannot convert argument 2 from 'bool' to 'void *' [C:\Users\bellenot\b
uild\release\tools\cling\lib\Interpreter\obj.clingInterpreter.vcxproj]

    3 Warning(s)
    3 Error(s)
```
Then I just commented them out, or replaced with LLVM_ON_WIN32 macro, and it works...

marsupial:
Fixed. Macro is there because:
1. I've tested this on a JIT using COFF and have no idea how/if it functions when you use ELF which master still does (and seemingly will for a while).
2. There were previous complaints that the code is 64-bit only. You can figure out how you want to enable/disable it in relation to that and the above.

bellenot:
OK, I see, thanks for the info. It makes perfect sense

Axel-Naumann:
Merged, thanks!

PR #166, opened 2017-07-04
Explicitly tell clang what std version to use.

marsupial:
Clang is deducing the language standard on it's own, then uses this value durring initialization, and cling then attempts to change it afterward.

Avoid this by telling clang what is desired early on.
And don't have Windows use GNU keywords by default.


Axel-Naumann:
Excellent, thanks a lot! Merged!

PR #168, opened 2017-07-05
Call CompilerOptions:: DefaultLanguage before setting -std=

marsupial:
Should fix issue over eagerly setting -std= when compiling a precompiled header.


Axel-Naumann:
Thanks! Moved to https://github.com/root-project/root/issues/746

PR #169, opened 2017-07-05
Support thread_local data.

marsupial:
thread_local is part of the standard, so somewhat necessary to be compliant.
Windows uses thread local storage for its implementation of C++ static semantics.

Currently cling will not complain about thread_local, but due to the way it is implemented on OS X and Linux will likely crash from any extensive usage.  By using emulated TLS the thread local data will be dynamically generated so that using them will not overwrite the owning process's thread local data stored in a section of the binary.


Axel-Naumann:
Thanks! Moved to https://github.com/root-project/root/issues/745.

PR #170, opened 2017-07-05
Access parent from a child interpreter, and const correct gCling.

marsupial:
Also makes RuntimeUniverse.h includable by languages other than C++, simplifying the code injected at startup.

By storing the parent-child relationship in the Interpreter gCling can be at the link level rather than writing it's value explicitly via casting.

This solves a mangling issue (at least on OS X), so that gCling can properly be made a const pointer.

The ultimate goal of this is to move the pointer **gCling** to a reference of **thisCling** for all Interpreters including children, who can then access their parent interpreter via **thisCling.parent()** or the top-most instance via **thisCling.ancestor()**.


marsupial:
This also opens a pathway to move what is currently static variables in CIFactory.cpp as a property of the root-level Interpreter and enforce that all children have the same include paths as a parent.

include/cling/Utils/AST.h
```diff
@@ -50,6 +50,18 @@ namespace utils {
...
+    ///
+    ///\param [in]  ND - Try to mangle this decl's name.
+    ///\param [in]  GD - Provide the GlobalDecl if what is to be mangled is a
+    /// constructor or destructor.
+    ///
+    ///\returns The mangled or pure name (when ND should not be mangled) on
+    /// success or an empty string on failure.
+    ///
+    std::string maybeMangleDeclName(const clang::NamedDecl* ND,
+                                    const clang::GlobalDecl* GD = nullptr);
```
pcanal: Why the change in interface style (returning the string rather than modifying a pass string)?   For consistency sake we should only have one style (The existing style **used to** be more performance but nowadays with move semantic this is no longer the case).

lib/Interpreter/Interpreter.cpp
```diff
@@ -265,6 +275,30 @@ namespace cling {
...
+          if (!Scope) {
+            cling::errs() << "Scope for gCling was not found\n";
+          }
+          else if (const clang::ValueDecl* gCling =
+              m_LookupHelper->findDataMember(Scope, "gCling",
+                                             LookupHelper::NoDiagnostics)) {
+            using namespace utils;
+            const std::string Name = Analyze::maybeMangleDeclName(gCling);
+            if (!Name.empty()) {
+              // gCling gets linked to top-most Interpreter.
```
pcanal: If I read the code correctly, the code below does not implement the comment (since m_Parenting[0] is always set to 'this' rather than the parent.  
In addition, I fail to see/understand the code that would prevent the symbol to be added multiple time (one per interpreter).

What am I missing?

include/cling/Interpreter/Interpreter.h
```diff
@@ -142,6 +142,11 @@ namespace cling {
     ///
     InvocationOptions m_Opts;
 
+    ///\brief Back pointer to *this* which stores parent information.
```
marsupial: This is necessary as putting a symbol into the JIT requires a constant location where it is available and **&this** is illegal.

include/cling/Utils/AST.h
marsupial: That was a separate PR as it had nothing to do with this, but pulled it in.

include/cling/Interpreter/Interpreter.h
pcanal: Indeed, after reading the whole patch, this is what I guess.  Could you still rephrase the doc string to be more explicit/clear?

include/cling/Utils/AST.h
pcanal: > That was a separate PR

Was or is?  Is it still intended to be changed?

lib/Interpreter/Interpreter.cpp
marsupial: My mistake, this was culled from a changes that moved to **thisCling** which links to the child as parent can always be reached through it.

include/cling/Utils/AST.h
marsupial: I rolled it into this one d14e9b4

include/cling/Interpreter/Interpreter.h
marsupial: I did

include/cling/Utils/AST.h
pcanal: Then I am a bit confused (i.e. I am missing something).  d14e9b4 seems to be in the master for a long while, while this PR seems to not include it ... what am I missing?

include/cling/Interpreter/Interpreter.h
pcanal: Thanks.  My apologies for not noticing.

include/cling/Utils/AST.h
marsupial: Huh? Now I'm confused.
d14e9b4 is right below this comment section.
If it appears in any master it would be an interesting SHA1 collision.

PR #171, opened 2017-07-06
Fix broken import of class types.

marsupial:
There are two issues, the first seems like cling's fault, the second possibly clangs.

1. Importing a CxxConstructor decl can fail as cling it is using the string name for lookup.
2. Importing C++ conversion operators does not work.


ferdymercury:

PR #174, opened 2017-07-06
Windows: Fix declaration for C99 and re-enable Gnu.C test.

marsupial:
The test was failing for a reason that likely shouldn't be ignored.

This reverts commit 5947e13cb99052b6f7a5b501244ee2be9be9d080.

### DIFF
--- a/lib/Interpreter/Interpreter.cpp
+++ b/lib/Interpreter/Interpreter.cpp
@@ -433,9 +433,12 @@ namespace cling {
 #else
     const char* Spec = "__clrcall";
 #endif
+    const bool ParamNames = LangOpts.C99;
     Strm << Linkage << " " << Spec << " int (*__dllonexit("
-         << "int (" << Spec << " *f)(void**, void**), void**, void**))"
-         "(void**, void**)";
+         << "int (" << Spec << " *f)(void**, void**),"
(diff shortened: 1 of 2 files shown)


bellenot:
And it's still failing...

marsupial:
Hopefully no longer.

bellenot:
> Remove C++11 test which Visual Studio cannot handle.

And what about the other platforms?

marsupial:
The test is for GNU keywords, having -std=gnu++14 -std=gnu++1z is enough as they test the same mechanism.
-std=gnu++11 really adds nothing but failure for Windows, and fixing for that would mean you're just testing that C++11 cannot run on Windows

If someone feels it is really necessary, then it would probably be best to have a directory devoted to C++11 tests that are not activated on Windows rather than allowing failure all over.

bellenot:
OK, thanks for the explanation @marsupial. If @vgvassilev confirm (and agrees), and if it fixes the test, then it is would be fine with me

marsupial:
Updated again to work-around the Windows issue without dropping it entirely.

vgvassilev:
Looks good to me.

pcanal:
@phsft-bot build

ferdymercury:
Moved to https://github.com/root-project/root/pull/16694

PR #177, opened 2017-07-10
Windows CI: Test multiple fixes against CERN/master.

marsupial:


bellenot:
I quickly tried it and more tests are failing on my local desktop. Even the new Windows tests are failing...

marsupial:
I'm still pushing into it, but last run on Appveyor was down to 6 fails with CMake, and 5 with Ninja.
Hopefully by the end of the day everything (minus exceptions) will pass on Appveyor.

The new Windows test will likely fail until moving to COFF or someone writes the code to support COMDATs on ELF. The point was to pull out the point of failure from multiple tests into one location so for example OutputRedirect.C -real- failures could be adressed.

bellenot:
OK, I see. Thanks. And I suppose this will supersede the ROOT [PR-745](https://github.com/root-project/root/pull/745) (originally Cling PR-169), right?

marsupial:
I just got tired of "This doesn't work" so wanted to gather all the Windows stuff into one place, as there can be multiple reasons for failures in a single test. And cherry-picking these out of my tree was often causing conflicts.

I think the the TLS stuff should go in regardless, as it will currently crash hard on OS X & Linux, so shouldn't depend on whenever these may get in (but it is necessary for Windows to pass as well).

ferdymercury:
@bellenot this one is probably outdated and can be closed? thks

PR #179, opened 2017-07-12
Fix shutdown issues.

marsupial:
UserInterface was being destroyed to early, possibly upsetting the text-input layer on Windows.
Remove an assert for Windows exception handling that was triggering too late in shutdown to be seen and hand condition at runtime.


lib/Utils/PlatformWin.cpp
```diff
@@ -788,11 +788,17 @@ void RegisterEHFrames(uintptr_t ImgBs, const EHFrameInfos& Frames, bool Block) {
 void DeRegisterEHFrames(uintptr_t ImgBase, const EHFrameInfos& Frames) {
   if (Frames.empty())
     return;
-  assert(getImageBaseMap().find(ImgBase) != getImageBaseMap().end());
 
-  // Remove the ImageBase from lookup
+  // There is a chance that DeRegisterEHFrames will without a prior call to
```
vgvassilev: I do not understand this comment.

marsupial: updated

PR #180, opened 2017-07-12
Fix recursive registration of __cxa_atexit & atexit functions.

marsupial:
IncrementalExecutor assumes that no atexit functions will ever be registered by the currently executing atexit function. This assumption is wrong and against the standard.

Additionally the implementation is somewhat inefficient in memory usage using an **llvm::SmallVector<128, CXAAtExitElement>** to store the registered functions, with each **CXAAtExitElement** holding a pointer to the **llvm::Module** that created it.

Linearly traversing the list and removing elements also never really releases the memory associated with a registered function.

### DIFF
--- a/include/cling/Utils/OrderedMap.h
+++ b/include/cling/Utils/OrderedMap.h
@@ -1,0 +1,116 @@
+//--------------------------------------------------------------------*- C++ -*-
+// CLING - the C++ LLVM-based InterpreterG :)
+// author: Roman Zulak
+//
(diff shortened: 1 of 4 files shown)


marsupial:
Ping.
This fixes static variables in destructors as well as returning memory to the host process when it is no longer needed.

bellenot:
Axel will most probably look at it when back from holidays

pcanal:
Thanks.

PR #190, opened 2017-11-07
Link to LLVM's BinaryFormat library, fixing shared library builds.

marsupial:


hahnjo:
> This should be superseded by #219 - which didn't need to link clingJupyter against BinaryFormat. Sure you need that?

Closing accordingly.

PR #196, opened 2017-11-28
[Fix #195] Add "language" field to kernel.json

### LINKED ISSUE #195: No "language" field in kernel.json
This creates integration problems, at least with Hydrogen Atom plugin.  
See https://github.com/nteract/hydrogen/issues/1129

Houkime:
Fixing compatibility with Hydrogen (and possibly sth. else since "language" is an ubiquitous field for kernel.json)
#195  and  nteract/hydrogen#1129

### DIFF
--- a/tools/Jupyter/kernel/cling-cpp11/kernel.json
+++ b/tools/Jupyter/kernel/cling-cpp11/kernel.json
@@ -5,5 +5,6 @@
       "-f",
       "{connection_file}",
       "--std=c++11"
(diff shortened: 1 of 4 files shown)


nikitakit:
By the way, I'm not sure about casing. I think the kernel in root-project/root does lowercase "c++" (I'm coming here from the hydrogen issue tracker)

Axel-Naumann:
Merged, thanks!

PR #198, opened 2017-11-30
Fix cling not found in jupyter kernel when `which cling` is a link to relative path (e.g. ../Cellar/cling/...)

jiegec:
On my mac, `ls -al /usr/local/bin/clang` shows `/usr/local/bin/cling -> ../Cellar/cling/0.5/bin/cling`, thus the code will use `../Cellar/cling/0.5/bin/cling` as the actual path of cling. However, cwd now is no longer `/usr/local/bin`.

### DIFF
--- a/tools/Jupyter/kernel/clingkernel.py
+++ b/tools/Jupyter/kernel/clingkernel.py
@@ -99,6 +99,8 @@ class ClingKernel(Kernel):
         super(ClingKernel, self).__init__(**kwargs)
         try:
             whichCling = os.readlink(shutil.which('cling'))
+            if not os.path.isabs(whichCling):
+                whichCling = os.path.join(os.path.dirname(shutil.which('cling')), whichCling)
         except OSError as e:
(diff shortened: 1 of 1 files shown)


jiegec:
Seems those travis build failed for unrelated reasons.

roynos:
I ran into the same problem. I think you have to check if it is a relative path before using join.

```python
 whichCling = os.readlink(shutil.which('cling'))
 if not os.path.isabs(whichCling):
     whichCling = os.path.join(os.path.dirname(shutil.which('cling')), whichCling)
```

Axel-Naumann:
Do you agree that #199 is a reasonable replacement for this PR?

jiegec:
Yep. Use that pr instead.

lie-yan:
I have a quick workaround. 
```
rm -f /usr/local/bin/cling
ln -s /usr/local/Cellar/cling/0.5_1/libexec/bin/cling /usr/local/bin/cling
```

PR #213, opened 2018-02-02
Fix cmake version check to work with cmake 3.10

saagarjha:
The cmake version check was failing because `"3.10.2" < "3.4.3"`. I've replaced the current string comparison check with one that splits on dots, converts the `string`s to `int`s, and performs a tuple comparison (`(3, 10, 2) > (3, 4, 3)`). It's not perfect but it's better than what's going on right now.

### DIFF
--- a/tools/packaging/cpt.py
+++ b/tools/packaging/cpt.py
@@ -695,7 +695,7 @@ def check_ubuntu(pkg):
             print(pkg.ljust(20) + '[OK]'.ljust(30))
     elif pkg == "cmake":
         CMAKE = os.environ.get('CMAKE', 'cmake')
-        if exec_subprocess_check_output('{cmake} --version'.format(cmake=CMAKE), '/').strip().split('\n')[0].split()[-1] < '3.4.3':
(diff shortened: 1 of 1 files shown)


saagarjha:
Looks like all the CI jobs timed out…

Axel-Naumann:
@vgvassilev objections against me merging this?

vgvassilev:
LGTM!

Axel-Naumann:
@vgvassilev now you can choose between this and PR #241 ... I think I have a slight preference for #241 but I let you decide, and we should merge one of them *soon*!

Axel-Naumann:
Apologies for not acting on this earlier. I have now merged #241 which is very similar to your patch. Thanks a lot for your contribution (even if it didn't end up in the repo due to my oversight)!

saagarjha:
Sure, no problem. Factoring this out into a function was probably a better idea anyways. The only issue I have with that pull request is that code is slightly longer, since it doesn't use the built-in tuple comparison as I do and does the comparison by hand, but that's not really a major concern.

PR #219, opened 2018-03-14
fix CMakeFiles to allow cling build with shared libraries

SimeonEhrig:


Axel-Naumann:
Merged into ROOT (will magically appear in cling soon) - thanks!

Issue #115, opened 2017-02-20
CUDA Support


ax3l:
Hi ROOT team,

I am opening this issue to document and discuss our plan to add CUDA support to Cling. Maybe someone can open and link an issue on https://root.cern/bugs, I neither have a CERN nor an external account and the tracker is not public for people without registration :)

The discussion started after a post of mine in the cling-dev mailing list, where @karies picked me up and we spinned of a longer private discussion.

> On 07.11.2016 10:34, Huebl, Axel wrote:
> [...] I am trying to compile a simple CUDA program, which is supported by clang
> already:
> 
> [...] internal include issues when starting cling via
>   ./cling -x cuda --cuda-path=$CUDA_ROOT --cuda-gpu-arch=sm_35
-L$CUDA_ROOT/lib -lstdc++ -lcudart [...]

By now, @karies did already fix all issues during startup and we think we are now at the point where the one could work on accessing clang's PTX emitter to generate PTX code and then link it.

```
$ cling -x cuda --cuda-gpu-arch=sm_35 -nocudainc
atexit not in Module!
at_quick_exit not in Module!

****************** CLING ******************
* Type C++ code and press enter to run it *
*             Type .q to exit             *
*******************************************
[cling]$ 
```

Clang currently translates CUDA code by linking the PTX code in a fat binary and passing it the CUDA driver during runtime, which generates SASS code (shader assembly) from it to execute. This is similar to what `nvcc` provides, besides that `nvcc` can additionally generate and link `SASS` code for a specific compute architecture directly, but that's not important here ([see this page for further details](http://llvm.org/docs/CompileCudaWithLLVM.html)).

From our discussions I understood, that cling already has similar functionality for e.g. [PowerPC](https://github.com/root-mirror/cling/blob/097a4e3fbd38b211709a8f5a2c5fb0bc5fe715e3/lib/Interpreter/IncrementalExecutor.cpp#L68) in place to target specific emitters and add their assembler artifacts. Can you guide us how one could add the same functionality for PTX?

At @gcoe-dresden, which is a collaboration of research groups in and around Dresden (Technical University, Max-Planck, Helmholtz-Zentrum Dresden-Rossendorf), we are currently discussing the possibilities for interactive simulations, RT profiling and tuning, teaching, rapid prototyping and interactive simulations and much more that one could get from a CUDA capable interpreter. Long story short: exciting possibilities!

[...]

PR #220, opened 2018-03-14
Cuda host side

SimeonEhrig:
This commits allows to enable the cuda mode of the clang compiler instance and translate cuda runtime code on the host side.

Follow functions will added:
- set the CUDA AuxTarget, if CUDA is detected
- detect the CUDA language at CompilerOptions and set new Flag
- add Argument --cuda-host-only if the Argument -xclang is set at start of clang
- set default c++ standard, if language is cuda
- load libcudart.so at start of cling -xcuda (at the moment, just linux is supported)

### DIFF
--- a/include/cling/Interpreter/InvocationOptions.h
+++ b/include/cling/Interpreter/InvocationOptions.h
@@ -58,6 +58,7 @@ namespace cling {
     unsigned HasOutput : 1;
     unsigned Verbose : 1;
     unsigned CxxModules : 1;
+    unsigned CUDA : 1;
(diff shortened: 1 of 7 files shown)


Axel-Naumann:
Can you remove the other branch's commits, please?

SimeonEhrig:
Ok, I've removed the CMakeList commit.

Axel-Naumann:
Merged (through ROOT). Thanks, @SimeonEhrig !

ax3l:
Link: #115

PR #225, opened 2018-04-14
fix up type printing for fully qualified templates

wolfv:
We had some issues with not fully qualified template arguments in the value printer (the xeus-cling project uses derived functions from ValuePrinter).

We found this change to fix our issues. 

Do you see any drawbacks?

### DIFF
--- a/lib/Interpreter/ValuePrinter.cpp
+++ b/lib/Interpreter/ValuePrinter.cpp
@@ -79,8 +79,10 @@ static std::string enclose(std::string Mid, const char* Begin,
 static std::string enclose(const clang::QualType& Ty, clang::ASTContext& C,
                            const char* Begin = "(", const char* End = "*)",
                            size_t Hint = 3) {
-  return enclose(cling::utils::TypeName::GetFullyQualifiedName(Ty, C),
-                 Begin, End, Hint);
+  clang::PrintingPolicy Policy(C.getPrintingPolicy());
+  Policy.PolishForDeclaration = true;
+  std::string Name = Ty.getAsString(Policy);
+  return enclose(Name, Begin, End, Hint);
 }
 
 static clang::QualType


Axel-Naumann:
@vgvassilev do I remember correctly that you tried this and it triggered failures in ROOT?

FYI, @wolfv - i'm currently trying what it takes to rewrite this as AST generation instead of source generation. That would get rid of the need of spelling type names correctly.

pcanal:
In addition, we use cling::utils::TypeName::GetFullyQualifiedName in other context, it really ought to be fixed or we will see similar problems in those context.

wolfv:
Just FYI the easiest way to trigger the bug is the following:

```
****************** CLING ******************
* Type C++ code and press enter to run it *
*             Type .q to exit             *
*******************************************
[cling]$ #include <vector>
[cling]$ auto a = std::vector<int>{1,2,3}
input_line_6:2:39: error: no template named 'allocator'; did you mean 'std::allocator'?
 cling::printValue(*(std::vector<int, allocator<int> >**)0x7ffd8caa5928);
                                      ^~~~~~~~~
                                      std::allocator
/home/wolfv/miniconda3/bin/../lib/gcc/../../gcc/include/c++/bits/allocator.h:108:11: note: 'std::allocator' declared here
    class allocator: public __allocator_base<_Tp>
          ^
ERROR in cling's callPrintValue(): cannot pass value!
(std::vector<int, allocator<int> > &) ERROR in cling's callPrintValue(): missing value string.
```

I've also backported the latest `clang` `getFullyQualifiedName` function, which has the same shortcomings as the cling one (which seems to be derived from a slightly older clang version).

So in the end it seems to be a bug in clang's `getFullyQualifiedName` which doesn't resolve template arguments in some cases. It might make sense to file a bug against the clang project.

I looked at the source code for `__PRETTY_FUNCTION__` to arrive at the solution I am pushing here.

vgvassilev:
As I said in the other issue, clang has a problem in printing ElaboratedTypes. I will fix it whenever I can reproduce that with standalone clang.

SylvainCorlay:
For reference, pointing to issue https://github.com/vgvassilev/cling/issues/180

SylvainCorlay:
It seems that that `GetFullyQualifiedName` is now part of the API of clang. So I guess that we could avoid this duplication for the next cling release.

SylvainCorlay:
> @vgvassilev do I remember correctly that you tried this and it triggered failures in ROOT?
>
> FYI, @wolfv - i'm currently trying what it takes to rewrite this as AST generation instead of source generation. That would get rid of the need of spelling type names correctly.

@Axel-Naumann note that we use this for the rich display in the xeus-cling kernel (also using code generation):

screen shot 2018-04-14 at 7 30 45 pm

Since the `getTypeString` is not part of the cling API, we had to duplicate some of the code from cling. I opened https://github.com/vgvassilev/cling/issues/177 asking to make this API public.

Let me know if your work to make this using AST generation would break our approach. 

Note that unlike the case of `printValue`, we used ADL so that people may overload the `mime_bundle_repr` in their own namespace instead of specializing for their own type.

SylvainCorlay:
Note: This does not seem to solve all the cases. I still have some cases where namespaces are missing.

Axel-Naumann:
Thanks everyone. Let me try to summarize:
- ROOT continues to need `GetFullyQualifiedName`, ideally moving to clang's
- there exists a clang version that has a different set of bugs (see https://github.com/root-project/root/pull/999 )
- @vgvassilev is hunting an issue that exists in possibly both versions
- I will collect failures in name printing to verify that `GetFullyQualifiedName` works, to be used as clang tests
- for value printing and value extraction I am evaluating moving to AST operations for the type printing part

@pcanal you were involved with clang's `GetFullyQualifiedName`. cling's does not take a set of "don't resolve" types (`Double32_t`) so I see nothing fundamental blocking us (neither cling nor ROOT's use of it) from moving to clang's.

@SylvainCorlay @wolfv I will leave this one on hold until we know the effect of @vgvassilev 's patch on printing of elaborate types.

@SylvainCorlay I'd appreciate if you could move to this repo's issue tracker, I am not following the issues on other repos' trackers, to avoid duplicates. Thanks!

SylvainCorlay:
> @SylvainCorlay I'd appreciate if you could move to this repo's issue tracker, I am not following the issues on other repos' trackers, to avoid duplicates. Thanks!

Done. I have moved all the recent issues.

You might have a chance to merge all the stars / issues of the two mirrors with GitHub customer service, if they are understanding!

pcanal:
> cling's does not take a set of "don't resolve" types (Double32_t) so I see nothing fundamental blocking us (neither cling nor ROOT's use of it) from moving to clang's.

Yes, it is time to switch over ....

PR #232, opened 2018-05-02
Interpreter and other classes have been modified for support null args

bisegni:
This patch permit to use interpreter without give the argc(0) and argv(nullptr)

### DIFF
--- a/lib/Interpreter/CIFactory.cpp
+++ b/lib/Interpreter/CIFactory.cpp
@@ -828,7 +828,11 @@ static void stringifyPreprocSetting(PreprocessorOptions& PPOpts,
 
     const size_t argc = COpts.Remaining.size();
     const char* const* argv = &COpts.Remaining[0];
-    std::vector<const char*> argvCompile(argv, argv+1);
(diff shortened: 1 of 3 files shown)


lib/Interpreter/CIFactory.cpp
```diff
@@ -828,7 +828,11 @@ static void stringifyPreprocSetting(PreprocessorOptions& PPOpts,
 
     const size_t argc = COpts.Remaining.size();
     const char* const* argv = &COpts.Remaining[0];
-    std::vector<const char*> argvCompile(argv, argv+1);
+    std::vector<const char*> argvCompile;
+    if(argc && argv) {
+        argvCompile = std::vector<const char*>(argv, argv+1);
+    }
```
pcanal: It might be better/needed to have a non-empty default for argvCompile?

ferdymercury: I think this repo is just a mirror, consider moving your PR here instead: https://github.com/root-project/root/pulls (subdirectory interpreter/cling). Thanks!

PR #233, opened 2018-05-07
Unique cuda module ctor name

SimeonEhrig:
This function pass change the symbol of the symbols __cuda_module_ctor and __cuda_module_dtor to unique symbols. It appends the module name and change the symbols to preprocessor compliant symbols.

Attention, the test case works fine, but it needs a cuda fatbinary. At the moment there is just a really ugly way to start the test with a cuda fatbinary. 

`LD_PRELOAD=/usr/local/cuda/lib64/libcudart.so cat <path_to_cling>/cling/src/tools/cling/test/CodeGeneration/CUDACtorDtor.C | <path_to_cling>/cling/build/./bin/cling --nologo -x cuda --cuda-host-only -l/usr/local/cuda/lib64/libcudart.so -Xclang -fcuda-include-gpubinary -Xclang <path_to_cling>/cling/build/bin/dummy.fatbin -Xclang -verify 2>&1 | <path_to_cling>/cling/build/./bin/FileCheck <path_to_cling>/cling/src/tools/cling/test/CodeGeneration/CUDACtorDtor.C`

In a future pull request (cuda device side), there will be code, which will handle the cuda fatbinary by the cling itself. If the cuda device side code is integrated, the testcase should works fine.

### DIFF
--- a/lib/Interpreter/BackendPasses.cpp
+++ b/lib/Interpreter/BackendPasses.cpp
@@ -186,7 +186,11 @@ void BackendPasses::CreatePasses(llvm::Module& M, int OptLevel)
   m_MPM[OptLevel].reset(new legacy::PassManager());
 
   m_MPM[OptLevel]->add(new KeepLocalGVPass());
-  m_MPM[OptLevel]->add(new UniqueCUDAStructorName());
+  // The function __cuda_module_ctor and __cuda_module_dtor will just generated,
+  // if a CUDA fatbinary file exist. Without file path there is no need for the
+  // function pass.
+  if(!m_CGOpts.CudaGpuBinaryFileNames.empty())
+    m_MPM[OptLevel]->add(new UniqueCUDAStructorName());
   m_MPM[OptLevel]->add(createTargetTransformInfoWrapperPass(
                                                    m_TM.getTargetIRAnalysis()));
 
--- a/lib/Interpreter/BackendPasses.cpp
+++ b/lib/Interpreter/BackendPasses.cpp
@@ -25,6 +25,7 @@
 //#include "clang/Basic/LangOptions.h"
 //#include "clang/Basic/TargetOptions.h"
 #include "clang/Frontend/CodeGenOptions.h"
+#include "clang/Basic/CharInfo.h"
 
 using namespace cling;
 using namespace clang;
@@ -81,7 +82,20 @@ namespace {
     bool runOnGlobal(GlobalValue& GV, StringRef & ModuleName){
       if(GV.hasName() && (GV.getName() == "__cuda_module_ctor"
           || GV.getName() == "__cuda_module_dtor") ){
-        GV.setName(GV.getName() + "_" + ModuleName);
+        llvm::SmallString<128> NewFunctionName;
+        NewFunctionName.append(GV.getName());
+        NewFunctionName.append("_");
+        NewFunctionName.append(ModuleName);
+
+        for (size_t i = 0; i < NewFunctionName.size(); ++i) {
+          // Replace everything that's not [a-zA-Z0-9._] with a _. This set happens
+          // to be the set of C preprocessing numbers.
+          if (!isPreprocessingNumberBody(NewFunctionName[i]))
+            NewFunctionName[i] = '_';
+        }
+
+        GV.setName(NewFunctionName);
+
         return true;
       }
 
--- a/lib/Interpreter/BackendPasses.cpp
+++ b/lib/Interpreter/BackendPasses.cpp
@@ -87,8 +87,8 @@ namespace {
         NewFunctionName.append(ModuleName);
 
         for (size_t i = 0; i < NewFunctionName.size(); ++i) {
-          // Replace everything that's not [a-zA-Z0-9._] with a _. This set happens
-          // to be the set of C preprocessing numbers.
+          // Replace everything that is not [a-zA-Z0-9._] with a _. This set
+          // happens to be the set of C preprocessing numbers.
           if (!isPreprocessingNumberBody(NewFunctionName[i]))
             NewFunctionName[i] = '_';
         }
@@ -106,7 +106,7 @@ namespace {
 
     bool runOnModule(Module &M) override {
       bool ret = false;
-      StringRef ModuleName = M.getName();
+      const StringRef ModuleName = M.getName();
       for (auto &&F: M)
         ret |= runOnFunction(F, ModuleName);
       return ret;
--- a/lib/Interpreter/BackendPasses.cpp
+++ b/lib/Interpreter/BackendPasses.cpp
@@ -69,6 +69,39 @@ namespace {
 
 char KeepLocalGVPass::ID = 0;
 
+namespace {
+
+  // Add a suffix to the CUDA module ctor/dtor to generate an unique name.
+  // It's necessary for lazy compilation. Without suffix, cling can't
+  // differentiate the ctor/dtor of the first module from the ctor/dtor of the
+  // following modules.
+  class UnqiueCUDACtorDtorName : public ModulePass {
+    static char ID;
+
+    bool runOnGlobal(GlobalValue& GV, StringRef & ModuleName){
+      if(GV.hasName() && (GV.getName() == "__cuda_module_ctor"
+          || GV.getName() == "__cuda_module_dtor") ){
+        GV.setName(GV.getName() + "_" + ModuleName);
+        return true;
+      }
+
+      return false;
+    }
+
+  public:
+    UnqiueCUDACtorDtorName() : ModulePass(ID) {}
+
+    bool runOnModule(Module &M) override {
+      bool ret = false;
+      StringRef ModuleName = M.getName();
+      for (auto &&F: M)
+        ret |= runOnGlobal(F, ModuleName);
+      return ret;
+    }
+  };
+}
+
+char UnqiueCUDACtorDtorName::ID = 0;
 
 BackendPasses::~BackendPasses() {
   //delete m_PMBuilder->Inliner;
@@ -140,6 +173,7 @@ void BackendPasses::CreatePasses(llvm::Module& M, int OptLevel)
   m_MPM[OptLevel].reset(new legacy::PassManager());
 
   m_MPM[OptLevel]->add(new KeepLocalGVPass());
+  m_MPM[OptLevel]->add(new UnqiueCUDACtorDtorName());
   m_MPM[OptLevel]->add(createTargetTransformInfoWrapperPass(
                                                    m_TM.getTargetIRAnalysis()));
 
--- a/lib/Interpreter/BackendPasses.cpp
+++ b/lib/Interpreter/BackendPasses.cpp
@@ -78,11 +78,11 @@ namespace {
   class UniqueCUDAStructorName : public ModulePass {
     static char ID;
 
-    bool runOnGlobal(GlobalValue& GV, const StringRef ModuleName){
-      if(GV.hasName() && (GV.getName() == "__cuda_module_ctor"
-          || GV.getName() == "__cuda_module_dtor") ){
+    bool runOnFunction(Function& F, const StringRef ModuleName){
+      if(F.hasName() && (F.getName() == "__cuda_module_ctor"
+          || F.getName() == "__cuda_module_dtor") ){
         llvm::SmallString<128> NewFunctionName;
-        NewFunctionName.append(GV.getName());
+        NewFunctionName.append(F.getName());
         NewFunctionName.append("_");
         NewFunctionName.append(ModuleName);
 
@@ -93,7 +93,7 @@ namespace {
             NewFunctionName[i] = '_';
         }
 
-        GV.setName(NewFunctionName);
+        F.setName(NewFunctionName);
 
         return true;
       }
@@ -108,7 +108,7 @@ namespace {
       bool ret = false;
       StringRef ModuleName = M.getName();
       for (auto &&F: M)
-        ret |= runOnGlobal(F, ModuleName);
+        ret |= runOnFunction(F, ModuleName);
       return ret;
     }
   };
--- a/lib/Interpreter/BackendPasses.cpp
+++ b/lib/Interpreter/BackendPasses.cpp
@@ -72,14 +72,13 @@ char KeepLocalGVPass::ID = 0;
 
 namespace {
 
-  // Add a suffix to the CUDA module ctor/dtor to generate an unique name.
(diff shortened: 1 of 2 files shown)


lib/Interpreter/BackendPasses.cpp
```diff
@@ -69,6 +70,52 @@ namespace {
 
 char KeepLocalGVPass::ID = 0;
 
+namespace {
+
+  // Add a suffix to the CUDA module ctor/dtor to generate an unique name.
```
Axel-Naumann: "a unique"

lib/Interpreter/BackendPasses.cpp
```diff
@@ -69,6 +70,52 @@ namespace {
...
+  // It's necessary for lazy compilation. Without suffix, cling can't
+  // differentiate the ctor/dtor of the first module from the ctor/dtor of the
```
Axel-Naumann: "can't distinguish ctor/dtor of subsequent modules" (else it sounds as if there's something special about the first module).

lib/Interpreter/BackendPasses.cpp
```diff
@@ -69,6 +70,52 @@ namespace {
...
+  // following modules.
+  class UnqiueCUDACtorDtorName : public ModulePass {
+    static char ID;
+
+    bool runOnGlobal(GlobalValue& GV, StringRef & ModuleName){
```
Axel-Naumann: `StringRef` shouldn't be passed by reference.

lib/Interpreter/BackendPasses.cpp
```diff
@@ -69,6 +70,52 @@ namespace {
...
+      return false;
+    }
+
+  public:
+    UnqiueCUDACtorDtorName() : ModulePass(ID) {}
+
+    bool runOnModule(Module &M) override {
+      bool ret = false;
+      StringRef ModuleName = M.getName();
+      for (auto &&F: M)
```
Axel-Naumann: `F` should be of type `llvm::Function`. Should `runOnGlobal` be called `runOnFunction` (and take a `Function`)?

test/CodeGeneration/CUDACtorDtor.C
```diff
@@ -0,0 +1,62 @@
...
+#include "llvm/IR/Module.h"
+#include <iostream>
+
+// It compares the cuda module ctor and dtor of two random modules.
+std::string ctor1, ctor2, dtor1, dtor2;
+
+auto M = gCling->getLatestTransaction()->getModule();
+
+for(auto I = M->begin(), E = M->end(); I != E; ++I){
+  if((*I).getName().startswith_lower("__cuda_module_ctor_")){
```
Axel-Naumann: Why does this test for a trailing `'_'` while the code in the pass does not?

test/CodeGeneration/CUDACtorDtor.C
```diff
@@ -0,0 +1,62 @@
...
+
+  if((*I).getName().startswith_lower("__cuda_module_dtor_")){
+    dtor1 = (*I).getName().str();
+  }
+}
+
+// The last module should be another, because of the for loop.
+M = gCling->getLatestTransaction()->getModule();
+
+for(auto I = M->begin(), E = M->end(); I != E; ++I){
```
Axel-Naumann: range for?

test/CodeGeneration/CUDACtorDtor.C
```diff
@@ -0,0 +1,62 @@
...
+  if((*I).getName().startswith_lower("__cuda_module_ctor_")){
+    ctor1 = (*I).getName().str();
+  }
+
+  if((*I).getName().startswith_lower("__cuda_module_dtor_")){
+    dtor1 = (*I).getName().str();
+  }
+}
+
+// The last module should be another, because of the for loop.
```
Axel-Naumann: Could you `assert()` that?

test/CodeGeneration/CUDACtorDtor.C
```diff
@@ -0,0 +1,62 @@
...
+    ctor2 = (*I).getName().str();
+  }
+
+  if((*I).getName().startswith_lower("__cuda_module_dtor_")){
+    dtor2 = (*I).getName().str();
+  }
+}
+
+// Check if the ctor and dtor of the two modules are different.
+bool result1 = ctor1 != ctor2
```
Axel-Naumann: `ctor1 != ctor2` should be enough? (I.e. no need to assign it to a variable.)

test/CodeGeneration/CUDACtorDtor.C
```diff
@@ -0,0 +1,62 @@
...
+// CHECK: (bool) true
+bool result2 = dtor1 != dtor2
+// CHECK: (bool) true
+
+// Check if the symbols are preprocessor compliant.
```
Axel-Naumann: How does that `find()` below check that?

lib/Interpreter/BackendPasses.cpp
```diff
@@ -69,6 +70,52 @@ namespace {
 
 char KeepLocalGVPass::ID = 0;
 
+namespace {
+
+  // Add a suffix to the CUDA module ctor/dtor to generate an unique name.
+  // It's necessary for lazy compilation. Without suffix, cling can't
```
ax3l: It's -> This is

Do not use appreviated in written language, e.g. can't -> "cannot"

test/CodeGeneration/CUDACtorDtor.C
```diff
@@ -0,0 +1,62 @@
...
+// RUN: cat %s | %cling -x cuda -Xclang -verify 2>&1 | FileCheck %s
+// REQUIRES: cuda-runtime
+
+#include "cling/Interpreter/Interpreter.h"
+#include "cling/Interpreter/Transaction.h"
+#include "llvm/IR/Function.h"
+#include "llvm/IR/Module.h"
+#include <iostream>
+
+// It compares the cuda module ctor and dtor of two random modules.
```
ax3l: Spelling suggestion: It compares -> Compare the ...

test/CodeGeneration/CUDACtorDtor.C
ax3l: Also checking ctor2/dtor2 instead would be nice since their result is not the init value zero (to check if the unique counter works).

lib/Interpreter/BackendPasses.cpp
```diff
@@ -69,6 +70,52 @@ namespace {
...
+
+      return false;
+    }
+
+  public:
+    UnqiueCUDACtorDtorName() : ModulePass(ID) {}
+
+    bool runOnModule(Module &M) override {
+      bool ret = false;
+      StringRef ModuleName = M.getName();
```
ax3l: I think this var can be made `const`

lib/Interpreter/BackendPasses.cpp
```diff
@@ -69,6 +70,52 @@ namespace {
...
+        return true;
+      }
+
+      return false;
+    }
+
+  public:
+    UnqiueCUDACtorDtorName() : ModulePass(ID) {}
+
+    bool runOnModule(Module &M) override {
```
ax3l: I think Module should be able to be passed as const ref

lib/Interpreter/BackendPasses.cpp
ax3l: also the parameters can probably be passed as `const`

lib/Interpreter/BackendPasses.cpp
Axel-Naumann: It's an override; I bet the base only has an overload taking a non-const `Module&`?

ax3l: uh, that's true here.

lib/Interpreter/BackendPasses.cpp
```diff
@@ -69,6 +70,52 @@ namespace {
 
 char KeepLocalGVPass::ID = 0;
 
+namespace {
+
+  // Add a suffix to the CUDA module ctor/dtor to generate an unique name.
+  // It's necessary for lazy compilation. Without suffix, cling can't
+  // differentiate the ctor/dtor of the first module from the ctor/dtor of the
+  // following modules.
+  class UnqiueCUDACtorDtorName : public ModulePass {
```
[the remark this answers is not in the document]
SimeonEhrig: I don't know the term 'structor'. But if it is a generic term for global/static ctor and dtor, we can rename the class.

test/CodeGeneration/CUDACtorDtor.C
SimeonEhrig: The original name of the function is `__cuda_module_ctor` or `__cuda_module_dtor`, which the pass should detect. If it detects, it append an '_' and the module name.

In the test case, I check, if the '_' exist, because it is a sign, that the name was modified. I also could get the name of module, append to the search string and use equals instead startswith. 

At the moment, it is a simplified test, to check if the function name was modified. I also can write a more complicated and safer test method.

Axel-Naumann: Thanks, makes perfect sense. Please simply add a comment in the test, explaining why you test for the extra trailing `'_'`.

lib/Interpreter/BackendPasses.cpp
SimeonEhrig: It's true, F is a llvm::Function object. I could use llvm::Function instead llvm::GlobalValue. Basically, `__cuda_module_ctor` and `__cuda_module_dtor` are just functions. But in the context of the program, they get the meaning of a global ctor and dtor.

I think, function is better, because we don't need to cast the object and in this moment of the compiler pipeline, it is just a simple function and no global ctor or dtor.

lib/Interpreter/BackendPasses.cpp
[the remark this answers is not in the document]
Axel-Naumann: > the term 'structor'. But if it is a generic term for global/static ctor and dtor

Yeah, "structor" is a term of art for "con**structor** and / or de**structor**".

lib/Interpreter/BackendPasses.cpp
Axel-Naumann: > they get the meaning of a global ctor and dtor

Well, any `llvm::Function` is a `llvm::Global`... These are all just symbols. Not sure I understand your argument, but I believe we agree on the conclusion!

lib/Interpreter/BackendPasses.cpp
```diff
@@ -69,6 +70,51 @@ namespace {
...
+    bool runOnFunction(Function& F, const StringRef ModuleName){
+      if(F.hasName() && (F.getName() == "__cuda_module_ctor"
+          || F.getName() == "__cuda_module_dtor") ){
+        llvm::SmallString<128> NewFunctionName;
+        NewFunctionName.append(F.getName());
+        NewFunctionName.append("_");
+        NewFunctionName.append(ModuleName);
+
+        for (size_t i = 0; i < NewFunctionName.size(); ++i) {
+          // Replace everything that's not [a-zA-Z0-9._] with a _. This set happens
```
ax3l: possible improvement: that's -> that is

lib/Interpreter/BackendPasses.cpp
```diff
@@ -69,6 +70,51 @@ namespace {
...
+
+      return false;
+    }
+
+  public:
+    UniqueCUDAStructorName() : ModulePass(ID) {}
+
+    bool runOnModule(Module &M) override {
+      bool ret = false;
+      StringRef ModuleName = M.getName();
```
ax3l: `const StringRef ModuleName = M.getName();` should be possible, no?

test/CodeGeneration/CUDACtorDtor.C
```diff
@@ -0,0 +1,73 @@
...
+// RUN: cat %s | %cling -x cuda -Xclang -verify 2>&1 | FileCheck %s
+// REQUIRES: cuda-runtime
+
+#include "cling/Interpreter/Interpreter.h"
+#include "cling/Interpreter/Transaction.h"
+#include "llvm/IR/Function.h"
+#include "llvm/IR/Module.h"
+#include <iostream>
+
+// Compare the the cuda module ctor and dtor of two random modules.
```
ax3l: typo: double `the`

test/CodeGeneration/CUDACtorDtor.C
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  if(I.getName().startswith_lower("__cuda_module_dtor_")){
+    dtor1 = I.getName().str();
+  }
+}
+
+auto M2 = gCling->getLatestTransaction()->getModule();
+
+// The last module should be another, because of the for loop.
```
ax3l: comment could be changed to:
```
// the two modules should have different names
```

test/CodeGeneration/CUDACtorDtor.C
```diff
@@ -0,0 +1,73 @@
+//------------------------------------------------------------------------------
+// CLING - the C++ LLVM-based InterpreterG :)
+//
+// This file is dual-licensed: you can choose to license it under the University
+// of Illinois Open Source License or the GNU Lesser General Public License. See
+// LICENSE.TXT for details.
+//------------------------------------------------------------------------------
+
+// The Test checks, if the symbols __cuda_module_ctor and __cuda_module_dtor are
+// unique for every module. Attention, for a working test case, a cuda
```
ax3l: How are we planning to provide that fatbin file and what is it's corresponding CUDA code?

test/CodeGeneration/CUDACtorDtor.C
```diff
@@ -36,7 +36,7 @@ for(auto &I : *M1){
 
 auto M2 = gCling->getLatestTransaction()->getModule();
 
-// The last module should be another, because of the for loop.
+// The two modules should have different names, because of the for loop.
```
ax3l: quick question: *which* for loop? I just don't find it :)

SimeonEhrig: https://github.com/SimeonEhrig/cling/blob/52c38ec2366308eb874431d8d3e7e176dd897f36/test/CodeGeneration/CUDACtorDtor.C#L26-L35

ax3l: Yes, but the names are not different be cause you read the name in the loop for M1 but they are different because you *create* two modules - with two calls to `getModule()`. Or do I get that wrong?

SimeonEhrig:
I have done all changes. I think, the commit is ready for merge.

ax3l:
Link: #115

PR #243, opened 2018-06-03
Fix handling of Control+C and Control+D

saagarjha:
This pull request fixes most of #212, as well as includes a fix for a typo that I noticed. I say "most" because <kbd>Control</kbd>+<kbd>C</kbd> doesn't interrupt the interpreter, as it does in other programs do. Is this something that's worth pursing?

PR #245, opened 2018-06-16
Add support for interrupting execution with ^C

saagarjha:
I finished up <kbd>Control</kbd>+<kbd>C</kbd> support by interrupting execution whenever a `SIGINT` is received, as described by #243. I still can't get cling to build on my personal computer, so this pull request was tested on a Linux machine instead. Do let me know if there's anything I need to change!

### DIFF
--- a/lib/Utils/Signal.cpp
+++ b/lib/Utils/Signal.cpp
@@ -1,0 +1,21 @@
+//------------------------------------------------------------------------------
+// CLING - the C++ LLVM-based InterpreterG :)
+// author:  Saagar Jha <saagar@saagarjha.com>
+//
+// This file is dual-licensed: you can choose to license it under the University
+// of Illinois Open Source License or the GNU Lesser General Public License. See
+// LICENSE.TXT for details.
+//------------------------------------------------------------------------------
+
+#include "cling/Utils/Signal.h"
+
+#include "llvm/Support/Signals.h"
+
+jmp_buf before_execution;
+
+void InterruptHandler() {
+  // LLVM automatically resets the handler after every call. Set it back to our
+  // handler for the next time we get a SIGINT.
+  llvm::sys::SetInterruptFunction(InterruptHandler);
+  longjmp(before_execution, 1);
+}
--- a/include/cling/Interpreter/Interpreter.h
+++ b/include/cling/Interpreter/Interpreter.h
@@ -131,6 +131,8 @@ namespace cling {
       kExeUnkownFunction,
       ///\brief The Transaction had no module (probably an error in CodeGen).
       kExeNoModule,
+      ///\brief Execution was interrupted (recieved a SIGINT during execution).
+      kExeInterrupted,
 
       ///\brief Number of possible results.
       kNumExeResults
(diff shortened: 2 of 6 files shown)


saagarjha:
Hey @Axel-Naumann, would you mind taking a look at this pull request? If there's anything wrong with it, it would be great if you'd let me know so I could fix it.

lib/Utils/Signal.cpp
```diff
@@ -0,0 +1,21 @@
...
+//------------------------------------------------------------------------------
+
+#include "cling/Utils/Signal.h"
+
+#include "llvm/Support/Signals.h"
+
+jmp_buf before_execution;
+
+void InterruptHandler() {
+  // LLVM automatically resets the handler after every call. Set it back to our
```
Axel-Naumann: Is there no way for us to teach llvm not to do that? If not, can we re-use their handler?

saagarjha: I'm not completely sure, but from what I saw there didn't seem like a way to do this. The LLVM signal handling code, [here](https://github.com/llvm-mirror/llvm/blob/3c8af5480a208e84dc6ffc24285611c15b51177b/lib/Support/Unix/Signals.inc#L228), seems to unconditionally unset the signal handler just before running it:

```c++
void (*IF)() = InterruptFunction;
Guard.unlock();
InterruptFunction = nullptr;
IF();        // run the interrupt function.
```

saagarjha:
@Axel-Naumann, does this break any signal handlers that you need to keep intact? I haven't really found a better way to do this, so if you have feedback on improving this I'm all ears.

Axel-Naumann:
Apologies that I come back to this only now. With MSVC I get:
```
C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.14.26428\include\csetjmp(18): error C2873: 'setjmp': symbol cannot be used in a using-declaration (compiling source file C:\build\workspace\root-pullrequests-build\root\interpreter\cling\lib\Interpreter\Interpreter.cpp) [C:\build\workspace\root-pullrequests-build\build\interpreter\llvm\src\tools\cling\lib\Interpreter\obj.clingInterpreter.vcxproj]
11:01:39          C:\build\workspace\root-pullrequests-build\root\interpreter\cling\lib\Interpreter\Interpreter.cpp(1083): error C3861: 'setjmp': identifier not found [C:\build\workspace\root-pullrequests-build\build\interpreter\llvm\src\tools\cling\lib\Interpreter\obj.clingInterpreter.vcxproj]
```

saagarjha:
Huh, that should work; are you sure that your Visual Studio install is fine? I don't have a Windows machine at the moment to test this, unfortunately; could I take a look at your `C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.14.26428\include\csetjmp` file perchance?

Axel-Naumann:
@bellenot help!

saagarjha:
Question: what happens if you place `#include <csetjmp>` with `#include <setjmp.h>`?

Axel-Naumann:
This has been addressed in the master. Thanks for your contribution!

PR #248, opened 2018-06-27
Fix doxygen error.

SimeonEhrig:
Fix doxygen error.


Axel-Naumann:
Merged through ROOT. Thanks, Simeon!

PR #252, opened 2018-08-02
Fix detection of Visual Studio directories

bellenot:
- Add the key name in the error messages, instead of simply RegQueryValueEx and RegOpenKeyEx
- Check for "VisualStudio" and/or "VCExpress" in the registry only if the Version is less than 15 (Visual Studio 2017 is not adding any entry in the registry anymore)
- Only check for Visual Studio Dirs with Visual Studio 14 or less (should we really try this at all? since in general we cannot mix different versions...)

This prevents flooding the users with many messages like:
RegOpenKeyEx: returned 2: The system cannot find the file specified.
RegQueryValueEx: returned 2: The system cannot find the file specified.

And apparently, only the Windows Kit 10 is needed (and not complete Visual Studio)...

### DIFF
--- a/lib/UserInterface/textinput/TerminalDisplayWin.cpp
+++ b/lib/UserInterface/textinput/TerminalDisplayWin.cpp
@@ -19,6 +19,12 @@
 
 #include <assert.h>
 
+#ifdef UNICODE
+#define filename L"CONOUT$"
+#else
+#define filename "CONOUT$"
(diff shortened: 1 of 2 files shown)


bellenot:
@marsupial I would appreciate your feedback on this one... :wink:

bellenot:
ping

PR #262, opened 2018-09-27
[WIP^2] add asm and llvm print out

wolfv:
As discussed on the discourse chat with Axel ( https://root-forum.cern.ch/t/get-assembly-llvm-bytecode-or-clang-ast-output/30800 )

This PR is just a WIP to share some tiny progress made towards adding support for printing out assembly and LLVM bytecode through cling macros.

So far I am in the stage of getting familiar with the LLVM/clang/cling internal functions. 

It appears to be quite easy to get the LLVM bytecode by dumping the generated module.

The process to obtain human-readable assembly seems to be a bit harder (or I just don't know how to do it). So far I've basically copy-pasted code that I found in Julia's coda ( https://github.com/JuliaLang/julia/blob/6caabc9d8b0146ef0edf5749d890a270d6a3da7e/src/disasm.cpp#L617 )

Right now I am stuck at getting the ArrayRef that (I guess) should contain the assembled assemby. I also don't know the true meaning of `slice` in this context. 
BUT it already successfully prints a `.text` assembly begin, as well as the added comment :)

Any help or pointers appreciated!

### DIFF
--- a/lib/Interpreter/IncrementalJIT.h
+++ b/lib/Interpreter/IncrementalJIT.h
@@ -23,10 +23,33 @@
 #include "llvm/ExecutionEngine/RTDyldMemoryManager.h"
 #include "llvm/Target/TargetMachine.h"
 
+#include <llvm/MC/MCInst.h>
+#include <llvm/MC/MCStreamer.h>
+#include <llvm/MC/MCSubtargetInfo.h>
+#include <llvm/MC/MCObjectFileInfo.h>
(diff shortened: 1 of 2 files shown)


ferdymercury:
This repo is just a mirror, consider moving your PR here instead: https://github.com/root-project/root/pulls (subdirectory interpreter/cling). Thanks!

wolfv:
It's been 5 years and I think cling is now also clang-repl! I'll close it :)

PR #270, opened 2018-11-04
Use -q instead of -qa when calling rpm

fghzxm:
Also, the original `rpm -qa` calls were missing the argument for `-qa`.  This commit fixes that.


ferdymercury:
I believe this repo is just a mirror, consider moving your PR here instead: https://github.com/root-project/root/pulls (subdirectory interpreter/cling). Thanks!

PR #284, opened 2019-02-05
Refactor incremental CUDA PTX compiler

SimeonEhrig:
# About

Two major modifications to cling are required to jiting CUDA code. On the one hand, the host side (x86 target) needs some modifications to understand the cuda host code, distinguish the [host and device](https://devblogs.nvidia.com/easy-introduction-cuda-c-and-c/) code, and embed the device code (PTX target) into the host code. On the other hand, a second jit compiler needed to compile the device code next to the host code.

# Current implementation

Currently the PTX jit compiler (cuda device code) is implemented in  [IncrementalCUDADeviceCompiler.h](https://github.com/SimeonEhrig/cling/blob/master/lib/Interpreter/IncrementalCUDADeviceCompiler.h) and [IncrementalCUDADeviceCompiler.cpp](https://github.com/SimeonEhrig/cling/blob/master/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp). The basic idea is to use file I/O and external tools via `llvm::sys::ExecuteAndWait`. For example, if we enter `N` statements, we compile the PTX code with:

```bash
clang++ -std=c++xx -Ox -S cling0.cu cling1.cu ... clingN.cu -o cling.ptx --cuda-gpu-arch=sm_xx -pthread --cuda-device-only
```
When we add a new statement, we execute the following command:

```bash
clang++ -std=c++xx -Ox -S cling0.cu cling1.cu ... clingN.cu clingN+1.cu -o cling.ptx --cuda-gpu-arch=sm_xx -pthread --cuda-device-only
```
There is also a necessary step to translate the PTX code to fatbin code. At the moment this step is not refactorized because it requires proprietary software and there is no API provided.

After compiling the fatbin code, the code is read and embedded by the cling.

## Problems

The current implementation has a lot of problems.
1. **Generating valid CUDA C++ code:** The PTX compiler cannot directly compile the source code entered by the user, because cling allows invalid syntax, e.g. function calls in global space. The handling of variables has a similar problem. Cling itself transform the code internally to valid C++ code. The transformation is performed in the AST-Tree. At the moment we use the ASTPrinter to generate C++ code from the AST. However, the ASTPrinter is designed to print error messages and not to generate valid C++ code. So there are many individual bugs in the printer.
2. **Performance:** At the moment there is no possibility to save any progress of the translation. So we have to compile the whole entire code plus the new statement for each input, which we do. So the compile time constantly increasing.
3. **Handling Cling passes:** Cling performs some internal transformations to provide new functions, e.g. the `NullDerefProtectionTransformer`, which protects the cling runtime from crashes caused by `nullptr` differentiation. Some transformations cause some problems on the device side and are not necessary.  
4. **File-I/O:** File I/O is slow, especially on HPC-Systems.
5. **Dependencies on external tools:** Sometime, especially on HPC systems, it is not easy to find all necessary tools. Integrating the tools at build time avoid some problems during runtime.

# Idea

The main idea is to use an integrated incremental compiler instance which can process the invalid source code from the prompt and generate PTX code incremental without having to compile the entire TU every time. I have already implemented a [prototype](https://github.com/SimeonEhrig/IncrementalPTXJIT), which implemented this idea. The prototype use some code from cling via `libcling.so` and use some code from the llvm and clang libraries. 

## Prototype
The prototype is divided into three parts. The first part is a simple prompt similar to the cling prompt for testing. The second part is a modified version of the cling interpreter instance, which I use as frontend. Details will follow. The changes are visible in the pull request. The last part is a PTX backend implement which I implemented with an unmodified llvm library.

### Frontend

Three modifications were necessary for the frontend.

1. Extending the [InvocationOptions.h](https://github.com/SimeonEhrig/cling/commit/3d0b2d166e48a4023255c2d2c9792101269c9cfa) to decide between the mode `CUDAHost` and `CUDADevice`. 
 * CUDAHost: normal cling functionality, e.g. jit and execute code with some extensions for CUDA host functionality 
 * CUDADevice: incremental compiling of CUDA device code to ptx without code execution.
2. Deactivation of some ASTTransformer that are not needed.
3. Added a new ASTTransformer to add the attribute `inline` to each `__device__` kernel. See this [commit](https://github.com/SimeonEhrig/cling/commit/a37e5596ec9da4680903cb7459612bbb37595f90). This is necessary because the PTX code generating functionality of the clang is not designed to handle many TUs. For example, if we have a `__device__` kernel [2] and a `__global__` kernel, the PTX code of the `__device__` kernel contains the definition of the `__device__` kernel in the PTX code. The PTX file generated for the `__global__` kernel contains the PTX code of the `__global__` kernel and an external declaration of the `__device__` function. But external declarations are forbidden in PTX code. So I needed a solution to copy the `__device__` kernel definition into the `__global__` kernel PTX file. 

[2] `__device__` kernel are only visible to `__global__` and other `__device__` kernels and cannot be called by the host.

In my prototype I use the functionality of the interpreter instance up to the generation of the llvm ir code (`llvm::Module`). Then I take the `llvm::Module` and pass it to a extra backend.  

### Backend

The backend is based on some llvm tutorials and is really generic. It takes a `llvm::Module` and translates it to PTX. It works really well, witht one exception I solved with the ASTTransformer. I already tested the generated PTX code of some basic functions with a hacked cling.  

# Questions

1. **PTX modification of the interpreter instance:** With the generaton of PTX code, there are three possible modes for the interpreter instance (C++, CUDA C++ host and PTX). Perhaps it make sense to use an inheritance hierarchy of the class `Interpreter` for clean code. On the other hand, I am not sure, if the class `Interpreter` has been designed for inheritance. Maybe it might require some refactoring work and make future development more difficult. Do you have any good reasons or ideas as to what I should do?
**Current solution:** `cling::Interpreter` has a pointer to a second `cling::Interpreter` instance. Some `if (CudeDevice)` statements within the class provide the correct functionality. 
2. **Inline ASTTransformer:** I think, I can replace my custom function pass with a solution of llvm https://github.com/root-project/cling/commit/d713ae033e49fae9d34f88607a254033b8c4e8c0. I still have to check it out. Maybe there is another, better solution to link the `__device__` kernel of one TU to a `__global__` kernel of another TU. Do you have any ideas? I have already checked the llvm module linker, but I dont't think the functionality is right for my problem.
**Current solution:** Use a self-written `ASTTransformer`.
3. **NVPTX backend without file I/O:** The NVPTX backend uses the function `llvm::MemoryBuffer::getFileOrSTDIN()` to load the fatbin code from file (see [source code](https://github.com/llvm/llvm-project/blob/9487963244e4c805cf0f5798d903bdb10012b59d/clang/lib/CodeGen/CGCUDANV.cpp#L501-L502) ). The path is set by the Cling. To get a full compiler stack without file I/O, we must avoid this step. Maybe we can use the virtual file system of clang or llvm. Or we try something with the STDIN. But I think that could cause a lot of problems. Otherwise, we have to change the clang source code so that it reads the code direct from a buffer.
**Current solution:** The clang code has been modified because `llvm::MemoryBuffer::getFileOrSTDIN()` does not support a virtual file system. 

# Tasks befor merge
These functions must be implemented so that the pull request is ready for merging.

- [x] Load fatbin code in NVPTX backend without file I/O (see Question 3.)
- [x] Add functionality at `cling::utils::getWrapPoint()` to handle the cuda function attributes `__global__`, `__device__` and `__host__`
- [ ] ~~fix bug with CUDA `__constant__` memory~~
- [x] fix bug with xeus-cling I/O

# Optional task
I moved the `__constant__` memory task to an optional task, because the solution more complicated, than I thought. `__constant__` is a frequently used feature, but is not required for the basic functionality of the CUDA extension.
- [ ] fix bug with CUDA `__constant__` memory

**Problem description:** https://github.com/SimeonEhrig/cling/issues/3
**Possible solution 1:** Use an ASTTransformer to remove the `__constant__` attribute from local variables. Is not easy to use because attributes are not stored in the AST. Attributes are stored in a side data structure (clang::Declarator). To check whether a statement has an attribute, we must perform the semantic analysis, which also checks properties of attributes which cause the `__constant__`-local-error. The built-in functions of clang for detecting and changing attributes should therefore not be usable. This means that we have to implement our own solution for detecting and changing attributes.     
**Possible solution 2:** Change the behavior of the ASTTransformer `DeclExtractor`. At the moment the `DeclExtractior` copies the declaration form the wrapper function to the global space and modify the local declaration a bit. If we could delete the local declaration completely, the error would not occur. But I don't know if the local declaration is necessary for any functionality.

# Overview of the most important changes

* modified `cling::Interpreter`
  * can be used as nvptx JIT
  * has a pointer to a second `cling::interpreter` instance -> host JIT has a device JIT
  * to avoid recursions and some errors in the device JIT, there are some `if(CudaDevice)` statements in the functions
* a new ASTTransformer that inline every cuda `__device__` kernel
  * is required because PTX is a plain assembler format without linking functionality
  * is needed for example, if we define a `__device__` kernel in transaction A and we want to use the `__device__` kernel in a `__global__` kernel of transaction B -> then we have to copy to whole definition of the `__device__` kernel into the PTX file of transaction B
* PTX and fatbinary are generated entirely in-memory
  * for PTX see modifications at `cling::Interpreter`
  * for the fatbinary code we need a modification at the `Clang` code, because, the Clang codeGen read the code via a function from a file that does not support a virtual file system
* CPT tool builds the nvptx beside the host target
  * avoid a lot of `#ifdef` guards
  * build is completly target independent
  * if the functionality is not used at runtime, it is also completly target independent
* Add the detection of CUDA attributes `__global__`, `__device__` and `__host__` to `SourceNormalization`
  * a general, intelligent approach would be better

### DIFF
--- a/lib/Interpreter/DeviceKernelInliner.cpp
+++ b/lib/Interpreter/DeviceKernelInliner.cpp
@@ -1,0 +1,30 @@
+//--------------------------------------------------------------------*- C++ -*-
+// CLING - the C++ LLVM-based InterpreterG :)
+// author:  Simeon Ehrig <s.ehrig@hzdr.de>
+//
+// This file is dual-licensed: you can choose to license it under the University
+// of Illinois Open Source License or the GNU Lesser General Public License. See
+// LICENSE.TXT for details.
+//------------------------------------------------------------------------------
+
+#include "DeviceKernelInliner.h"
+
+#include <clang/AST/Attr.h>
+
+#include <llvm/Support/Casting.h>
+
+namespace cling {
+
+DeviceKernelInliner::DeviceKernelInliner(clang::Sema *S) : ASTTransformer(S) {}
+
+ASTTransformer::Result DeviceKernelInliner::Transform(clang::Decl *D) {
+  clang::FunctionDecl *F = llvm::dyn_cast<clang::FunctionDecl>(D);
+  if (F && F->hasAttr<clang::CUDADeviceAttr>()) {
+    if (!F->isInlineSpecified()) {
+      F->setInlineSpecified(true);
+    }
+  }
+  return Result(D, true);
+}
+
+} // namespace cling
--- a/lib/Interpreter/DeviceKernelInliner.cpp
+++ b/lib/Interpreter/DeviceKernelInliner.cpp
@@ -18,9 +18,8 @@ namespace cling {
 DeviceKernelInliner::DeviceKernelInliner(clang::Sema *S) : ASTTransformer(S) {}
 
 ASTTransformer::Result DeviceKernelInliner::Transform(clang::Decl *D) {
-  clang::FunctionDecl *F = llvm::dyn_cast<clang::FunctionDecl>(D);
-  if (F && F->hasAttr<clang::CUDADeviceAttr>()) {
-    if (!F->isInlineSpecified()) {
+  if (clang::FunctionDecl* F = llvm::dyn_cast<clang::FunctionDecl>(D)) {
+    if (F->hasAttr<clang::CUDADeviceAttr>()) {
       F->setInlineSpecified(true);
     }
   }
--- a/lib/Interpreter/DeviceKernelInliner.h
+++ b/lib/Interpreter/DeviceKernelInliner.h
@@ -7,8 +7,8 @@
 // LICENSE.TXT for details.
 //------------------------------------------------------------------------------
 
-#ifndef DEVICEKERNELINLINER_H
-#define DEVICEKERNELINLINER_H
+#ifndef CLING_DEVICEKERNELINLINER_H
+#define CLING_DEVICEKERNELINLINER_H
 
 #include "ASTTransformer.h"
 
@@ -30,11 +30,11 @@ namespace cling
 class DeviceKernelInliner : public ASTTransformer
 {
 public:
-    DeviceKernelInliner ( clang::Sema* S );
+    DeviceKernelInliner(clang::Sema* S);
 
-    ASTTransformer::Result Transform ( clang::Decl * D ) override;
+    ASTTransformer::Result Transform(clang::Decl * D) override;
 };
 
 }
 
-#endif // DEVICEKERNELINLINER_H
+#endif // CLING_DEVICEKERNELINLINER_H
--- a/lib/Interpreter/DeviceKernelInliner.h
+++ b/lib/Interpreter/DeviceKernelInliner.h
@@ -24,12 +24,11 @@ namespace cling
 // a __global__ kernel uses a __device__ function, this design caused an error.
 // Instead of generating the PTX code of the __device__ kernel in the same file
 // as the __global__ kernel, there is only an external declaration of the
-// __device__ function. However, normal PTX code does not support an external
-// declaration of functions.
+// __device__ function. However, PTX does not support an external declaration of
+// functions.
 
 class DeviceKernelInliner : public ASTTransformer
 {
-
 public:
     DeviceKernelInliner ( clang::Sema* S );
 
--- a/lib/Interpreter/DeviceKernelInliner.h
+++ b/lib/Interpreter/DeviceKernelInliner.h
@@ -1,0 +1,41 @@
+//--------------------------------------------------------------------*- C++ -*-
+// CLING - the C++ LLVM-based InterpreterG :)
+// author:  Simeon Ehrig <s.ehrig@hzdr.de>
+//
+// This file is dual-licensed: you can choose to license it under the University
+// of Illinois Open Source License or the GNU Lesser General Public License. See
+// LICENSE.TXT for details.
+//------------------------------------------------------------------------------
+
+#ifndef DEVICEKERNELINLINER_H
+#define DEVICEKERNELINLINER_H
+
+#include "ASTTransformer.h"
+
+
+namespace cling
+{
+
+// This ASTTransformer adds an inline attribute to any CUDA __device__ kernel
+// that does not have the attribute. Inlining solves a problem caused by
+// incremental compilation of PTX code. In a normal compiler, all definitions
+// of __global__ and __device__ kernels are in the same translation unit. In
+// the incremental compiler, each kernel has its own translation unit. In case
+// a __global__ kernel uses a __device__ function, this design caused an error.
+// Instead of generating the PTX code of the __device__ kernel in the same file
+// as the __global__ kernel, there is only an external declaration of the
+// __device__ function. However, normal PTX code does not support an external
+// declaration of functions.
+
+class DeviceKernelInliner : public ASTTransformer
+{
+
+public:
+    DeviceKernelInliner ( clang::Sema* S );
+
+    ASTTransformer::Result Transform ( clang::Decl * D ) override;
+};
+
+}
+
+#endif // DEVICEKERNELINLINER_H
--- a/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
+++ b/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
@@ -46,7 +46,7 @@ namespace cling {
       return;
     }
 
-    setCuArgs(CI.getLangOpts(), invocationOptions, optLevel,
+    setCuArgs(CI.getLangOpts(), invocationOptions,
               CI.getCodeGenOpts().getDebugInfo(),
               llvm::Triple(CI.getTargetOpts().Triple));
 
@@ -55,7 +55,7 @@ namespace cling {
     std::vector<std::string> argv = {
         "cling",
         m_CuArgs->cppStdVersion.c_str(),
-        m_CuArgs->optLevel.c_str(),
+        "-O" + std::to_string(optLevel),
         "-x",
         "cuda",
         "-S",
@@ -108,7 +108,6 @@ namespace cling {
   void IncrementalCUDADeviceCompiler::setCuArgs(
       const clang::LangOptions& langOpts,
       const cling::InvocationOptions& invocationOptions,
-      const int intprOptLevel,
       const clang::codegenoptions::DebugInfoKind debugInfo,
       const llvm::Triple hostTriple) {
     std::string cppStdVersion;
@@ -125,8 +124,6 @@ namespace cling {
     if (cppStdVersion.empty())
       llvm::errs()
           << "IncrementalCUDADeviceCompiler: No valid c++ standard is set.\n";
-
-    const std::string optLevel = "-O" + std::to_string(intprOptLevel);
 
     uint32_t smVersion = 20;
     if (!invocationOptions.CompilerOpts.CUDAGpuArch.empty()) {
@@ -186,13 +183,13 @@ namespace cling {
       fatbinFlags |= FatBinFlags::HostLinux;
 
     m_CuArgs.reset(new IncrementalCUDADeviceCompiler::CUDACompilerArgs(
-        cppStdVersion, optLevel, hostTriple, smVersion, fatbinFlags,
+        cppStdVersion, hostTriple, smVersion, fatbinFlags,
         invocationOptions.Verbose(), debug, additionalPtxOpt));
   }
 
   void IncrementalCUDADeviceCompiler::addHeaderSearchPathFlags(
       std::vector<std::string>& argv,
-      const std::shared_ptr<clang::HeaderSearchOptions> headerSearchOptions) {
+      const std::shared_ptr<clang::HeaderSearchOptions> &headerSearchOptions) {
     for (clang::HeaderSearchOptions::Entry e :
          headerSearchOptions->UserEntries) {
       if (e.Group == clang::frontend::IncludeDirGroup::Quoted) {
@@ -208,7 +205,7 @@ namespace cling {
   // FIXME: add the same arguments as the cling::Interpreter class -> need some
   // modifications in the cling::Transaction class to store information from the
   // device compiler
-  bool IncrementalCUDADeviceCompiler::process(const llvm::StringRef input) {
+  bool IncrementalCUDADeviceCompiler::process(const std::string& input) {
     if (!m_Init) {
       llvm::errs()
           << "Error: Initializiation of CUDA Device Code Compiler failed\n";
@@ -234,7 +231,7 @@ namespace cling {
   }
 
   // FIXME: see process()
-  bool IncrementalCUDADeviceCompiler::declare(const llvm::StringRef input) {
+  bool IncrementalCUDADeviceCompiler::declare(const std::string& input) {
     if (!m_Init) {
       llvm::errs()
           << "Error: Initializiation of CUDA Device Code Compiler failed\n";
@@ -333,8 +330,10 @@ namespace cling {
       return false;
     }
 
-    // implementation ist copied from clangJIT
-    // (https://github.com/hfinkel/llvm-project-cxxjit/)
+    // implementation is adapted from clangJIT
+    // (https://github.com/hfinkel/llvm-project-cxxjit/blob/cxxjit/clang/lib/CodeGen/JIT.cpp)
+    // void *resolveFunction(const void *NTTPValues, const char **TypeStrings,
+    //                       unsigned Idx)
 
     // The outer header of the fat binary is documented in the CUDA
     // fatbinary.h header. As mentioned there, the overall size must be a
@@ -410,7 +409,6 @@ namespace cling {
                  << "file path: " << m_FilePath << "\n"
                  << "fatbin file path: " << m_FatbinFilePath << "\n"
                  << "m_CuArgs c++ standard: " << m_CuArgs->cppStdVersion << "\n"
-                 << "m_CuArgs opt level: " << m_CuArgs->optLevel << "\n"
                  << "m_CuArgs host triple: " << m_CuArgs->hostTriple.str()
                  << "\n"
                  << "m_CuArgs Nvidia SM Version: " << m_CuArgs->smVersion
--- a/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
+++ b/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
@@ -72,6 +72,13 @@ namespace cling {
     argv.insert(argv.end(), m_CuArgs->additionalPtxOpt.begin(),
                 m_CuArgs->additionalPtxOpt.end());
 
+    // add included files to the cling ptx
+    for (const char* c : invocationOptions.CompilerOpts.Remaining) {
+      std::string s(c);
+      if (s.find("-include") == 0)
+        argv.push_back(s);
+    }
+
     std::vector<const char*> argvChar;
     argvChar.resize(argv.size() + 1);
 
--- a/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
+++ b/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
@@ -234,7 +234,6 @@ namespace cling {
   }
 
   bool cling::IncrementalCUDADeviceCompiler::generatePTX() {
-
     std::error_code EC;
     llvm::raw_fd_ostream os(m_PTXFilePath, EC, llvm::sys::fs::F_None);
     if (EC) {
--- a/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
+++ b/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
@@ -9,21 +9,21 @@
 
 #include "IncrementalCUDADeviceCompiler.h"
 
+#include "cling/Interpreter/Interpreter.h"
 #include "cling/Interpreter/InvocationOptions.h"
 #include "cling/Interpreter/Transaction.h"
-#include "cling/Utils/Paths.h"
 
-#include "clang/AST/Decl.h"
-#include "clang/AST/DeclGroup.h"
 #include "clang/Frontend/CompilerInstance.h"
 #include "clang/Lex/HeaderSearchOptions.h"
 
-#include "llvm/ADT/Triple.h"
-#include "llvm/Support/FileSystem.h"
-#include "llvm/Support/Path.h"
 #include "llvm/Support/Process.h"
 #include "llvm/Support/Program.h"
 #include "llvm/Support/raw_ostream.h"
+#include <llvm/IR/LegacyPassManager.h>
+#include <llvm/Support/TargetRegistry.h>
+#include <llvm/Support/TargetSelect.h>
+#include <llvm/Target/TargetMachine.h>
+#include <llvm/Target/TargetOptions.h>
 
 #include <algorithm>
 #include <string>
@@ -47,46 +47,54 @@ namespace cling {
         m_FatbinFilePath(CI.getCodeGenOpts().CudaGpuBinaryFileNames.empty()
                              ? ""
                              : CI.getCodeGenOpts().CudaGpuBinaryFileNames[0]),
-        m_DummyCUPath(m_FilePath + "dummy.cu"),
-        m_PTXFilePath(m_FilePath + "cling.ptx"),
-        m_GenericFileName(m_FilePath + "cling") {
+        m_PTXFilePath(m_FilePath + "cling.ptx") {
     if (m_FatbinFilePath.empty()) {
       llvm::errs() << "Error: CudaGpuBinaryFileNames can't be empty\n";
       return;
     }
 
-    if (!generateHelperFiles()) return;
     if (!findToolchain(invocationOptions)) return;
     setCuArgs(CI.getLangOpts(), invocationOptions, optLevel,
               CI.getCodeGenOpts().getDebugInfo());
 
-    m_HeaderSearchOptions = CI.getHeaderSearchOptsPtr();
+    // cling -std=c++xx -Ox -x cuda -S --cuda-gpu-arch=sm_xx --cuda-device-only
+    // ${include headers} ${-I/paths} [-v] [-g] ${m_CuArgs->additionalPtxOpt}
+    std::vector<std::string> argv = {"cling",
+                                     m_CuArgs->cppStdVersion.c_str(),
+                                     m_CuArgs->optLevel.c_str(),
+                                     "-x",
+                                     "cuda",
+                                     "-S",
+                                     m_CuArgs->ptxSmVersion.c_str(),
+                                     "--cuda-device-only"};
+    addHeaderSearchPathFlags(argv, CI.getHeaderSearchOptsPtr());
+    if (m_CuArgs->verbose) argv.push_back("-v");
+    if (m_CuArgs->debug) argv.push_back("-g");
+    argv.insert(argv.end(), m_CuArgs->additionalPtxOpt.begin(),
+                m_CuArgs->additionalPtxOpt.end());
 
-    // This code will write to the first .cu-file. It is necessary that some
-    // cling generated code can be handled.
-    const std::string initialCUDADeviceCode =
-        "extern void setValueNoAlloc(void* vpI, void* vpV, void* vpQT, char "
-        "vpOn, float value);\n"
-        "extern void setValueNoAlloc(void* vpI, void* vpV, void* vpQT, char "
-        "vpOn, double value);\n"
-        "extern void setValueNoAlloc(void* vpI, void* vpV, void* vpQT, char "
-        "vpOn, long double value);\n"
-        "extern void setValueNoAlloc(void* vpI, void* vpV, void* vpQT, char "
-        "vpOn, unsigned long long value);\n"
-        "extern void setValueNoAlloc(void* vpI, void* vpV, void* vpQT, char "
-        "vpOn, const void* value);\n";
+    std::vector<const char*> argvChar;
+    argvChar.resize(argv.size() + 1);
 
-    std::error_code EC;
-    llvm::raw_fd_ostream cuFile(m_FilePath + "cling0.cu", EC,
-                                llvm::sys::fs::F_Text);
-    if (EC) {
-      llvm::errs() << "Could not open " << m_FilePath << "cling0.cu"
-                   << EC.message() << "\n";
+    std::transform(argv.begin(), argv.end(), argvChar.begin(),
+                   [&](const std::string& s) { return s.c_str(); });
+
+    // argv list have to finish with a nullptr.
+    argvChar.push_back(nullptr);
+
+    // create incremental compiler instance
+    m_PTX_interp.reset(new Interpreter(argvChar.size(), argvChar.data()));
+
+    if (!m_PTX_interp) {
+      llvm::errs() << "Could not create PTX interpreter instance\n";
       return;
     }
 
-    cuFile << initialCUDADeviceCode;
-    cuFile.close();
+    // initialize NVPTX backend
+    LLVMInitializeNVPTXTargetInfo();
+    LLVMInitializeNVPTXTarget();
+    LLVMInitializeNVPTXTargetMC();
+    LLVMInitializeNVPTXAsmPrinter();
 
     m_Init = true;
   }
@@ -110,9 +118,11 @@ namespace cling {
 
     const std::string optLevel = "-O" + std::to_string(intprOptLevel);
 
+    std::string smVersion = "sm_20";
     std::string ptxSmVersion = "--cuda-gpu-arch=sm_20";
     std::string fatbinSmVersion = "--image=profile=compute_20";
     if (!invocationOptions.CompilerOpts.CUDAGpuArch.empty()) {
+      smVersion = invocationOptions.CompilerOpts.CUDAGpuArch;
       ptxSmVersion =
           "--cuda-gpu-arch=" + invocationOptions.CompilerOpts.CUDAGpuArch;
       fatbinSmVersion = "--image=profile=compute_" +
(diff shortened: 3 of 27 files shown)


lib/Interpreter/IncrementalParser.cpp
```diff
@@ -908,24 +910,31 @@ namespace cling {
   void IncrementalParser::SetTransformers(bool isChildInterpreter) {
     // Add transformers to the IncrementalParser, which owns them
     Sema* TheSema = &m_CI->getSema();
+    // if the interpreter compiles ptx code, some transformers should not used
```
ax3l:
```suggestion
    // if the interpreter compiles ptx code, some transformers should not be used
```

test/Interfaces/invocationFlags.C
```diff
@@ -80,5 +84,18 @@ IOpts.CompilerOpts.Remaining
 // Windows translates -nostdinc++ to -nostdinc++. Ignore that fact for the test.
 // CHECK-NEXT: {{.*}} { "progname", "-xcuda", "FileToExecuteA", "-isysroot", "APAth", {{.*}}, "-v", "FileToExecuteB" }
 
+// this flag allows to compile ptx code with the interpreter instance
+// CUDAHost and CUDADevice must not be true at the same time
+// if --cuda-device-only is set, it isn't important if -xcuda is set
```
ax3l: Maybe that's easier to read:
```suggestion
// --cuda-device-only implies -xcuda
```

lib/Interpreter/DeviceKernelInliner.h
```diff
@@ -0,0 +1,41 @@
...
+
+// This ASTTransformer adds an inline attribute to any CUDA __device__ kernel
+// that does not have the attribute. Inlining solves a problem caused by
+// incremental compilation of PTX code. In a normal compiler, all definitions
+// of __global__ and __device__ kernels are in the same translation unit. In
+// the incremental compiler, each kernel has its own translation unit. In case
+// a __global__ kernel uses a __device__ function, this design caused an error.
+// Instead of generating the PTX code of the __device__ kernel in the same file
+// as the __global__ kernel, there is only an external declaration of the
+// __device__ function. However, normal PTX code does not support an external
```
ax3l:
```suggestion
// __device__ function. However, PTX does not support an external
```

lib/Interpreter/DeviceKernelInliner.h
```diff
@@ -0,0 +1,41 @@
...
+// declaration of functions.
+
+class DeviceKernelInliner : public ASTTransformer
+{
+
```
ax3l: CI wants you to remove this blank line :)

lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
```diff
@@ -47,46 +47,54 @@ namespace cling {
...
+
+    if (!m_PTX_interp) {
+      llvm::errs() << "Could not create PTX interpreter instance\n";
       return;
     }
 
-    cuFile << initialCUDADeviceCode;
-    cuFile.close();
+    // initialize NVPTX backend
+    LLVMInitializeNVPTXTargetInfo();
```
ax3l: CI says:
```
error: ‘LLVMInitializeNVPTXTargetInfo’ was not declared in this scope
     LLVMInitializeNVPTXTargetInfo();
error: ‘LLVMInitializeNVPTXTarget’ was not declared in this scope
     LLVMInitializeNVPTXTarget();
error: ‘LLVMInitializeNVPTXTargetMC’ was not declared in this scope
     LLVMInitializeNVPTXTargetMC();
error: ‘LLVMInitializeNVPTXAsmPrinter’ was not declared in this scope
     LLVMInitializeNVPTXAsmPrinter();
```

missing includes?

lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
```diff
@@ -224,216 +204,74 @@ namespace cling {
...
-    std::transform(argv.begin(), argv.end(), argvChar.begin(),
-                   [&](const std::string& s) { return s.c_str(); });
-
-    // Argv list have to finish with a nullptr.
-    argvChar.push_back(nullptr);
-
-    std::string executionError;
-    int res = llvm::sys::ExecuteAndWait(m_ClangPath.c_str(), argvChar.data(),
-                                        nullptr, {}, 0, 0, &executionError);
+  bool cling::IncrementalCUDADeviceCompiler::generatePTX() {
```
ax3l: CI wants you to remove this empty newline at the beginning of a function :)

lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
SimeonEhrig: There is a problem with the Travis CI configuration. Travis builds cling with the cmake argument `-DLLVM_TARGETS_TO_BUILD=host`. So only the x86 backend is built.

ax3l: But that's ok to do. Maybe you want to add guards for the build target?

SimeonEhrig: @Axel-Naumann Do you have a better solution? A `#ifdef` guard is just a workaround right now. It's just avoids the test build with the new functionality and that's not a good solutions. I think enabling the nvptx backend is the better solution. It should no affect the build on systems that don't provide a CUDA runtime.

ax3l:
CC @Axel-Naumann @vgvassilev ping :) Please feel free to provide feedback to Simeon's new approach.

lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
SimeonEhrig: I solved the problem by adding the `NVPTX` backend to `-DLLVM_TARGETS_TO_BUILD=host`. It does not cause any problems on non CUDA system. It just increases the size a little.

SimeonEhrig:
> NVPTX backend without file I/O: The NVPTX backend uses the function llvm::MemoryBuffer::getFileOrSTDIN() to load the fatbin code from file (see source code ). The path is set by the Cling. To get a full compiler stack without file I/O, we must avoid this step. Maybe we can use the virtual file system of clang or llvm. Or we try something with the STDIN. But I think that could cause a lot of problems. Otherwise, we have to change the clang source code so that it reads the code direct from a buffer.

@Axel-Naumann @vgvassilev 
Do you have any idea how I can solve this? I have seen that you have already done something with virtual file systems in cling. There is also a [post](http://lists.llvm.org/pipermail/llvm-dev/2019-June/132964.html) on the llvm-dev list.

SimeonEhrig:
For the commit 94075f969 is a clang patch needed: 
[0001-Add-buffer-for-CUDA-fatbinary-in-CGCUDANV-backend.patch.zip](https://github.com/root-project/cling/files/3405544/0001-Add-buffer-for-CUDA-fatbinary-in-CGCUDANV-backend.patch.zip)

ax3l:
@SimeonEhrig can you quickly solve the merge conflict, please?

SimeonEhrig:
I have successfully rebased my branch. The tests are passed. The CI doesn't have the Clang Patch, so they will fail.

Unfortunately, redefinition with CUDA doesn't work at the moment. I will try to solve it because the feature is very nice.

lib/Interpreter/IncrementalCUDADeviceCompiler.h
```diff
@@ -35,120 +36,78 @@ namespace llvm {
...
-    void addHeaderSearchPathFlags(llvm::SmallVectorImpl<std::string>& argv);
-
-    ///\brief Start an clang compiler with nvptx backend. Read the content of
-    /// cling.cu and compile it to a new PCH file. If predecessor PCH file is
-    /// existing, it will included.
+    ///\param [in] headerSearchOptions - Contains information about all include
+    /// paths.
+    void addHeaderSearchPathFlags(
+        std::vector<std::string>& argv,
+        const std::shared_ptr<clang::HeaderSearchOptions> headerSearchOptions);
```
Axel-Naumann: I think taking a `const std::shared_ptr<clang::HeaderSearchOptions> &headerSearchOptions` might make more sense - doesn't incr/decr the use count just for arg passing.

lib/Interpreter/IncrementalCUDADeviceCompiler.h
```diff
@@ -157,49 +116,78 @@ namespace cling {
...
     void setCuArgs(const clang::LangOptions& langOpts,
                    const cling::InvocationOptions& invocationOptions,
                    const int intprOptLevel,
-                   const clang::codegenoptions::DebugInfoKind debugInfo);
-
-    ///\brief Save .cu file, if cuda device code compiler failed at translation.
-    ///
-    ///\returns The error code of the rename operation
-    std::error_code saveFaultyCUfile();
+                   const clang::codegenoptions::DebugInfoKind debugInfo,
```
Axel-Naumann: Why are these new parameters (and the `const int intprOptLevel` before) marked `const`?

lib/Interpreter/IncrementalCUDADeviceCompiler.h
```diff
@@ -157,49 +116,78 @@ namespace cling {
...
     ///       The value will be copied, because a change of it is not allowed.
     ///\param [in] invocationOptions - Contains values for the arguments of
-    ///       clang and the NVIDIA tool fatbinary.
+    ///       the interpreter instance and the NVIDIA tool fatbinary.
     ///\param [in] CI - Will be used for m_CuArgs and the include path handling.
     IncrementalCUDADeviceCompiler(
-        const std::string& filePath, const int optLevel,
-        const cling::InvocationOptions& invocationOptions,
+        std::shared_ptr<llvm::SmallVectorImpl<char>> fatbinBuffer,
+        const int optLevel, const cling::InvocationOptions& invocationOptions,
```
Axel-Naumann: Why `const int optLevel` instead of `int optLevel`?

lib/Interpreter/IncrementalCUDADeviceCompiler.h
```diff
@@ -157,49 +116,78 @@ namespace cling {
...
         const clang::CompilerInstance& CI);
 
+    ///\brief Returns a reference to the PTX interpreter
+    ///
+    ///\return std::unique_ptr< cling::Interpreter >&
+    ///
+    std::unique_ptr<Interpreter>& getPTXInterpreter() { return m_PTX_interp; }
```
Axel-Naumann: Why return the `unique_ptr`-ness of the interpreter? This allows the caller to do `getPTXInterpreter().resetI()`, and basically explicitly so (i.e. as an expected interface usage) because the returned `std::unique_ptr<Interpreter>&` is non-const. Suggest:
```suggestion
    Interpreter &getPTXInterpreter() { return *m_PTX_interp; }
```
if `m_PTX_interp` is guaranteed `!= nullptr`, else
```suggestion
    Interpreter *getPTXInterpreter() { return m_PTX_interp.get(); }
```

lib/Interpreter/IncrementalCUDADeviceCompiler.h
```diff
@@ -157,49 +116,78 @@ namespace cling {
...
+    /// input is going to be processed. Otherwise if is known, for example
+    /// only header files are going to be processed it is much faster to run the
+    /// specific interface for doing that - in the particular case - declare().
+    ///
+    ///\param [in] input - The input directly from the UI. Attention, the string
+    /// must not be wrapped or transformed.
+    ///
+    ///\returns true, if all stages of generating fatbin runs right and a new
+    /// fatbin file is written.
+    bool process(const llvm::StringRef input);
```
Axel-Naumann: Why is the parameter `const`? Same for the others occurrences in other functions, where parameters are passed by value.

lib/Interpreter/Interpreter.cpp
```diff
@@ -240,6 +241,20 @@ namespace cling {
     if (m_OptLevel == -1)
       setDefaultOptLevel(getCI()->getCodeGenOpts().OptimizationLevel);
 
+    if (!isInSyntaxOnlyMode() && m_Opts.CompilerOpts.CUDAHost) {
+      // Using a buffer instead of file I/O in the CUDA host backend is enabled
+      // by a valid pointer.
```
Axel-Naumann: "valid pointer" where? To what? which pointer? Suggest:
```suggestion
      // by getCI()->getCodeGenOpts().CudaGpuBinaryBuffer being set.
```
if that is what you mean.

lib/Interpreter/Interpreter.cpp
```diff
@@ -454,6 +469,15 @@ namespace cling {
 
   Transaction* Interpreter::Initialize(bool NoRuntime, bool SyntaxOnly,
                               llvm::SmallVectorImpl<llvm::StringRef>& Globals) {
+    // initialization is done by the host interpreter
```
Axel-Naumann: Possibly better:
```suggestion
    // Initialize the host interpreter to run host code, and pass device code to the device compiler using its `declare()` function
```
but not sure that's what you try to say here?

lib/Interpreter/Interpreter.cpp
```diff
@@ -776,6 +803,9 @@ namespace cling {
   Interpreter::process(const std::string& input, Value* V /* = 0 */,
                        Transaction** T /* = 0 */,
                        bool disableValuePrinting /* = false*/) {
+    if (!isInSyntaxOnlyMode() && m_Opts.CompilerOpts.CUDAHost)
```
Axel-Naumann: Would it be better to make these `virtual`, and have a derived `Interpreter`for the CUDAHost case?

test/CUDADeviceCode/CUDAInclude.C
```diff
@@ -0,0 +1,45 @@
+//------------------------------------------------------------------------------
+// CLING - the C++ LLVM-based InterpreterG :)
+//
```
Axel-Naumann: You're welcome to add ` // author:  Simeon Ehrig <s.ehrig@hzdr.de> ` also for tests.

tools/packaging/cpt.py
```diff
@@ -494,7 +494,7 @@ def compile(arg, build_libcpp):
 
     build = Build()
     cmake_config_flags = (srcdir + ' -DLLVM_BUILD_TOOLS=Off -DCMAKE_BUILD_TYPE={0} -DCMAKE_INSTALL_PREFIX={1} '
-                          .format(build.buildType, TMP_PREFIX) + ' -DLLVM_TARGETS_TO_BUILD=host ' +
+                          .format(build.buildType, TMP_PREFIX) + ' -DLLVM_TARGETS_TO_BUILD="host;NVPTX" ' +
```
Axel-Naumann: I'd prefer doing that at the `CMakeLists.txt` level of cling, given that - if not using cpt.py - people must *know* to add `NVPTX`. Can cling add `NVPTX` as `LLVM_TARGETS_TO_BUILD`?

lib/Interpreter/DeviceKernelInliner.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include <llvm/Support/Casting.h>
+
+namespace cling {
+
+DeviceKernelInliner::DeviceKernelInliner(clang::Sema *S) : ASTTransformer(S) {}
+
+ASTTransformer::Result DeviceKernelInliner::Transform(clang::Decl *D) {
+  clang::FunctionDecl *F = llvm::dyn_cast<clang::FunctionDecl>(D);
+  if (F && F->hasAttr<clang::CUDADeviceAttr>()) {
+    if (!F->isInlineSpecified()) {
```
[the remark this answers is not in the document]
SimeonEhrig: Should this be a performance/storage optimization or am I mistaken?

I have tested it and it works. So it's fine for me. I am also adding a second optimization. I have removed the check `!F->isInlineSpecified()` because the result is always the same. The variables `IsInlineSpecified` and `IsInline` of the function declaration are `true`. Setting a bool variable should be faster than an if statement.

lib/Interpreter/Interpreter.cpp
SimeonEhrig: Yes, it would be useful to use `virtual` to improve code quality. I think a derived class would make it easier to see what changes are necessary for the CUDA mode.

I used the "if" solution because I wasn't sure if the `virtual` statement could negatively affect performance, but I see that the destructor is already `virtual`. So this is not a problem and I see no further disadvantage. 

I will check a possible solution in a future PR.

lib/Interpreter/DeviceKernelInliner.cpp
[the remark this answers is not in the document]
Axel-Naumann: > Should this be a performance/storage optimization or am I mistaken?

No, this is about keeping variables at the closes scope possible, by moving the declaration of `F` into the `if` condition.

lib/Interpreter/Interpreter.cpp
Axel-Naumann: Thanks, for me it's fine to return the `Transaction`, please go ahead.

lib/Interpreter/Interpreter.cpp
Axel-Naumann: Good arguments - thanks, I'm counting on you here :-)

Axel-Naumann:
Any idea why Travis is unhappy: https://travis-ci.org/root-project/cling/jobs/592618226#L2146 ?

SimeonEhrig:
Yes, the Clang patch is missing in the CI.

See: https://github.com/root-project/cling/pull/284#issuecomment-512720551

Axel-Naumann:
Sorry I missed that. I'm extremely hesitant to add clang patches that are not also upstreamed. Could you open a Differential with the clang folks, arguing your case?

SimeonEhrig:
I know that clang patches are not the best solution, but I haven't found another solution. I've already checked if I can use the llvm virtual file system. Unfortunately, the `getFileOrSTDIN` function cannot use the virtual filesystem. But maybe you have another idea when you see the differential:

```diff
diff --git a/include/clang/Frontend/CodeGenOptions.h b/include/clang/Frontend/CodeGenOptions.h
index 71730a2..9e2cf68 100644
--- a/include/clang/Frontend/CodeGenOptions.h
+++ b/include/clang/Frontend/CodeGenOptions.h
@@ -209,6 +209,10 @@ public:
   /// object file.
   std::vector<std::string> CudaGpuBinaryFileNames;
 
+  /// A buffer that contains the fatbinary code to forward to CUDA runtime
+  /// back-end for incorporating them into host-side object file.
+  std::shared_ptr<llvm::SmallVectorImpl<char>> CudaGpuBinaryBuffer;
+
   /// The name of the file to which the backend should save YAML optimization
   /// records.
   std::string OptRecordFile;
diff --git a/lib/CodeGen/CGCUDANV.cpp b/lib/CodeGen/CGCUDANV.cpp
index d24ef0a..66d6f63 100644
--- a/lib/CodeGen/CGCUDANV.cpp
+++ b/lib/CodeGen/CGCUDANV.cpp
@@ -254,7 +254,8 @@ llvm::Function *CGNVCUDARuntime::makeRegisterGlobalsFn() {
 /// \endcode
 llvm::Function *CGNVCUDARuntime::makeModuleCtorFunction() {
   // No need to generate ctors/dtors if there are no GPU binaries.
-  if (CGM.getCodeGenOpts().CudaGpuBinaryFileNames.empty())
+  if (!CGM.getCodeGenOpts().CudaGpuBinaryBuffer &&
+      CGM.getCodeGenOpts().CudaGpuBinaryFileNames.empty())
     return nullptr;
 
   // void __cuda_register_globals(void* handle);
@@ -281,16 +282,7 @@ llvm::Function *CGNVCUDARuntime::makeModuleCtorFunction() {
   // to be cleaned up in destructor on exit. Then associate all known kernels
   // with the GPU binary handle so CUDA runtime can figure out what to call on
   // the GPU side.
-  for (const std::string &GpuBinaryFileName :
-       CGM.getCodeGenOpts().CudaGpuBinaryFileNames) {
-    llvm::ErrorOr<std::unique_ptr<llvm::MemoryBuffer>> GpuBinaryOrErr =
-        llvm::MemoryBuffer::getFileOrSTDIN(GpuBinaryFileName);
-    if (std::error_code EC = GpuBinaryOrErr.getError()) {
-      CGM.getDiags().Report(diag::err_cannot_open_file) << GpuBinaryFileName
-                                                        << EC.message();
-      continue;
-    }
-
+  auto buildFatbinarySection = [&](const llvm::StringRef FatbinCode) {
     const char *FatbinConstantName =
         CGM.getTriple().isMacOSX() ? "__NV_CUDA,__nv_fatbin" : ".nv_fatbin";
     // NVIDIA's cuobjdump looks for fatbins in this section.
@@ -305,14 +297,12 @@ llvm::Function *CGNVCUDARuntime::makeModuleCtorFunction() {
     // Fatbin version.
     Values.addInt(IntTy, 1);
     // Data.
-    Values.add(makeConstantString(GpuBinaryOrErr.get()->getBuffer(), 
-                                  "", FatbinConstantName, 8));
+    Values.add(makeConstantString(FatbinCode, "", FatbinConstantName, 8));
     // Unused in fatbin v1.
     Values.add(llvm::ConstantPointerNull::get(VoidPtrTy));
-    llvm::GlobalVariable *FatbinWrapper =
-      Values.finishAndCreateGlobal("__cuda_fatbin_wrapper",
-                                   CGM.getPointerAlign(),
-                                   /*constant*/ true);
+    llvm::GlobalVariable *FatbinWrapper = Values.finishAndCreateGlobal(
+        "__cuda_fatbin_wrapper", CGM.getPointerAlign(),
+        /*constant*/ true);
     FatbinWrapper->setSection(FatbinSectionName);
 
     // GpuBinaryHandle = __cudaRegisterFatBinary(&FatbinWrapper);
@@ -331,6 +321,27 @@ llvm::Function *CGNVCUDARuntime::makeModuleCtorFunction() {
 
     // Save GpuBinaryHandle so we can unregister it in destructor.
     GpuBinaryHandles.push_back(GpuBinaryHandle);
+  };
+
+  // If there is a valid buffer with fatbinary code, embed the buffer.
+  // Otherwise, embed fatbinary code from files.
+  if (CGM.getCodeGenOpts().CudaGpuBinaryBuffer) {
+    const llvm::StringRef GpuBinaryBuffer(
+        CGM.getCodeGenOpts().CudaGpuBinaryBuffer->data(),
+        CGM.getCodeGenOpts().CudaGpuBinaryBuffer->size());
+    buildFatbinarySection(GpuBinaryBuffer);
+  } else {
+    for (const std::string &GpuBinaryFileName :
+         CGM.getCodeGenOpts().CudaGpuBinaryFileNames) {
+      llvm::ErrorOr<std::unique_ptr<llvm::MemoryBuffer>> GpuBinaryOrErr =
+          llvm::MemoryBuffer::getFileOrSTDIN(GpuBinaryFileName);
+      if (std::error_code EC = GpuBinaryOrErr.getError()) {
+        CGM.getDiags().Report(diag::err_cannot_open_file)
+            << GpuBinaryFileName << EC.message();
+        continue;
+      }
+      buildFatbinarySection(GpuBinaryOrErr.get()->getBuffer());
+    }
   }
 
   CtorBuilder.CreateRetVoid();
```

SimeonEhrig:
I have another idea. Maybe we can replace the `getFileOrSTDIN` function with another function that supports the virtual filesystem and bring it into the clang upstream. Then we can port it back to cling's clang-repo. What do you think?

tools/packaging/cpt.py
SimeonEhrig: Yeah, that's a problem. Unfortunately I cannot solve the problem in the `CMakeLists.txt` of cling, because it has to be activated in the clang build too. I think the best way to activate the `NVPTX` backend automatically is to add some code to llvm's `CMakeLists.txt`. Then it will also be enabled in the Clang and Cling project. Unfortunately, this will be a commit, which will not go upstream.

Axel-Naumann:
>  Maybe we can replace the getFileOrSTDIN function with another function

Perfect, yes!

SimeonEhrig:
> > Maybe we can replace the getFileOrSTDIN function with another function
> 
> Perfect, yes!

And now the "problem" of this solution. Backporting shouldn't be easy, because the clang code base is really old. In the meantime, the nvptx backend has been extended with functions for HIP[1]. So I think we have to port a lot of files back. I'm not sure, maybe we should stash this feature, because you wrote in #291 that you want to update the clang base at the end of the year. With a new base, the back port should be easy. The feature itself is not so important. It prevents the fatbin code from being written to a file and read by the same process again. The impact on performance is not noticeable. The main advantage of the feature is that it improves the independence of the file system. Especially for HPC it is very helpful.

1] [HIP](https://gpuopen.com/compute-product/hip-convert-cuda-to-portable-c-code/) is the AMD version of CUDA. Unlike CUDA, it can address AMD **and** NVIDIA GPUs. If the target is an NVIDIA GPU, it works like a wrapper framework. Therefore, it made sense to integrate the HIP part for NVIDIA GPUs into the nvptx backend. Often it is just an if-else that changes the function name from `cudaSomething()` to `hipSomething()`.

lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
```diff
@@ -223,316 +198,216 @@ namespace cling {
...
+  }
+
+  bool IncrementalCUDADeviceCompiler::generateFatbinary() {
+    // implementation ist copied from clangJIT
+    // (https://github.com/hfinkel/llvm-project-cxxjit/)
+
+    // The outer header of the fat binary is documented in the CUDA
+    // fatbinary.h header. As mentioned there, the overall size must be a
+    // multiple of eight, and so we must make sure that the PTX is.
+    while (m_PTX_code.size() % 7)
```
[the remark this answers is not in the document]
SimeonEhrig: This part of the code confuses me because there is a [commit](https://github.com/hfinkel/llvm-project-cxxjit/commit/d032b2c6925cc5b5e5e2bd681df2c50dc22df634) that explicitly changes the code:

```diff
      // The outer header of the fat binary is documented in the CUDA
      // fatbinary.h header. As mentioned there, the overall size must be a
      // multiple of eight, and so we must make sure that the PTX is.
-      while (DevCD->DevAsm.size() % 8)
+      while (DevCD->DevAsm.size() % 7)
           DevCD->DevAsm += ' ';
+      DevCD->DevAsm += '\0';

      // NVIDIA, unfortunatly, does not provide full documentation on their
      // fatbin format. There is some information on the outer header block in
 ```

I think the change is wrong because the alignment is 7, but it works. Because I can't understand it, I wouldn't change it.

Here is the content of the `fatbinary.h`

```c++
/*
 *  Copyright 2010 NVIDIA Corporation.  All rights reserved.
 *
 *  NOTICE TO USER: The source code, and related code and software
 *  ("Code"), is copyrighted under U.S. and international laws.
 *
 *  NVIDIA Corporation owns the copyright and any patents issued or
 *  pending for the Code.
 *
 *  NVIDIA CORPORATION MAKES NO REPRESENTATION ABOUT THE SUITABILITY
 *  OF THIS CODE FOR ANY PURPOSE.  IT IS PROVIDED "AS-IS" WITHOUT EXPRESS
 *  OR IMPLIED WARRANTY OF ANY KIND.  NVIDIA CORPORATION DISCLAIMS ALL
 *  WARRANTIES WITH REGARD TO THE CODE, INCLUDING NON-INFRINGEMENT, AND
 *  ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
 *  PURPOSE.  IN NO EVENT SHALL NVIDIA CORPORATION BE LIABLE FOR ANY
 *  DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES
 *  WHATSOEVER ARISING OUT OF OR IN ANY WAY RELATED TO THE USE OR
 *  PERFORMANCE OF THE CODE, INCLUDING, BUT NOT LIMITED TO, INFRINGEMENT,
 *  LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT,
 *  NEGLIGENCE OR OTHER TORTIOUS ACTION, AND WHETHER OR NOT THE
 *  POSSIBILITY OF SUCH DAMAGES WERE KNOWN OR MADE KNOWN TO NVIDIA
 *  CORPORATION.
 *
 */

#ifndef fatbinary_INCLUDED
#define fatbinary_INCLUDED

/* 
 * This is the fat binary header structure. 
 * Because all layout information is contained in all the structures, 
 * it is both forward and backward compatible. 
 * A new driver can interpret an old binary 
 * as it will not address fields that are present in the current version. 
 * An old driver can, for minor version differences, 
 * still interpret a new binary, 
 * as the new features in the binary will be ignored by the driver.
 *
 * This is the top level type for the binary format. 
 * It points to a fatBinaryHeader structure. 
 * It is followed by a number of code binaries.
 * The structures must be 8-byte aligned, 
 * and are the same on both 32bit and 64bit platforms.
 *
 * The details of the format for the binaries that follow the header
 * are in a separate internal header.
 */

typedef struct fatBinaryHeader *computeFatBinaryFormat_t;

/* ensure 8-byte alignment */
#if defined(__GNUC__)
#define __align__(n) __attribute__((aligned(n)))
#elif defined(_WIN32)
#define __align__(n) __declspec(align(n))
#else
#error !! UNSUPPORTED COMPILER !!
#endif

/* Magic numbers */
#define FATBIN_MAGIC 0xBA55ED50
#define OLD_STYLE_FATBIN_MAGIC 0x1EE55A01

#define FATBIN_VERSION 0x0001

/*
 * This is the fat binary header structure. 
 * The 'magic' field holds the magic number. 
 * A magic of OLD_STYLE_FATBIN_MAGIC indicates an old style fat binary. 
 * Because old style binaries are in little endian, we can just read 
 * the magic in a 32 bit container for both 32 and 64 bit platforms. 
 * The 'version' fields holds the fatbin version.
 * It should be the goal to never bump this version. 
 * The headerSize holds the size of the header (must be multiple of 8).
 * The 'fatSize' fields holds the size of the entire fat binary, 
 * excluding this header. It must be a multiple of 8.
 */
struct __align__(8) fatBinaryHeader
{
  unsigned int           magic;
  unsigned short         version;
  unsigned short         headerSize;
  unsigned long long int fatSize;
};

/* Code kinds supported by the driver */
typedef enum {
  FATBIN_KIND_PTX      = 0x0001,
  FATBIN_KIND_ELF      = 0x0002,
  FATBIN_KIND_OLDCUBIN = 0x0004,
} fatBinaryCodeKind;

#endif /* fatbinary_INCLUDED */
```

lib/Interpreter/IncrementalCUDADeviceCompiler.h
SimeonEhrig: The `llvm::Triple` is necessary for multi-platform support. I could reduce the interface to `setCuArgs(const clang::CompilerInstance& CI, const cling::InvocationOptions& invocationOptions)` if it is better.

`const int intprOptLevel` is a relic of the old implementation. The old implementation did not allow to change the optimization level at runtime. The new implementation should allow this.  I will check it in a future commit.

lib/Interpreter/IncrementalCUDADeviceCompiler.h
SimeonEhrig: See the comment before.

lib/Interpreter/IncrementalCUDADeviceCompiler.h
SimeonEhrig: `const` is only an optimization. Should I switch to `const std::string &` like the original interfaces of `cling::Intpereter`? `llvm::StringRef` is just a reference. Therefore it is lightweight.

DavidPoliakoff:
@SimeonEhrig: @ax3l just linked me this somewhere else. This is really excellent work. Hopefully in about a month or so I'll be able to play with Kokkos and this, if we have good CUDA support it could be fun to start playing with rapid development of Kokkos, switching kernels between host and device and the like. Half commenting just to say well done, half to be notified when this merges :)

ax3l:
Thank you for the kind words! Yes, one of our main motivations is tuning and playing with [Alpaka](https://github.com/ComputationalRadiationPhysics/alpaka) as well :)

SimeonEhrig:
Thank you so much. At the moment I am preparing a Jupyter notebook for a lecture with [jitify](https://github.com/NVIDIA/jitify) and the [redefinition feature](https://github.com/root-project/root/pull/4214). It's really nice to use, but just a "workaround". After the PR one of the next features will be the redefinition in CUDA mode. Then we can use it with the Runtime API, Alpaca, Kokkos, ... :-D

lib/Interpreter/IncrementalCUDADeviceCompiler.h
SimeonEhrig: replaced `const llvm::StringRef` with `const std::string&`

SimeonEhrig:
@Axel-Naumann I have removed the part of the code that requires changes to the clang base. Now the tests pass (except for some timeouts). There are only two comments left, which I can't solve without your feedback.

Axel-Naumann:
I have applied this to ROOT (we sync cling from ROOT). The PR https://github.com/root-project/root/pull/4616 will also run ROOT's test suite. I do not expect any issues there. I am closing this as all cling-related issues seem to be addressed; the rest happens in ROOT's PR!

Open questions:
- PTX-string as multiple of 8 bytes seems to contradict the code
- enable / disable nvptx backend

PR #341, opened 2020-04-29
IsClassOrFunction: Handle signedness modifiers

hahnjo:
When using these modifiers, a type can consist of multiple tokens
as for example "unsigned int". However the keyword "int" can also
be omitted, making "unsigned" itself also a valid type.

Note that this only handles the most basic case for the modifiers.
The size modifiers "short", "long", and "long long" would need
similar treatment. Moreover the standard permits any order for the
type specifiers, ie "unsigned long long int" and "long int unsigned
long" are both valid (and actually the same type).

### DIFF
--- a/lib/Utils/SourceNormalization.cpp
+++ b/lib/Utils/SourceNormalization.cpp
@@ -60,6 +60,31 @@ class MinimalPPLexer: public Lexer {
     return llvm::StringRef();
   }
 
+  ///\brief Skip an identifier of a function.
+  bool SkipIdentifier(Token& Tok) {
+    if (Tok.isNot(tok::raw_identifier)) {
+      // If we're not at an identifier, we might be still be in return value:
+      // A::B::C funcname() or int * funcname()
+      if (!SkipScopes(Tok))
+        return false;
+      if (!SkipPointerRefs(Tok))
+        return false;
+    }
+
+    // Function or class name should be in Tok now
+    if (Identifier(Tok).empty())
+      return false;
+
+    // Advance to argument list or method name
+    if (!LexClean(Tok))
+      return false;
+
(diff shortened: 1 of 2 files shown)


hahnjo:
gentle ping - are you taking pull requests over GitHub or shall I send it somewhere else?

lib/Utils/SourceNormalization.cpp
```diff
@@ -220,11 +221,19 @@ class MinimalPPLexer: public Lexer {
...
+          SeenSignedness = true;
+
         // Advance past keyword for below
         if (!LexClean(Tok))
           return kNONE;
+      } else if (First.equals("signed") || First.equals("unsigned")) {
+        SeenSignedness = true;
       }
 
+SkipIdentifier:
```
Axel-Naumann, line 236: Any chance we can avoid the `goto`?

hahnjo: I first considered a structure like
```c
while (true) {
  // ...
  if (SeenSignedness) {
    continue;
  }
  break;
}
```
but went with `goto` because the infinite loop is just hiding the intent. I'll try refactoring the code into a function, that should work equally well.

hahnjo: `SkipIdentifier` is now a function which is potentially called twice.

Alternatively I've tried `AdvanceTo(Tok, tok::l_paren)` instead of the whole logic now in `SkipIdentifier`. Unfortunately this also promotes lambdas or anything else with a brace in it to top level declarator which is probably not wanted.

Axel-Naumann:
Thanks for your contribution! Merged into ROOT - this will automagically end up in cling.

PR #353, opened 2020-08-27
Include windows.h in DynamicLibraryManagerSymbol.cpp 

### LINKED ISSUE #352: Missing windows include
I am working on the conda-forge packaging of 0.7, and all seems to work well on Linux and OSX

PR is here: https://github.com/conda-forge/cling-feedstock/pull/39.

On windows, the build is terminating with:

```
C:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\um\winnt.h(154): fatal error C1189: #error:  "No Target Architecture"
```

SylvainCorlay:
This fixes #352 (`fatal error C1189: #error:  "No Target Architecture"`).

The reason for the bug is that `libloaderapi.h` and `memoryapi.h` are not supposed to be included directly apparently, and it is checked that the right platform macro is defined, while it is only defined in windows.h.

### DIFF
--- a/lib/Interpreter/DynamicLibraryManagerSymbol.cpp
+++ b/lib/Interpreter/DynamicLibraryManagerSymbol.cpp
@@ -36,6 +36,7 @@
 #endif // __APPLE__
 
 #ifdef LLVM_ON_WIN32
+#include <windows.h>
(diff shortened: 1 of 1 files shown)


Axel-Naumann:
Thanks, merged! Our repo sync will upload this to cling.git in a bit.

PR #363, opened 2020-10-20
Fixed cling argument --cuda-path

SimeonEhrig:
The cling argument `--cuda-path` is necessary if the CUDA SDK is not installed under `/usr/local/cuda`, e.g. on HPC systems. The integration tests are also updated to handle a CUDA SDK under a non-standard location.

# Additional diagnostic

To find the bug, I also add some diagnostic functions for the PTX compiler.

1. Now error messages are prefixed in CUDA mode to better decide which of the two compiler pipelines is causing the error.

example
```
- example before:
error: cannot find CUDA installation.  Provide its path via --cuda-path, or pass -nocudainc to build without CUDA includes.
error: cannot find libdevice for sm_20. Provide path to different CUDA installation via --cuda-path, or pass -nocudalib to build without linking with libdevice.
error: cannot find CUDA installation.  Provide its path via --cuda-path, or pass -nocudainc to build without CUDA includes.

- example after:
cling: error: cannot find CUDA installation.  Provide its path via --cuda-path, or pass -nocudainc to build without CUDA includes.
cling-ptx: error: cannot find libdevice for sm_20. Provide path to different CUDA installation via --cuda-path, or pass -nocudalib to build without linking with libdevice.
cling-ptx: error: cannot find CUDA installation.  Provide its path via --cuda-path, or pass -nocudainc to build without CUDA includes.
```

2. Now, the class `cudaIncrementalDeviceCompiler` available through reflection via the `gCling` object.

### DIFF
--- a/include/cling/Interpreter/IncrementalCUDADeviceCompiler.h
+++ b/include/cling/Interpreter/IncrementalCUDADeviceCompiler.h
@@ -87,6 +87,9 @@ namespace cling {
     ///\brief Contains the PTX code of the current input
     llvm::SmallString<1024> m_PTX_code;
 
+    ///\brief Keep the ptx compiler args for reflection during runtime.
+    std::vector<std::string> argv;
+
     ///\brief Add the include paths from the interpreter runtime to a argument
     /// list.
     ///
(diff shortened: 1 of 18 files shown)


SimeonEhrig:
Merged via Root PR: https://github.com/root-project/root/pull/6707

PR #365, opened 2020-10-23
Setup Github Actions CI for Linux and macOS

reikdas:
We might want to move from Travis and Appveyor to using Github Actions for CI because Github Actions is more robust and provides a lot of advantages such as more concurrent builds, longer timeouts, individually examinable steps, etc. Also, we would be able to have CI for Windows, Linux and macOS in the same infrastructure since Github Actions supports all 3.
There are noticeably more builds in the presented Github Actions CI than the Travis CI because we are able to test Cling builds both from an LLVM binary as well as with compiling LLVM from source.

We will try and integrate Windows CI into Github Actions in a different PR because Windows builds (using cpt) seem to be broken currently and might need some additional effort to fix.

I found one downside of using Github Actions -  

- We cannot include macOS tests with XCode 9.3 (that we are currently testing in Travis), because Github Actions does not provide it. (We would need a self hosted runner to test this)

We do intend to remove travis at a later date if we integrate Github Actions CI, but we cannot yet since we would have to first integrate Deployment and Nightly Releases that are currently handled by Travis into our Github Actions. (@ax3l I remember you mentioned something about working on Continuous Deployments for Cling some time ago. Was there any progress on that? Otherwise I can try and get that done)

Something else that we need to look into wrt CI, is a way to test Cling's CUDA codegen as a part of CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,100 @@
+name: Cling CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+
+        include:
+          - name: ubuntu-16.04-gcc
+            os: ubuntu-16.04
+            compiler: gcc
+
+          - name: ubuntu-16.04-clang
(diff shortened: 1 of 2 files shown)


vgvassilev:
Looks good to me. What else we need to set up for this repository for this to work?

reikdas:
Merged via https://github.com/root-project/root/pull/6720

PR #366, opened 2020-11-02
Fix pip installation of packages via cpt and add Python 2 test to Github Actions CI

reikdas:
I think we should (atleast internally) formalize which Python versions cpt should support. 

We could do either of - 
1. Follow official Python release end of lifes, and only support Python 3.6 and above (and formally stop supporting Python versions as they reach their end of life). We should also probably add tests for the Python versions cpt officially supports to the new Github Actions CI.
2. Do not declare any supported Python versions on the README, but expect users to have atleast >= Python 2.7.9 and >= Python 3.4. This ensures that the Python version on the user's system comes with pip installed and makes it easier for cpt to install packages using pip. 

This PR addresses the latter and adds a Python 2.7 test to Github Actions CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,39 +20,58 @@ jobs:
           - name: ubuntu-16.04-gcc
             os: ubuntu-16.04
             compiler: gcc
+            python-version: 3.8
 
           - name: ubuntu-16.04-clang
             os: ubuntu-16.04
             compiler: clang
+            python-version: 3.8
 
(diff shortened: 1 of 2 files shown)


reikdas:
The default Python on Ubuntu 16.04, Ubuntu 18.04 and macOS 10.15 (the versions hosted on Github Actions), have the default Python as Python 2. So we are testing Python 2 through these and Python 3 through Ubuntu 20.04 - we don't need to explicitly test Python 2.

PR #382, opened 2021-01-23
Fix #381: Fix decode of str

### LINKED ISSUE #381: cpt.py failed to detect compiler: AttributeError: 'str' object has no attribute 'decode'
I am running `cpt.py` on my x86_64 systemd gentoo that has `python3.8.6` installed and used as the default python interpreter:

```
$ cpt.py --create-dev-env Release --last-stable pkg --stdlib libc++ --compiler clang --with-workdir cling/
Cling Packaging Tool (CPT)
Arguments vector: ['/home/nobodyxu/.local/bin/cpt.py', '--create-dev-env', 'Release', '--last-stable', 'pkg', '--stdlib', 'libc++', '--compiler', 'clang', '--with-workdir', 'cling/']

╔══════════════════════════════════════════════════════════════════════════════╗
║ cling (x86_64)                               Sat, 23 Jan 2021 18:41:41 +1100 ║
╚══════════════════════════════════════════════════════════════════════════════╝
Thread Model: POSIX
Operating System: Linux
Distribution: Gentoo
Release:
Revision: 2.7
Architecture: x86_64
Traceback (most recent call last):
  File "/home/nobodyxu/.local/bin/cpt.py", line 2037, in <module>
    cInfo = exec_subprocess_check_output(args['compiler'] + ' --version', srcdir).decode('utf-8')
AttributeError: 'str' object has no attribute 'decode'
```

I looked into function `exec_subprocess_check_output` and found the reason for the error:

```
def exec_subprocess_check_output(cmd, cwd):
    cmd = _convert_subprocess_cmd(cmd)
    out = ''
    try:
        out = subprocess.check_output(cmd, cwd=cwd, shell=False,
                                      stdin=subprocess.PIPE, stderr=subprocess.STDOUT).decode('utf-8')
    except subprocess.CalledProcessError as e:
        _perror(e)
    finally:
        return out
```

The function `exec_subprocess_check_output` has already `decode` the `bytes` into `str`, but in line 2037, where the error occurs, it is **decoded again**:

```
  cInfo = exec_subprocess_check_output(args['compiler'] + ' --version', srcdir).decode('utf-8')
```

Hence the error.

NobodyXu:
Fixed #381

### DIFF
--- a/tools/packaging/cpt.py
+++ b/tools/packaging/cpt.py
@@ -2034,7 +2034,7 @@ print('Revision: ' + REV)
 print('Architecture: ' + platform.machine())
 if args['compiler']:
   cInfo = None
-  cInfo = exec_subprocess_check_output(args['compiler'] + ' --version', srcdir).decode('utf-8')
+  cInfo = exec_subprocess_check_output(args['compiler'] + ' --version', srcdir)
   print("Compiler: '%s' : %s" % (args['compiler'], cInfo.split('\n',1)[0] if cInfo else ''))
 
 if len(sys.argv) == 1:


reikdas:
We don't actually need the `--compiler` flag in cpt. It is possible to set the build compiler using the `--with-cmake-flags` option.

PR to drop the flag here - https://github.com/root-project/root/pull/7598

reikdas:
Closed in favour of root-project/root#7598

PR #401, opened 2021-03-18
Add support for an auto exec script

### LINKED ISSUE #230: Some Feature request on cmake scripts
Scenario:
I have a project, with build tools of cmake, when I try to run it with cling. It prompts some headers are not included.

So I was wondering if there is a way to make use of the cmake script to load the build environment into cling and run directly?

Appreciate the great work!

sudo-panda:
Fixes #230 

On startup cling will automatically look for a file named .clinginit in the home directory and execute it as a cling script.

### DIFF
--- a/tools/driver/cling.cpp
+++ b/tools/driver/cling.cpp
@@ -19,6 +19,7 @@
 #include "llvm/Support/PrettyStackTrace.h"
 #include "llvm/Support/ManagedStatic.h"
 
+#include <stdlib.h>
 #include <iostream>
 #include <fstream>
 #include <vector>
@@ -106,6 +107,31 @@ int main( int argc, char **argv ) {
     Interp.loadFile(Lib);
 
   cling::UserInterface Ui(Interp);
+
+  // Get the home directory
+  std::string HomeDir;
+#ifdef __unix__
+  HomeDir = getenv("HOME");
+  HomeDir += "/";
+#elif defined(_WIN32)
+  HomeDir = getenv("DRIVE");
+  const char *HomePath = getenv("HOMEPATH");
+  HomeDir += HomePath;
+  HomeDir += "\\";
+#endif
+
+  // Look for .clinginit in the home directory and if found process it like a cling script
+  std::string CLingInit = HomeDir + ".clinginit";
+  const std::string Filepath = Interp.lookupFileOrLibrary(CLingInit);
+  if (!Filepath.empty()) {
+    std::ifstream File(Filepath);
+    std::string Line;
+    cling::Interpreter::CompilationResult Result;
+    while (std::getline(File, Line)) {
+      Ui.getMetaProcessor()->process(Line, Result, 0);
+    }
+  }
+
   // If we are not interactive we're supposed to parse files
   if (!Opts.IsInteractive()) {
     for (const std::string &Input : Opts.Inputs) {


tools/driver/cling.cpp
```diff
@@ -106,6 +107,31 @@ int main( int argc, char **argv ) {
...
+  // Get the home directory
+  std::string HomeDir;
+#ifdef __unix__
+  HomeDir = getenv("HOME");
+  HomeDir += "/";
+#elif defined(_WIN32)
+  HomeDir = getenv("DRIVE");
+  const char *HomePath = getenv("HOMEPATH");
+  HomeDir += HomePath;
+  HomeDir += "\\";
```
bellenot, line 120: You can maybe use the `%USERPROFILE%` environment variable instead?
```
   HomeDir = getenv("USERPROFILE");
```

sudo-panda: Alright

tools/driver/cling.cpp
```diff
@@ -106,6 +107,29 @@ int main( int argc, char **argv ) {
...
+
+  // Get the home directory
+  std::string HomeDir;
+#ifdef __unix__
+  HomeDir = getenv("HOME");
+  HomeDir += "/";
+#elif defined(_WIN32)
+  HomeDir = getenv("USERPROFILE");
+  HomeDir += "\\";
+#endif
```
Axel-Naumann, line 119: There's [`llvm::sys::home_directory()`](https://llvm.org/doxygen/namespacellvm_1_1sys_1_1path.html#a1286bcdea03c97a18eb1f41af524f3d3) for that.

tools/driver/cling.cpp
```diff
@@ -106,6 +107,29 @@ int main( int argc, char **argv ) {
...
+
+  // Look for .clinginit in the home directory and if found process it like a cling script
+  std::string CLingInit = HomeDir + ".clinginit";
+  const std::string Filepath = Interp.lookupFileOrLibrary(CLingInit);
```
Axel-Naumann, line 123: What is this adding? How is `Filepath` different from `CLingInit`? If this is about "does the file exist", you might as well just create the `ifstream` and check how it's doing.

tools/driver/cling.cpp
```diff
@@ -106,6 +107,29 @@ int main( int argc, char **argv ) {
...
+#ifdef __unix__
+  HomeDir = getenv("HOME");
+  HomeDir += "/";
+#elif defined(_WIN32)
+  HomeDir = getenv("USERPROFILE");
+  HomeDir += "\\";
+#endif
+
+  // Look for .clinginit in the home directory and if found process it like a cling script
+  std::string CLingInit = HomeDir + ".clinginit";
```
Axel-Naumann, line 122:
```suggestion
  std::string ClingInit = HomeDir + ".clinginit";
```

Axel-Naumann, line 122: I believe most other tools suffix `_init` which would suggest
```suggestion
  std::string CLingInit = HomeDir + ".cling_init";
```

tools/driver/cling.cpp
```diff
@@ -106,6 +107,29 @@ int main( int argc, char **argv ) {
...
+  const std::string Filepath = Interp.lookupFileOrLibrary(CLingInit);
+  if (!Filepath.empty()) {
+    std::ifstream File(Filepath);
+    std::string Line;
+    cling::Interpreter::CompilationResult Result;
+    while (std::getline(File, Line)) {
+      Ui.getMetaProcessor()->process(Line, Result, 0);
+    }
```
Axel-Naumann, line 130: How's that different from `cling::MetaProcessor::readInputFromFile()`?

tools/driver/cling.cpp
sudo-panda: `gdb` uses `.gdbinit` so I stuck with that but `.cling_init` is definitely better.

tools/driver/cling.cpp
sudo-panda: I had used these statements because if you look further down at [line 146](https://github.com/root-project/cling/blob/e23cf8438e590d2310efb49c073b0115497e4b3e/tools/driver/cling.cpp#L146) cling processes the input scripts in a similar way. Here again, using `cling::MetaProcessor::readInputFromFile()` seems to be the better option.

sudo-panda:
@bellenot , @Axel-Naumann could you please review the changes.

tools/driver/cling.cpp
```diff
@@ -106,6 +106,20 @@ int main( int argc, char **argv ) {
     Interp.loadFile(Lib);
 
   cling::UserInterface Ui(Interp);
+
+  // Get the home directory
+  llvm::SmallString<32> HomeDir;
+  llvm::sys::path::home_directory(HomeDir);
+
+  // Look for .cling_init in the home directory and if found process it like a cling script
+  llvm::SmallString<128> ClingInit(HomeDir.c_str());
```
Axel-Naumann, line 115: Why do we need a second variable? We should just append to the variable filled by `home_directory`

Axel-Naumann:
@reikdas all the CI builds fail - is this expected?

Axel-Naumann (changes requested):

tools/driver/cling.cpp
sudo-panda: I felt it was more expressive that way but your way is more efficient

reikdas:
> @reikdas all the CI builds fail - is this expected?

Yup. `cpt.py` hasn't been updated for LLVM 9 yet.

Axel-Naumann:
Thanks @reikdas ! Do you have an ETA for a fix? Just asking regarding scheduling this - whether I should wait for this fix or get this in (through ROOT) without.

tools/driver/cling.cpp
```diff
@@ -106,6 +106,17 @@ int main( int argc, char **argv ) {
...
   cling::UserInterface Ui(Interp);
+
+  // Look for .cling_init in the home directory and if found process it like a cling script
+  llvm::SmallString<128> ClingInit;
+  llvm::sys::path::home_directory(ClingInit);
+
+  llvm::sys::path::append(ClingInit, ".cling_init");
+
+  if(llvm::sys::fs::exists(ClingInit)) {
+    Ui.getMetaProcessor()->readInputFromFile(ClingInit.c_str(), nullptr, -1, true);
```
Axel-Naumann, line 117: `readInputFromFile` takes a `StringRef`; better provide parameter names as comments for non-obvious argument values, better cast `-1` to `size_t`:
```suggestion
    Ui.getMetaProcessor()->readInputFromFile(ClingInit, /*result*/ nullptr, /*posOpenCurly*/ (size_t)-1, /*lineByLine*/ true);
```
Please repair my change to follow 80char limit.

tools/driver/cling.cpp
```diff
@@ -106,6 +106,17 @@ int main( int argc, char **argv ) {
     Interp.loadFile(Lib);
 
   cling::UserInterface Ui(Interp);
+
+  // Look for .cling_init in the home directory and if found process it like a cling script
```
Axel-Naumann, line 110: That seems to exceed the 80chars/line limit.

tools/driver/cling.cpp
```diff
@@ -106,6 +106,17 @@ int main( int argc, char **argv ) {
...
+  llvm::SmallString<128> ClingInit;
+  llvm::sys::path::home_directory(ClingInit);
+
+  llvm::sys::path::append(ClingInit, ".cling_init");
+
+  if(llvm::sys::fs::exists(ClingInit)) {
```
Axel-Naumann, line 116:
```suggestion
  if (llvm::sys::fs::exists(ClingInit)) {
```

Axel-Naumann, line 116: These empty lines don't add much; suggest:
```suggestion
  llvm::SmallString<128> ClingInit;
  llvm::sys::path::home_directory(ClingInit);
  llvm::sys::path::append(ClingInit, ".cling_init");
  if(llvm::sys::fs::exists(ClingInit)) {
```

Axel-Naumann (changes requested):
Style changes - but that looks like the last refinement. Thanks!

reikdas:
> Thanks @reikdas ! Do you have an ETA for a fix? Just asking regarding scheduling this - whether I should wait for this fix or get this in (through ROOT) without.

A fix shouldn't be too far off. It is already fixed for builds using LLVM source (in my fork). We have hit kind of a roadblock with getting cpt to build Cling using an LLVM binary, but hopefully it will be resolved soon (ping @vgvassilev).

PR #402, opened 2021-03-25
[WIP][RFC][CUDA] Fix __constant__ variable not declared as global

### LINKED ISSUE #396: [CUDA] __constant__ memory declaration doesn't work
# Problem
The following code doesn't works in cuda mode:
``` C++
[cling]$ __constant__ int array[1024];
```

The following error is thrown:
```
input_line_3:2:2: error: __constant__ variables must be global
 __constant__ int array[1024];
 ^
/usr/local/cuda-8.0/include/host_defines.h:195:22: note: expanded from macro '__constant__'
        __location__(constant)
```

# possible cause
The bug should be inside the cling instance. Maybe the following source code will jited internal:
``` C++
void __cling_Un1Qu31(void* vpClingValue) {
 __constant__ int array[1024];
;
}
__constant__ int array[1024];
```

sudo-panda:
Fixes #396

### DIFF
--- a/lib/Interpreter/Interpreter.cpp
+++ b/lib/Interpreter/Interpreter.cpp
@@ -816,6 +816,8 @@ namespace cling {
     CO.ResultEvaluation = (bool)V;
     // CO.IgnorePromptDiags = 1; done by EvaluateInternal().
     CO.CheckPointerValidity = 1;
+    if(wrapReadySource.rfind("__constant__", 0) != -1)
+      wrapPoint = wrapReadySource.size();
     if (EvaluateInternal(wrapReadySource, CO, V, T, wrapPoint)
                                                      == Interpreter::kFailure) {
       return Interpreter::kFailure;


sudo-panda:
@SimeonEhrig  I have put up a commit but I am not sure if the current change is the best way to go.

We need to check if CUDA mode is activated before we check the `__constant__` declaration specifier. But it brings its own challenges as in CUDA mode the namespace wrapper seems applied twice. Once inside the process function for the the original C++ Interpreter [here](https://github.com/root-project/cling/blob/20cf2b234bcd1fcd1841e892247ee2ea322fab16/lib/Interpreter/Interpreter.cpp#L821) and once when the line is processed via the NVPTX Interpreter called [here](https://github.com/root-project/cling/blob/20cf2b234bcd1fcd1841e892247ee2ea322fab16/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp#L216). Now when the line is processed via the NVPTX Interpreter the `CUDAHost` Compiler option is set to `0` but the original Interpreter has it set to `1`. Is there any other way to check if CUDA mode is enabled?

Also is there a better way of checking if the declaration is a `__constant__` declaration?

SimeonEhrig:
@sudo-panda Great work! It was like expected.

I think, it make sense to move the check there: https://github.com/root-project/cling/blob/df5fbb3c46a92a0841145cde0655e26c88ddff1b/lib/Utils/SourceNormalization.cpp#L402

Yes, it is true that the code is processed twice, once for the host (x86) and once for the device (PTX). This video gives a small explanation:  https://youtu.be/HEGDII5lAfo?t=546

Short summary of the video. The `IncrementalCUDADeviceCompiler` class contains three things: setup functions for the PTX compiler, a `cling::Interpreter` instance that parses the code, and a PTX CodeGen. That means, the parse function of cling::Interpreter is executed twice. The first time for the host side and the second time for the device side. Here you can see an example: : https://github.com/root-project/cling/blob/20cf2b234bcd1fcd1841e892247ee2ea322fab16/lib/Interpreter/Interpreter.cpp#L956-L958 

To decide which interpreter instance is running, there are the options: `m_Opts.CompilerOpts.CUDAHost` and `m_Opts.CompilerOpts.CUDADevice`.

So in this case we need to query `if(CUDAHost || CUDADevice)` since your fix is relevant to both compiler instances. However, if we move the code to the `cling::utils::getWrapPoint` function, we can use the `LangOpts.CUDA` variable since it is also valid for the host and device compilers.

Finally, I have some bad news. Your fix is only one part to enable constant memory. There is also a problem with the `cudaGetSymbolAddress` function. It is related to this problem: https://github.com/root-project/cling/issues/395
At the moment this example does not work:

```C++
#include <iostream>

__constant__ int cA[1];

__global__ void g(int *out){
  *out = cA[0];
}

int A[] = {42};
int *cA_ptr;
cudaGetSymbolAddress((void **)&cA_ptr, cA);
cudaMemcpy(cA_ptr, A, sizeof(int), cudaMemcpyHostToDevice);

int host = -1;
int *device;

cudaMalloc((void**)&device, sizeof(int));
g<<<1,1>>>(device);
cudaMemcpy(&host, device, sizeof(int), cudaMemcpyDeviceToHost);

// should print 42
std::cout << host << std::endl;
```

I will spend some time with this problem. You can do it too if you want, but I think it's much more complicated. Nevertheless, your fix is really helpful and we will continue to bring it in the master.

PR #455, opened 2022-05-09
fix DynamicLibraryManager::loadLibrary if the parameter resolve is `false`

SimeonEhrig:
The const reference of lResolved in the statement `const std::string& canonicalLoadedLib = resolved ? libStem.str() : lResolved;` is a copy for unexplained reason

It also fixes the bug, that the `libcudart.so` is not correctly load at start up in CUDA mode.

### DIFF
--- a/lib/Interpreter/DynamicLibraryManager.cpp
+++ b/lib/Interpreter/DynamicLibraryManager.cpp
@@ -356,12 +356,17 @@ namespace cling {
     }
 
     std::string lResolved;
-    const std::string& canonicalLoadedLib = resolved ? libStem.str() : lResolved;
     if (!resolved) {
       lResolved = lookupLibrary(libStem);
       if (lResolved.empty())
         return kLoadLibNotFound;
     }
+
+    // For reasons that are not clear, canonicalLoadedLib is a copy of lResolved and not 
+    // a reference if resolved is false.
+    // Therefore the type of canonicalLoadedLib const std::string and not a reference 
+    // to make it clear.
+    const std::string canonicalLoadedLib = resolved ? libStem.str() : lResolved;
 
     if (m_LoadedLibraries.find(canonicalLoadedLib) != m_LoadedLibraries.end())
       return kLoadLibAlreadyLoaded;


SimeonEhrig:
@vgvassilev I didn't found any explanation, why this is a bug, but I could reproduce it with GGC 9 and Clang 13.

vgvassilev:
@SimeonEhrig, can you elaborate a bit more? Otherwise this patch looks good to me. Would we be able to create a test for it?

SimeonEhrig:
> @SimeonEhrig, can you elaborate a bit more (EDIT: elaborate more in the commit message)? Otherwise this patch looks good to me. Would we be able to create a test for it?

I will extend the commit message.
Maybe we should change `const std::string& canonicalLoadedLib = resolved ? libStem.str() : lResolved;` to `const std::string canonicalLoadedLib = resolved ? libStem.str() : lResolved;` because at least for the false case it is a copy. Otherwise it could confuse people in future.

I have to check, if we can test the `loadLibrary()` function with a unresolved path and the integration tests.

vgvassilev:
Can you move this PR on the ROOT side?

SimeonEhrig:
> The reason is something else here: In the statement `const std::string& canonicalLoadedLib = resolved ? libStem.str() : lResolved;` the `std::string` returned from `libStem.str()` is a temporary object and binding it to a const reference is probably UB (not a language lawyer, might not be the right term here; but it's definitely not correct). Because of this, some compiler optimizations do funny things, see https://godbolt.org/z/71joz13aT and try replacing the `get_str()` call with `b`. Neither of this should make a difference because `argc` is always 1, but it changes the output. So yes, the change definitely improves things, please open on the ROOT side with an updated explanation smiley

@hahnjo  Thanks for the example. I did a stupid mistake and could not reproduce it. But I think your explanation is wrong. 
First assigning a temporary object to a const reference is not undefined behavior. Actual, it is a C++ feature and allows to extend the lifetime of the temporary object to the lifetime of the const reference: https://en.cppreference.com/w/cpp/language/lifetime#Temporary_object_lifetime

Also the optimizer is not responsible for it. It is done by the compiler frontend, see cppinsight.io (https://cppinsights.io/s/66802de3). The compiler add the copy constructor for sting `a` if the other branch uses the return value of `get_str`:

```c++
const std::string & ref = argc ? a : get_str();
// resolved to by the compiler:
const std::basic_string<char> & ref = static_cast<bool>(argc) ? std::basic_string<char>(a) : get_str();
```

And I think, I found the reason, why the compiler insert the copy constructor. It is a implicit cast because the value types of `a` and the return type `get_str()` are different. `a` is a rvalue and the return value of `get_str()` is a prvalue. Rule 3.3.1 in the [cppreference](https://en.cppreference.com/w/cpp/language/operator_other#Conditional_operator) says for this case, that the second argument needs to be casted to a prvalue. Therefore the compiler insert the copy constructor and we have our "strange" behavior.

If you agree with my explanation, I will change the comment and the commit message and open the PR in ROOT.

hahnjo:
@SimeonEhrig I think I agree, thanks for digging up more details (I have to admit I stopped after being able to reproduce in a simple example). Yes, please open a PR targeting the ROOT repository.

SimeonEhrig:
automatic merged via ROOT PR (https://github.com/root-project/root/pull/10593)

Issue #426, opened 2021-08-24
'cmakeDir' is not defined


indrakumarreddy007:
while Compiling Cpt.py in windows, I am getting the following error.

File "cpt.py", line 2308, im <module>
   get_win_dep()
File "cpt.py", line 1347, in get_win_dep
 os.rename(os.path.join(tmp_bin_dir, 'cmake-3.6.2-win64-x64'), cmakeDir)
NameERROR: name 'cmakeDir' is not defined

cmakeDir is not declared in cpt.py

when i tried to build using cmaketext.txt, i am facing 

CMake Error at CMakeLists.txt:42 (message):
  llvm-config not found -- LLVM_CONFIG-NOTFOUND

Could you please help to build cling project.
![ClingError](https://user-images.githubusercontent.com/37694439/130565553-3003f8b9-a45a-4986-a2bb-b3c49282c2c1.PNG)

poopsicles:
editing the python script to add cmakeDir = TMP_PREFIX + "\bin\cmake\" worked for me

PR #458, opened 2022-05-26
Fixed undefined cmakeDir error mentioned in issue #426

saisoma123:
Added cmakeDir for windows building of cling, this removed the error mentioned in #426, although the undefined llvm_flags issue is still there. This is interesting since it seems like this same llvm_flags issue happens for both Windows and Ubuntu, and possibly on other platforms.

### DIFF
--- a/tools/packaging/cpt.py
+++ b/tools/packaging/cpt.py
@@ -1342,7 +1342,7 @@ def get_win_dep():
     tmp_bin_dir = os.path.join(TMP_PREFIX, 'bin')
     zip.extractall(tmp_bin_dir)
     print('Remove file: ' + os.path.join(TMP_PREFIX, 'cmake-3.6.2.zip'))
-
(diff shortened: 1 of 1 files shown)


Darley-Wey:
> Added cmakeDir for windows building of cling, this removed the error mentioned in #426, although the undefined llvm_flags issue is still there. This is interesting since it seems like this same llvm_flags issue happens for both Windows and Ubuntu, and possibly on other platforms.

I also met this problem with llvm_flags on windows, so how to fix it?

saisoma123:
Well I did this on Ubuntu for llvm_flags, I called the download_llvm_binary() function within the with_binary_llvm block that’s within the last-stable block. This fixed the llvm_flag issue, but now there’s a source directory mismatch error that I’m currently working on fixing. I’m assuming something similar would happen on windows I can test it.

Issue #515, opened 2024-01-21
Update build-from-source instruction


lackhole:
### Explain what you would like to see improved
Update **[Building from source](https://github.com/root-project/cling?tab=readme-ov-file#building-from-source)** section of README.
LLVM has been updated

### Optional: share how it could be improved
* `nvptx` should be `NVPTX`. Build targets are case-sensitive. See https://llvm.org/docs/CMake.html
* `CMAKE_BUILD_TYPE` is mandatory.

### To Reproduce
Copy and paste current README
#### First error
```
CMake Error at CMakeLists.txt:90 (message):
  

  No build type selected.  You need to pass -DCMAKE_BUILD_TYPE=<type> in
  order to configure LLVM.

  Available options are:

    * -DCMAKE_BUILD_TYPE=Release - For an optimized build with no assertions or debug info.
    * -DCMAKE_BUILD_TYPE=Debug - For an unoptimized build with assertions and debug info.
    * -DCMAKE_BUILD_TYPE=RelWithDebInfo - For an optimized build with no assertions but with debug info.
    * -DCMAKE_BUILD_TYPE=MinSizeRel - For a build optimized for size instead of speed.

  Learn more about these options in our documentation at
  https://llvm.org/docs/CMake.html#cmake-build-type
```

#### Second error
(After providing `CMAKE_BUILD_TYPE`)
```
CMake Error at CMakeLists.txt:898 (message):
  The target `nvpts' is experimental and must be passed via
  LLVM_EXPERIMENTAL_TARGETS_TO_BUILD.
```
It is hard to resolve this error only with the diagnostic message since the real reason is that target is case-sensitive, not because `nvptx` is experimental.

### Additional context
Should I open a PR?

ferdymercury:
> Should I open a PR?

Thanks for reporting, yes please, at https://github.com/root-project/root

gabemorris12:
This is still not fixed and it's a single line in the readme.

ferdymercury:
Feel free to transfer this PR (https://github.com/root-project/cling/pull/518) to the upstream repo: https://github.com/root-project/root/pulls

sunflower-knight:
I cant build cling from source ； download the clingv1.0 release  source code from github release ; download root project's llvm code in branch cling-latest ; 
When I compile ，it told me 

``` bash
 Building CXX object tools/cling/lib/Interpreter/....clingInterpreter.dir/DynamicLibraryManagerSymbol.cpp.o
[...]

PR #518, opened 2024-01-31
Fix CMake build instructions

Arcnor:
This is a very simple PR to improve the CMake build instructions. The existing ones are a bit too convoluted and can be simplified a lot, plus there is an issue mentioned at #515 about the `NVPTX` using the wrong name, which leads to a misleading error. (The first issue mentioned there about the build type only shows when you configure CMake with a non-multi-config generator, so I've left that one alone for the time being).

I hope this helps!

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,13 +40,10 @@ Our nightly binary snapshots can be found
 ### Building from Source
 
 ```sh
-git clone https://github.com/root-project/llvm-project.git
-cd llvm-project
-git checkout cling-latest
-cd ../
+git clone https://github.com/root-project/llvm-project.git -b cling-latest
 git clone <cling>
-mkdir cling-build && cd cling-build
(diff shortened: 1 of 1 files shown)


daniel347x-cs:
This simple PR seems super important to me. I wonder how many people, almost like myself, spending hours trying to follow the build instructions and failing, simply give up and don't use it or become involved.

Seems a shame not to make the very simple change of providing proper build instructions in the README that work in 2024.

ferdymercury:
I believe this repo acts as a mirror, please move your PR here instead: https://github.com/root-project/root/pulls (subdirectory interpreter/cling). Thanks

LiAuTraver:
Really spent a whole day failing finally found issue😭wish I could find this PR earlier but, until now neither the `root` nor this repo has fixed this instruction error, sadly it's half-year passed.

LiAuTraver:
Yes you are right. Plus i searched the PR/issue at `root` but not found any similiar problem. I create one myself https://github.com/root-project/root/pull/15972#issue-2386422790. I really wish people come after do not ran into this problem, for it's a painful process building from source on an ordinary computer which results my PC was forced to shutdown several times. 😭

ferdymercury:
Thanks for moving the PR there, I'll close this one then.

PR #528, opened 2024-07-02
Minor build guide tweaks [dup,closed]

LiAuTraver:
I'm not sure if it's just me, but after spending an entire day building Cling, CMake kept notifying me that the target `nvptx` was not found.  It's really confused someone who has little experience with CMake like me. Then I searched the LLVM repository and found that the target name was capitalized as `NVPTX`. CMake is clearly case-sensitive. If it's not just me, then there's an issue with the build guide. I simply changed `nvptx` to `NVPTX` and added the comment `# if failed, try nvptx instead` after the command just in case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,9 +44,10 @@ git clone https://github.com/root-project/llvm-project.git
 cd llvm-project
 git checkout cling-latest
 cd ../
-git clone <cling>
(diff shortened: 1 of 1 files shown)


LiAuTraver:
i found it duplicate. i'll close this pr

PR #538, opened 2024-11-30
fix error handling when parsing invaild code

Vipul-Cariappa:


Vipul-Cariappa:
ping @vgvassilev

mcbarton:
@Vipul-Cariappa We will need this patch available in the v1.1 branch as well, as that is what we are building against in the ci

vgvassilev:
Can you move that over to root-project/root -- this is where cling gets its patches from.

PR #565, opened 2026-03-23
Disable concurrent compilation to prevent dangling Module pointers

0xeb:
## Summary

Cling uses a single `LLVMContext` (`SingleThreadedContext`) and keeps raw pointers to `Module`s (`Transaction::m_CompiledModule`) for later unloading via `TransactionUnloader::unloadModule`. When concurrent compilation is enabled, LLJIT sets `CloneToNewContextOnEmit` which clones each `Module` into a fresh context during materialization, destroying the original. That turns `m_CompiledModule` into a dangling pointer and crashes in `RevertTransaction` (use-after-free in the Module's function list).

**Fix**: Call `Builder.setSupportConcurrentCompilation(false)` in the `IncrementalJIT` constructor.

## Test plan

- Without this fix, `undo()` / `RevertTransaction` crashes with use-after-free
- With the fix, 34/34 tests pass including undo/rollback tests
- Tested on ARM64 Linux and Windows x64

### DIFF
--- a/lib/Interpreter/IncrementalJIT.cpp
+++ b/lib/Interpreter/IncrementalJIT.cpp
@@ -551,6 +551,15 @@ IncrementalJIT::IncrementalJIT(
   Builder.setDataLayout(m_TM->createDataLayout());
   Builder.setExecutorProcessControl(std::move(EPC));
 
+  // Cling uses a single LLVMContext (SingleThreadedContext) and keeps raw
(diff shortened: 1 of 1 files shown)


ferdymercury:
Thanks. PRs should be submitted instead to this repo: https://github.com/root-project/root/
This one just acts as mirror.

